PR #2, opened 2024-10-10
Create pull_request_template.md, add CODEOWNERS for default reviewers, tiny README update

kathrynalpert:
First pass at a pull request template

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,4 +1,4 @@
-[Reference the ticket you are addressing and provide a succinct and descriptive title for the pull request, e.g., "SCOUT-12 Improve caching mechanism for API calls"]
+# [Reference the ticket you are addressing and provide a succinct and descriptive title for the pull request, e.g., "SCOUT-12 Improve caching mechanism for API calls"]
 
 ## Type of change
 - [ ] New feature
@@ -8,10 +8,12 @@
 - [ ] Documentation update
 - [ ] Test update
 
-## Product description
+## Description
+
+### Product
 [Provide a summary explanation of your changes from a product/user perspective. More details should be found in your updates to the user docs.]
 
-## Technical description
+### Technical
 [Provide a summary explanation of your changes from a technical perspective. More details should be found in your updates to the technical docs.]
 
 ## Impact
@@ -33,7 +35,7 @@
 ### Backward compatibility
 [Any changes to the data model, APIs, dependencies?]
 
-## Test description
+## Testing
 [Detail the testing you have performed to ensure that these changes function as intended. Include information about the test automation you've added, as well as the manual testing you've performed. Be sure to reference areas of impact, above.]
 
 ## Note for reviewers
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,12 +1,17 @@
 # [Reference the ticket you are addressing and provide a succinct and descriptive title for the pull request, e.g., "SCOUT-12 Improve caching mechanism for API calls"]
 
 ## Type of change
+- [ ] Work behind a feature flag
 - [ ] New feature
 - [ ] Improvement
 - [ ] Bug fix
-- [ ] Refactor
+- [ ] Refactor (code improvement with no functional changes)
 - [ ] Documentation update
 - [ ] Test update
+
+If this pull request is for work that is behind a feature flag, or for documentation or test updates, most of the details below are not required; The level of attention to each is left to the discretion of the developer. 
+
+For all other change types, the developer should attempt to provide as much detail as is reasonable.
 
 ## Description
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,3 @@
-# Pull request template
-
-## Title
 [Reference the ticket you are addressing and provide a succinct and descriptive title for the pull request, e.g., "SCOUT-12 Improve caching mechanism for API calls"]
 
 ## Type of change
@@ -18,7 +15,6 @@
 [Provide a summary explanation of your changes from a technical perspective. More details should be found in your updates to the technical docs.]
 
 ## Impact
-[Discuss the impact of your changes on the project.]
 
 ### Security 
 
@@ -49,4 +45,4 @@
 - [ ] I have added or updated user documentation, if appropriate.
 - [ ] I have added or updated technical documentation, including an architectural decision record, if appropriate.
 - [ ] I have added unit tests, unless this is a test code PR.
-- [ ] I have added end-to-end tests.
+- [ ] I have added end-to-end tests, unless this is a documentation-only PR.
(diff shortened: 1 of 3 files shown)


johnflavin (commented):
As a reviewer I think all this context could be very helpful. But as a developer I think about the time it would take to gather all this information and make sure I can check all the boxes (literally) and it seems this puts a pretty high barrier in front of making a pull request.

This seems to run at cross purposes to trunk-based development, which encourages a high rate of merging smaller changes. This template would make creating a pull request a heavy and time-consuming operation, which disincentivizes creating a lot of pull requests and would tend to make them bigger and more complete.

I don't know what to do with that idea, though. Do we want to be doing a lot of lightweight PRs with work in progress (so long as nothing is broken)? Or do we want to be doing fewer, heavier PRs with rich context and complete features?

kathrynalpert:
This is what I would like to see from a developer before they hand off their work to QA. The aim is to provide structure that supports developers in validating their work and makes quality a shared responsibility. 

GitHub supports multiple PR templates, perhaps we could differentiate between "subtask" PRs and "handoff" PRs.

.github/pull_request_template.md
```diff
@@ -0,0 +1,52 @@
+# Pull request template
+
+## Title
+[Reference the ticket you are addressing and provide a succinct and descriptive title for the pull request, e.g., "SCOUT-12 Improve caching mechanism for API calls"]
```
johnflavin, line 4: I would think this could be the title of the PR itself rather than within a `Title` section in the body.

.github/pull_request_template.md
```diff
@@ -0,0 +1,52 @@
+# Pull request template
```
johnflavin, line 1: Nit: I'm not entirely sure what a pull request created from this template would look like, but I think this title would be included in every pull request. That seems unnecessary.

.github/pull_request_template.md
```diff
@@ -0,0 +1,52 @@
...
+- [ ] Test update
+
+## Product description
+[Provide a summary explanation of your changes from a product/user perspective. More details should be found in your updates to the user docs.]
+
+## Technical description
+[Provide a summary explanation of your changes from a technical perspective. More details should be found in your updates to the technical docs.]
+
+## Impact
+[Discuss the impact of your changes on the project.]
```
johnflavin, line 21: I'm not sure how I would answer the prompt in this section.

johnflavin (commented):
> This is what I would like to see from a developer before they hand off their work to QA. The aim is to provide structure that supports developers in validating their work and makes quality a shared responsibility.

Would some of this information be more appropriate to include in a JIRA ticket? That way the individual PRs don't need to represent the totality of the work, and the information is more accessible to people who may not be living and breathing the repo every day.

That said, I'm kind of inclined to just go with this pretty much as is and try it out. If we feel like something is wrong we can always make adjustments later.

.github/pull_request_template.md
andylassiter: This does seem vague.

andylassiter:
The burden could be a bit high if we are doing more regular small commits, but I'm also inclined to roll with this and we can adapt as needed.

.github/pull_request_template.md
kathrynalpert: Updated

.github/pull_request_template.md
```diff
@@ -0,0 +1,50 @@
+# [Reference the ticket you are addressing and provide a succinct and descriptive title for the pull request, e.g., "SCOUT-12 Improve caching mechanism for API calls"]
+
+## Type of change
```
cmoore01, line 3: Is the idea that a PR should check exactly 1 of these boxes, or `1-n`?

.github/pull_request_template.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Impact
+
+### Security 
+
+##### Authorization
+[Do your changes add or modify user roles? Do they impact the data users are able to see, or the actions they are able to take? Could a user modify a REST API path and see data they aren't supposed to see? Were you mindful of least privilege?]
+
+##### Appsec
+[Did you review your changes with application security in mind? See the [OWASP list](https://github.com/0xRadi/OWASP-Web-Checklist).]
```
cmoore01, line 27: What is intended to be done here? This checklist looks prohibitively expensive to execute (where every PR would involve a month of security testing)

johnflavin (approved):
LGTM

.github/pull_request_template.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Performance
+[At what scale do we expect this to operate? How have you verified that it can do so?]
```
cmoore01, line 30: Is this feasible? Requiring developer scale testing for large scale feels like a huge impediment to getting anything committed.

cmoore01 (commented):
My biggest overall concern is this process feels in tension with #3 . The process here feels rather heavy where small, frequent PRs from #3 do not seem feasible.

.github/pull_request_template.md
kathrynalpert: I was thinking exactly one, but unfortunately, it doesn't look like radio buttons are supported. That said, I don't suppose it's problematic to have folks check all that apply.

.github/pull_request_template.md
kathrynalpert: My aim is to prompt the developer to think about whether any of these concerns are relevant to the work they've performed, definitely not expecting them to go through the full checklist with every change. E.g., are you accepting new inputs that should be sanitized? Are you providing a way for a malicious user to DDoS the system or execute arbitrary code? Did you make a change that could have broken session expiration? ... etc.

.github/pull_request_template.md
kathrynalpert: I think it's reasonable to ask what scale we're planning to support with the feature/change, and correspondingly, the scale at which we've tested. If we don't have time/ability to test at scale, then we need to either be clear that we don't support it, or be pretty confident that we could support it, even without personally verifying it (e.g., using some off the shelf tool that has a scalability guarantee).

kathrynalpert:
Agreed, we are all noticing the tension between this template and the TBD "small, frequent PRs." I've attempted to address this concern with [my last change](https://github.com/washu-tag/scout/pull/2/commits/8cd9db006b5d32f98e135937b0c79b6dffd3a2ed). If the PR is for work behind a feature flag, the full template is not appropriate. However, when the developer removes that feature flag, the corresponding PR will need to be QAed before merge, and the developer should work through the checklist before that handoff. Hopefully the actual amount of code changed in the PR is not hugely significant (maybe it's constrained to just removing the feature flag and updating tests & docs, which may also need a feature flag concept).

We used to have the developer merging feature code onto `develop` and putting a bunch of info into a Jira ticket for QA (QA notes, technical changes, etc.). This means a potentially flawed feature lands on the main line while we wait for QA, which is in tension with the TBD concept of the trunk being "always releasable." Now, we have the developer put more information into the PR directly and handoff to QA before the code makes it onto the main line.

andylassiter (approved):

PR #11, opened 2025-01-28
SCOUT-27 SCOUT-28 Integration tests for scout & JSON+YAML linting

cmoore01:
# SCOUT-27 SCOUT-28 adds integration tests running queries against delta lake with spark and asserting results. Also enables linting for JSON and YAML files with helm templates excluded

## Type of change
- [X] New feature
- [X] Improvement
- [ ] Bug fix
- [X] Documentation update
- [X] Test update

### Product
N/A

### Technical
This adds a new `tests` project using gradle for running some queries against a delta lake with spark. There is currently an assumption that the delta lake contains exactly a specific data set, but we could relax that constraint at a later time. The test spec depends on a subproject of the data-generator.

##### Authorization
N/A

##### Appsec
N/A

### Performance
N/A

### Data
N/A

### Backward compatibility
N/A

## Testing
I ran the tests against our existing delta lake and verified they all passed, along with manually tweaking the expected results on several of the tests and seeing them fail.

## Note for reviewers
N/A

## Checklist
- [X] I have added or updated user documentation, if appropriate.
- [X] I have added unit tests, unless this is a test code PR.
- [X] I have added end-to-end tests, unless this is a documentation-only PR.


tests/src/test/resources/config/local.json.sample
```diff
@@ -0,0 +1,11 @@
+{
+  "sparkConfig": {
+    "spark.sql.catalog.spark_catalog": "org.apache.spark.sql.delta.catalog.DeltaCatalog",
+    "spark.hadoop.fs.s3a.access.key": "admin",
+    "spark.hadoop.fs.s3a.secret.key": "password",
+    "spark.hadoop.fs.s3a.endpoint": "http://10.27.107.3:9000",
```
johnflavin, line 6: This means we can't run this test on CI right now, correct? It would only succeed if run manually on the existing dev VMs.

tests/src/test/resources/spark_queries.json
```diff
@@ -0,0 +1,146 @@
...
+        "expectedNumResults": 397,
+        "additionalValidation": {
+          "type": ".FixedColumnsValidator",
+          "fixedColumns": {
+            "zds_1_study_instance_uid": ["None"],
+            "msh_12_version_id": ["2.4"]
+          }
+        }
+      },
+      "sql": "SELECT * FROM syntheticdata WHERE zds_1_study_instance_uid='None'",
```
johnflavin, line 30: These `"None"` values are the first things I want to clean up. I have the code ready to do that. But that raises questions for me about these tests, which I'll put in another comment.

johnflavin (approved):
I think this is a good first step on testing the data. But it's definitely based on some simplifying assumptions, as you said. 

If/when I modify the code that writes this data in a way that changes assumptions the tests are based on, we would presumably want to update the tests too. But we don't necessarily have to. I could rerun the data ingestion to a new table and leave a preserved immutable test table that has the right data to make the tests pass. But if we do that, what are these tests really testing? We're just reading the same file over and over and asserting that we know what's in it. That's not particularly valuable in demonstrating our application works properly.

The real value of the tests would be to do a read/write round trip. We would have some sample HL7 logs stored with the tests, we deploy the application in a limited way and run the ingest orchestration, then read our newly written data set and assert on its contents.

tests/src/test/resources/config/local.json.sample
cmoore01: In CI, we could always construct a different JSON config and provide it to the tests.

cmoore01:
> The real value of the tests would be to do a read/write round trip. We would have some sample HL7 logs stored with the tests, we deploy the application in a limited way and run the ingest orchestration, then read our newly written data set and assert on its contents.

I agree, but is the application actually to this point yet? That is, can I make a github action to deploy it, tell it ingest a bunch of HL7 log files, and create a delta lake?

johnflavin:
> I agree, but is the application actually to this point yet? That is, can I make a github action to deploy it, tell it ingest a bunch of HL7 log files, and create a delta lake?

In principle it is possible to do that, yes. All you'd need to set up would be minio, temporal, and the two temporal workers that implement the ingest workflow. I can't be sure it would work because I haven't done it, but it seems doable. Also we don't have an ansible playbook for it, so it would require a lot of knowledge of the right `helm` commands to run.

PR #19, opened 2025-02-05
Linting update example

cmoore01:
**This PR is not intended to be merged**

For the gradle projects within the repo, we are currently using `google_checks.xml` directly from the checkstyle jar, which tries to support the Google Java style guide as closely as possible. The style we frequently use naturally matches up closely to Google Java, but not quite exactly. Additionally, there are many, many warnings in the current linting jobs that will quickly just become noise we become de-sensitized to and ignore.

This PR has a suggestion of a checkstyle configuration we could adopt instead that is a fork of the original `google_checks.xml`. As an example to visualize this, `Example.java` has no warnings or info messages with `google_checks.xml`, while `ExampleCustom.java` has these two info messages with the custom checkstyle XML:
```
[ant:checkstyle] [INFO] /Users/moore.c/dev/scout/orchestration/temporal-java/src/main/java/edu/washu/tag/temporal/ExampleCustom.java:5:1: Missing a Javadoc comment. [MissingJavadocType]
[ant:checkstyle] [INFO] /Users/moore.c/dev/scout/orchestration/temporal-java/src/main/java/edu/washu/tag/temporal/ExampleCustom.java:12:5: Missing a Javadoc comment. [MissingJavadocMethod]
```

With my reasoning, the changes in the custom XML are (you can also diff against https://github.com/checkstyle/checkstyle/blob/checkstyle-10.21.2/src/main/resources/google_checks.xml):
1. `<module name="FileTabCharacter">` upgraded from `warning` to `error`. It is rather easy to keep tabs out of the code, so we should just make it a hard rule and fail the linting job if they are detected.
2. `LineLength` max updated from 100 to 160. 100 characters seems quite restrictive that enforces too many line returns that might make code more tedious to look at.
3. `<module name="SingleSpaceSeparator">` added with `error` severity. This should catch code like:
```
ObjectMapper objectMapper = new ObjectMapper();
int          a            = 10;
```
4. `<module name="OneTopLevelClass">` and `<module name="OneStatementPerLine">` upgraded to errors, with the same reasoning as the first point.
5. Indentation configuration updated from:
```
<module name="Indentation">
            <property name="basicOffset" value="2"/>
            <property name="braceAdjustment" value="2"/>
            <property name="caseIndent" value="2"/>
            <property name="throwsIndent" value="4"/>
            <property name="lineWrappingIndentation" value="4"/>
            <property name="arrayInitIndent" value="2"/>
</module>
```
to:
```
<module name="Indentation">
            <property name="basicOffset" value="4"/>
            <property name="braceAdjustment" value="4"/>
            <property name="caseIndent" value="4"/>
            <property name="throwsIndent" value="4"/>
            <property name="lineWrappingIndentation" value="4"/>
            <property name="arrayInitIndent" value="4"/>
</module>
```
2 spaces are rather cramped and it can be a bit difficult to identify indentation levels. Note that the severity has not been changed, so the default of "warning" still holds.
6. `<module name="MissingJavadocMethod">` and `<module name="MissingJavadocType">` downgraded from "warning" to "info". There are many small methods where we should not need Javadoc.

---
What do I need from you?

What do you think of these changes? Should they be more aggressive? Should _every_ current warning be an error instead? I could update the default severity to be an error, but it means every check is going to fail and prevent merging until we update the style of the code.

### DIFF
--- a/linting_java_checkstyle.xml
+++ b/linting_java_checkstyle.xml
@@ -1,0 +1,451 @@
+<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC
+  "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
+  "https://checkstyle.org/dtds/configuration_1_3.dtd">
+
+<!--
+    Custom checkstyle configuration forked from the google_checks.xml within
+    checkstyle 10.21.2
+
+    Checkstyle is very configurable. Be sure to read the documentation at
+    http://checkstyle.org (or in your downloaded distribution).
+
+    To completely disable a check, just comment it out or delete it from the file.
+    To suppress certain violations please review suppression filters.
+
+    Original Authors: Max Vetrenko, Mauryan Kansara, Ruslan Diachenko, Roman Ivanov.
+ -->
+
+<module name="Checker">
+
+  <property name="charset" value="UTF-8"/>
+
+  <property name="severity" value="${org.checkstyle.google.severity}" default="warning"/>
+
+  <property name="fileExtensions" value="java, properties, xml"/>
+  <!-- Excludes all 'module-info.java' files              -->
+  <!-- See https://checkstyle.org/filefilters/index.html -->
+  <module name="BeforeExecutionExclusionFileFilter">
+    <property name="fileNamePattern" value="module\-info\.java$"/>
+  </module>
+
+  <module name="SuppressWarningsFilter"/>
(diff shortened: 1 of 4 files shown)


andylassiter:
I agree with these changes. They align with how I code and strike a good balance. I don't think we should be overly aggressive.

cmoore01:
See #32

PR #20, opened 2025-02-06
SCOUT-32 Ansible IaC

kathrynalpert:
# SCOUT-32 Ansible IaC

## Type of change
- [x] New feature
- [ ] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

### Product
Once nodes have been provisioned, this feature attempts to facilitate a one-command installation of the Scout stack (really, one command per node, but this will be revisited when we add support for multinode k3s cluster).

### Technical
* Setup Ansible playbooks (and a Makefile to run them all in one go)
* Update existing IaC to play nicely with the Ansible config
* Add minio IaC

##### Authorization
Ansible playbooks will run as root, locally on each node.

##### Appsec
Much of the IaC is depending on cross-service activity being in-cluster and all external-facing services (LoadBalancer) being closed off to just our Tailnet. We will need to review more carefully as we relax these requirements.

I didn't research if injecting bad values into inventory.yaml can create issues. I would say this is largely the responsibility of the infra team running the playbooks. We provide `inventory.example.yaml`; it is up to the team running Ansible to populate their own `inventory.yaml`.

### Performance
N/A for the most part. No changes to existing IaC wrt performance.

### Data
N/A, Ansible seems to take care of input validation.

### Backward compatibility
I believe previous helm install commands will still work, but ideally folks are provisioning clusters using Ansible from here on out.

## Testing
Running `make all` after uninstalling k3s, visiting services via port forwarding. 
* Ensure our Temporal workflow (ingest-hl7-log) runs through to completion and creates appropriate objects in MinIO
* Ensure MinIO console is up and functional
* Ensure Grafana dashboards are up and functional
* Ensure Grafana alerting works
* View Loki logs through Grafana
* Check prometheus UI, ensure all connections are green

## Note for reviewers
Subsequent efforts will include 
* Updates to GHA to use Ansible for testing
* Extending this to set up a multinode k3s cluster.

## Checklist


helm/minio/values.yaml
```diff
@@ -0,0 +1,524 @@
...
+    - name: loki-chunks
+    - name: loki-ruler
+    - name: loki-admin
+    - name: lake
+
+  ###
+  # Array of Kubernetes secrets from which the Operator generates MinIO users during tenant provisioning.
+  #
+  # Each secret should specify the ``CONSOLE_ACCESS_KEY`` and ``CONSOLE_SECRET_KEY`` as the access key and secret key for that user.
+  users: []
```
johnflavin, line 307: This doesn't need to be done now, but a note for the future: it would be a good practice to create separate minio user accounts for all the services with corresponding limited ACLs.

ansible/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+It is assumed that you have provisioned a node with access to a sufficiently large "local" disk
+before running Ansible. We will use Rancher's local path provisioner for provisioning
+our persistent volumes. You may modify `../helm/storageclass.yaml` to change this.
+
+Start by copying `inventory.example.yaml` to `inventory.yaml`. You will want to replace the
+`vars` section with the values that are appropriate for your installation. Note that `helm_plugins_dir` and `kubeconfig_yaml`
+are default locations that you probably want to leave alone. 
+
+Some of these values in `inventory.yaml` should be treated as secrets, which you can manage with Ansible vault.
```
johnflavin, line 15: This password section is confusing. Is ansible generating random passwords every time we run it, or is it storing them in a file? 

> For example, if you put a password (or a script that returns a password in `vault/get_pwd.sh`, then you can run, for example:
> 
> ```bash
> openssl rand -hex 32 | ansible-vault encrypt_string --vault-password-file /vault/pwd.sh
> ```

One line says the password file is `vault/get_pwd.sh`, but the command after it says `/vault/pwd.sh`. Leaving aside the file name difference, one is a local path and the other absolute. Which should we use? Are we supposed to pre-create a script which returns passwords, or is that script generated by ansible when we run this `ansible-vault` command? If we're making it, what is the format?

> ...which will give you a string that you can then place in the `inventory.yaml` as the value for the appropriate var.

Does this mean we are manually storing passwords? Or is `ansible-vault` managing that for us?

johnflavin (changes requested):

helm/orchestration/temporal-worker/templates/deployment.yaml
```diff
@@ -42,10 +42,13 @@ spec:
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
-          {{- with .Values.volumeMounts }}
           volumeMounts:
+          {{- with .Values.volumeMounts }}
```
johnflavin, line 46: I don't think this is right. The `volumeMounts` should be inside the `{{- with .Values.volumeMounts }}` so that if there aren't any `volumeMounts` in the values file, this section doesn't appear. 

The `config-volume` that gets added below should not be here, it should be in the `temporal-java` values file.

The reason being that, since the `temporal-worker` chart is shared between the java and python workers, as it is the python worker mounts in an `application.yaml` that it does not need or use.

helm/orchestration/temporal-worker/templates/deployment.yaml
```diff
@@ -54,10 +57,13 @@ spec:
           envFrom:
             {{- toYaml . | nindent 12 }}
           {{- end }}
-      {{- with .Values.volumes }}
       volumes:
+      {{- with .Values.volumes }}
         {{- toYaml . | nindent 8 }}
       {{- end }}
+        - name: config-volume
```
johnflavin, line 64: Same comment as the mounts. This should be in the java worker values file.

helm/orchestration/temporal-worker/values.yaml
```diff
@@ -95,3 +95,21 @@ nodeSelector: {}
 tolerations: []
 
 affinity: {}
+
+config:
```
johnflavin, line 99: I don't think this config should go in the chart values file because here it can't be templatized. Like, this has hard-coded names and namespaces for temporal and minio services. Those values depend on the deploy.

(To be totally fair, this is still an improvement over the current state where the values are hard-coded into the application.)

johnflavin (changes requested):

helm/orchestration/temporal-worker/values.yaml
kathrynalpert: They can be templatized, which is what I'm doing in line 99 of the ansible/playbooks/orchestrator.yaml.

helm/orchestration/temporal-worker/templates/deployment.yaml
kathrynalpert: The reason for this change is that the following lines (mounting the config) are always there, so unlike before, we always have something in the `volumeMounts` section.

I agree that it's sort of clunky to mount the application.yaml into both java and python workers, but that may be telling us it is somewhat improper for them to share a chart at all. I did initially try putting this in the java values file, but then I didn't have access to `{{ include "temporal-worker.fullname" . }}`, which felt a little clunky to pull up into the values file. In sum, I think the better approach may be two different charts, but I felt that was a little out of scope for this changeset.

helm/orchestration/temporal-worker/templates/deployment.yaml
kathrynalpert: Same comment as the mounts, let's track it there.

helm/orchestration/temporal-worker/templates/deployment.yaml
johnflavin: I agree. It is pointing to the conclusion that these should be two charts. And also that it is out of scope for this change.

helm/orchestration/temporal-worker/values.yaml
johnflavin: I see, I didn’t put together that you had added this here but were more or less overwriting it from ansible.

johnflavin (approved):

helm/minio/values.yaml
kathrynalpert: https://xnat.atlassian.net/browse/SCOUT-38

helm/orchestration/temporal-worker/templates/deployment.yaml
kathrynalpert: https://xnat.atlassian.net/browse/SCOUT-41

PR #22, opened 2025-02-13
SCOUT-39 Multinode k3s

kathrynalpert:
# SCOUT-39 Multinode k3s

## Type of change
- [x] New feature
- [ ] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

### Product
Update the Ansible playbooks to install a k3s multinode cluster, versus only supporting a single, control-plane node.

### Technical
Add a new playbook for agents that uses the k3s install script with `K3S_URL` and `K3S_TOKEN` environment variables so that they will join an existing cluster, versus creating their own. This was necessary because the k3s role we are using for our control node install expects that Ansible is run across the whole cluster, whereas we are running Ansible independently on each node.

The agent nodes wait for the server to be available, so I didn't need to add that logic. 

I am not setting any node selectors, affinities, taints, or tolerances, which means that currently, if the server is spun up first, no pods would be scheduled on the workers (unless autoscaling meant new pods were added after install). This will be a next step and subsequent PR.

##### Authorization
These changes allow the runner of the Ansible playbooks to specify a unix group that will be permitted to execute `kubectl` commands on the server. This way, engineering team members will not need root access in order to manage the k8s cluster.

##### Appsec
I didn't have any particular appsec considerations for this change.

### Performance
This should improve performance in that we can distribute work across the cluster.

### Data
N/A

### Backward compatibility
N/A this stuff isn't in use yet

## Testing
Manually tested by building a cluster of big-01 (server) and big-02 (agent). After updating firewall settings per [k3s docs](https://docs.k3s.io/installation/requirements#networking), cluster worked as expected, including:
* Temporal workflow executed
* Temporal web UI
* Grafana dashboards, logs, and alerts
* Prometheus PromQL queries from web UI
* Minio web UI

Ensured that playbooks executed on the proper node (e.g., agents ran k3s join only, server ran everything else)

## Note for reviewers
Subsequent efforts:
* Updates to helm charts to specify which pods run where

Commit list: because I used a subtree push, there's a dupe commit message. I'll squash merge and remove that so we don't have a messy history.

## Checklist


ansible/inventory.example.yaml
```diff
@@ -3,26 +3,25 @@ k3s_cluster:
   children:
     server:
       hosts:
-        leader:
+        FQDN-leader.edu:
```
johnflavin, line 6: If the nodes don't have an FQDN, can you still use the previous style, which seems to be an internal name and an IP? I don't think this is something we have to deal with now, just thinking about customizing this in the future.

johnflavin (approved):

andylassiter (approved):

ansible/inventory.example.yaml
kathrynalpert: The only constraint here is that the agents need to be able to talk to the worker at this address.

PR #25, opened 2025-02-18
SCOUT-40 Full Ansible install and integration tests in CI

cmoore01:
# SCOUT-40 Full Ansible install and integration tests in CI

## Type of change
- [X] New feature
- [ ] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [X] Test update

### Product
N/A

### Technical
The ansible playbooks are executed with a few tweaks for the ci use case. This slots into the existing placeholder in the full ci workflow built out previously.

##### Authorization
N/A

##### Appsec
N/A

### Performance
N/A

### Data
N/A

### Backward compatibility
N/A

## Testing
We have a fully green, passing end-to-end github actions workflow from this :)

## Note for reviewers
This PR is not as complete as I would like it to be ideally. I am rushing it so that I can try it with Flavin's rewrite of the temporal code to hopefully avoid a race condition I was seeing in these tests. I am currently working around that by launching the first temporal job, waiting 30 seconds, and then launching the rest. That appears to work, but it would be much better to remove it after checking that Flavin's fix indeed works.

@kathrynalpert One thing I would like a double check on is the "clone scout" ansible playbook. That playbook is problematic here because it hardcodes the repo URL and branch. Considering I'm already downloading the repo, I replaced that playbook with a simple copy over to the place that playbook would clone to.

The parts I would like to improve later after evaluating Flavin's changes are:
1. More clever variable usage to keep `temporal_java_helm_values` and `temporal_python_helm_values` out of `ansible/inventory.example.yaml`
2. In `k3s-image-import` , use the derived image version rather than hardcoding `latest`.
3. Waiting on `temporal-schema` by label rather than name.

## Checklist
- [X] I have added or updated user documentation, if appropriate.


.github/ci_resources/temporal-java.values.yaml
```diff
@@ -0,0 +1,20 @@
...
+      name: s3-secret
+  - configMapRef:
+      name: s3-env
+env:
+  - name: TZ
+    value: 'US/Central'
+volumes:
+  - name: hl7logs
+    hostPath:
+      path: /var/lib/scout/tests/staging_test_data/hl7
```
kathrynalpert, line 16: We can use variables in ansible for the image repos & hl7 log path so you don't have to maintain these files

.github/workflows/ci.yaml
```diff
@@ -60,8 +60,76 @@ jobs:
...
+        run: cp .github/ci_resources/inventory.yaml ansible/
+      - name: 'Install prerequisites'
+        shell: bash
+        run: sudo pip3 install kubernetes
+      - name: 'ansible-galaxy dependencies'
+        shell: bash
+        run: cd ansible && sudo /opt/pipx_bin/ansible-galaxy install -r collections/requirements.yaml
+      - name: 'Copy scout'
+        shell: bash
+        run: sudo cp -r . /var/lib/scout # in CI workflow, we want to use current code, not "main" branch. TODO: don't hardcode path here
```
kathrynalpert, line 75: This seems reasonable to me. I can also make the branch variable, but it seems silly to clone something we have

.github/workflows/ci.yaml
```diff
@@ -60,8 +60,76 @@ jobs:
...
+      - name: 'Install k3s'
+        shell: bash
+        run: |
+          cd ansible && sudo /opt/pipx_bin/ansible-playbook -vvv -i inventory.yaml --diff playbooks/k3s-cluster.yaml
```
kathrynalpert, line 79: Do we need `-vvv` here?

kathrynalpert (approved):

.github/ci_resources/temporal-java.values.yaml
cmoore01: Good idea, I'll look into that for the next PR.

.github/workflows/ci.yaml
cmoore01: Probably not, I'll lower it in the next PR.

PR #29, opened 2025-02-19
SCOUT-40 Cleanup on CI workflow

cmoore01:
# SCOUT-40 Cleanup on CI workflow

## Type of change
- [ ] New feature
- [X] Improvement
- [X] Bug fix
- [ ] Documentation update
- [ ] Test update

### Product
CI tests now use Flavin's latest changes to submit one single temporal job for all dates rather than `n` different single-date jobs.

### Technical
1. The script to derive the docker version for an image from `docker-build-cache` was refactored into its own standalone action so that it could be reused when setting the version of the images in k3s. This removes the hardcoding of `latest` when pushing to k3s.
2. Adds some variables than _can_ be specified in ansible inventory to override HL7 mount dir and temporal-images. These are set in GHA for tests to use local images and the correct test data, but will default to their original values if unset. This removes the need for forked copies of the helm values files in the previous PR.
3. Fixes a bug where an empty array response from temporal for checking workflow statuses was considered "done" (caused by an "all" check being vacuously true).

##### Authorization
N/A

##### Appsec
N/A

### Performance
N/A

### Data
N/A

### Backward compatibility
N/A

## Testing
Full GHA CI workflow is passing

## Note for reviewers
For @kathrynalpert in particular, is this the cleanest way of accomplishing the image/log overrides with ansible? It doesn't feel clean that I had to duplicate the default value that already exists in the helm files.

## Checklist


kathrynalpert:
I'm not sure if this would work, but could you use `default(omit)` to have it fall back on the values file?

cmoore01:
> I'm not sure if this would work, but could you use `default(omit)` to have it fall back on the values file?

I don't think so. I tried this:
```
- value: 'volumes[0].hostPath.path={{ log_mount | default(omit) }}'
```
with the variable temporarily commented out, and the command coming out of ansible failed pretty badly with this getting generated:
```
--set volumes[0].hostPath.path=__omit_place_holder__2a6d8c1533a8261ee768dd55e7c6368a3ee5285b
```

It looks like it might be possible to accomplish with some conditional manipulation of the list itself, but I haven't found the right syntax yet.

johnflavin (approved):

PR #35, opened 2025-02-26
Set max attempts = 5 on java activities

johnflavin:
## Type of change
- [ ] New feature
- [ ] Improvement
- [X] Bug fix
- [ ] Documentation update
- [ ] Test update

### Product
Cap the number of times the java temporal worker retries its activities to 5. This should prevent a failed activity from retrying forever and hanging the workflow.

### Technical
In the java workflow implementation, set an activity option that caps at 5 the number of times the java activities will be attempted.
This is currently hard-coded. In the future we could consider reading this from the application properties which would make it adjustable at deploy time.

##### Authorization
N/A

##### Appsec
N/A

### Performance
This will hopefully have no performance impact, or perhaps positive in that an activity won't be able to retry and fail forever. But I haven't tested it.

### Data
This should improve edge case handling.

### Backward compatibility
N/A

## Testing
None

## Checklist
- [ ] My code adheres to the coding and style guidelines of the project.


johnflavin:
We will need to make other tweaks to activity timeouts and such. Will open a new PR when that is all ready.

PR #37, opened 2025-02-27
SCOUT-59 Set up persistence for Temporal dbs

kathrynalpert:
# SCOUT-59 Set up persistence for Temporal dbs

## Type of change
- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Documentation update
- [ ] Test update

### Product
Temporal data was not persisting across restarts, causing issues with temporal stack coming up properly after, say, a machine reboot.

### Technical
I add proper persistent volumes for Elastic and Cassandra.

##### Authorization
N/A

##### Appsec
N/A

### Performance
N/A

### Data
N/A

### Backward compatibility
N/A

## Testing
Bring up temporal, run a workflow. Restart the elastic pod, confirm temporal is still functional. Uninstall and reinstall temporal chart. Confirm no data loss.

## Note for reviewers
We should still move to Operators for these db services, but hopefully this gets us to a point where we can do more testing.

## Checklist


ansible/playbooks/orchestrator.yaml
```diff
@@ -37,28 +38,42 @@
           cassandra:
             config:
               cluster_size: 1
+            persistence:
+              enabled: true
+              storageClass: local-path-retain
```
johnflavin, line 43: (Not a problem, just something I don't understand:) Is this storage class a standard thing or do we need to define it in our playbooks? And does this mount a host path or autoprovision a pv in the regular way k3s does?

ansible/playbooks/orchestrator.yaml
```diff
@@ -37,28 +38,42 @@
...
+                value: /temporal
             ingress:
               enabled: true
-              ingressClassName: 'traefik'
+              ingressClassName: traefik
               hosts:
                 - '{{ server_hostname }}/temporal'
 
+    - name: Wait for Temporal schema job to complete
+      command: kubectl -n temporal wait --for=condition=complete --timeout=300s job/temporal-schema-1
```
johnflavin, line 66: Does this wait block other ansible actions from happening in parallel? I know we want to wait for temporal before declaring the whole application up and ready, but while temporal is coming up ansible could move on and tell other parts of the app to come up instead of just hanging around waiting.

johnflavin (approved):

ansible/playbooks/orchestrator.yaml
kathrynalpert: We [define it in our playbooks](https://github.com/washu-tag/scout/blob/main/ansible/playbooks/k3s.yaml#L95). See https://github.com/washu-tag/scout/blob/main/helm/storageclass.yaml. It uses the rancher local path provisioner, but sets reclaimPolicy to `Retain` instead of the default `Delete`.

PR #39, opened 2025-02-28
Rename HL7 column names, use Spark to read hl7 files, join modality map, extract patient ids, and write to deltalake

andylassiter:
# Rename HL7 column names, use Spark to read hl7 files, join modality map, extract patient ids, and write to deltalake

## Type of change
- [ ] New feature
- [x] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

### Product / Technical
- Makes HL7 column names more user friendly
- Switch to Spark for processing hl7 records

##### Authorization
N/A

##### Appsec
N/A

### Performance
- Switch to Spark should improve performance at scale.

### Data
- Errors will kill the workflow.

### Backward compatibility
N/A

## Testing
- Tested workflow successful for about a dozen dates. Have not tested at scale.

## Note for reviewers
N/A

## Checklist
- [x] I have added or updated technical documentation, including an architectural decision record, if appropriate.


helm/jupyter/notebook/requirements.txt
```diff
@@ -1,2 +1,2 @@
-deltalake
+delta-spark
 jupyterlab-git
```
andylassiter, line 2: @johnflavin Confirming we can replace deltalake with delta-spark.

orchestration/temporal-python/src/temporalpy/hl7extractor/deltalake.py
```diff
@@ -92,68 +113,170 @@ def read_hl7_input(hl7input: Iterable[str]) -> Iterator[MessageData]:
...
+        .config("spark.databricks.delta.merge.repartitionBeforeWrite.enabled", "true")
+        .config("spark.sql.extensions", "io.delta.sql.DeltaSparkSessionExtension")
+        .config(
+            "spark.sql.catalog.spark_catalog",
+            "org.apache.spark.sql.delta.catalog.DeltaCatalog",
+        )
+        .config("spark.hadoop.fs.s3a.access.key", s3a_access_key)
+        .config("spark.hadoop.fs.s3a.secret.key", s3a_secret_key)
+        .config("spark.hadoop.fs.s3a.endpoint", s3a_endpoint)
+        .config("spark.hadoop.fs.s3a.endpoint.region", s3a_region)
```
andylassiter, line 145: @johnflavin Anything else need to be passed from Ansible to Spark config?

helm/jupyter/notebook/requirements.txt
johnflavin: That's correct. Don't need to use the non-spark `deltalake` library anymore.

orchestration/temporal-python/src/temporalpy/hl7extractor/deltalake.py
johnflavin: I think this is sufficient. I'd prefer it to be written to a spark config file rather than pulled from the environment, but honestly this is fine for now and we can do a config file later.

johnflavin (commented):
What's here looks fine to me. But this will break the CI tests in a couple ways:
1. They pass a path to input data to the workflow based on the mount path, and we are changing the mount path here.
2. The tests run queries and check results using column names and those all got changed.

orchestration/temporal-python/src/temporalpy/hl7extractor/deltalake.py
andylassiter: Agree. I'd like to have that file to be shared with Jupyter, too.

.github/ci_resources/inventory.yaml
```diff
@@ -17,7 +17,7 @@ k3s_cluster:
     base_dir: /var/lib
     scout_repo_dir: WORK_DIR
     kubeconfig_group: 'root'
-    log_mount: '{{ scout_repo_dir }}/tests/staging_test_data/hl7'
+    orchestration_data_dir: '{{ scout_repo_dir }}/tests/staging_test_data'
```
kathrynalpert, line 20: Do we need to put a modality map spreadsheet [here](https://github.com/washu-tag/scout/tree/main/tests/staging_test_data/) for the python worker to use during the CI tests?

kathrynalpert (commented):
Pending ci test fixes

.github/ci_resources/inventory.yaml
johnflavin: Yes, we do need to do that. Charlie created one that is on another branch. Could cherry-pick that commit onto this branch to get it. https://github.com/washu-tag/scout/commit/0078b438017d75889bd30092fa2504d56272f1c4

orchestration/temporal-python/src/temporalpy/ingesthl7worker.py
```diff
@@ -50,13 +52,14 @@ async def main(argv=None):
         "TEMPORAL_ADDRESS", "temporal-frontend.temporal:7233"
     )
     temporal_namespace = os.environ.get("TEMPORAL_NAMESPACE", "default")
+    modality_map_path = os.environ.get("MODALITY_MAP_PATH", "/data/modalitymap.csv")
```
johnflavin, line 55: Another note about the modality map and CI. I added this parameter that is global to the python worker specifying where the modality map can be found. (I also added an activity-level input that can overrides this value for a particular workflow invocation, but currently the workflow itself doesn't have an input for this and doesn't pass a value to the activity. So this env is the only way to set this right now.)

So if we write the test modality map to `{{ scout_repo_dir }}/tests/staging_test_data/modalitymap.csv` then it will get picked up during CI testing. If we put it somewhere else, we'd need to set this env saying where.

.github/ci_resources/inventory.yaml
cmoore01: The `hl7` path is going to need to get into the temporal launch somehow, so if it gets taken out here, it will need to be added in the launch script. The temporal workflow does (doesn't?) seem to support arbitrary depth search for the log files: it either has to be directly in that directory or 1 level deep within a year subdir.

andylassiter: The [Launch temporal ingestion test is failing](https://github.com/washu-tag/scout/actions/runs/13595793946/job/38012570543?pr=39) with error `No log files found`, I presume this is why.

tests/src/test/resources/spark_queries.json
```diff
@@ -37,13 +37,13 @@
...
           "description": "patient born after 1990-12-31",
-          "columnName": "pid_7_date_time_of_birth",
+          "columnName": "birth_date",
           "truncation": 8,
           "comparisonValue": 19901231,
           "comparisonOperator": "GT"
         }
       },
-      "sql": "SELECT * FROM syntheticdata WHERE SUBSTRING(pid_7_date_time_of_birth, 1, 8) > '19901231'",
+      "sql": "SELECT * FROM syntheticdata WHERE SUBSTRING(birth_date, 1, 8) > '19901231'",
```
johnflavin, line 46: You might need to change this query more substantially. `birth_date` is a proper SQL date type now so we don't need to do any substring operations on it.

tests/src/test/resources/spark_queries.json
```diff
@@ -54,7 +54,7 @@
           "type": ".FixedColumnsValidator",
           "fixedColumns": {
             "orc_2_placer_order_number": [null],
```
johnflavin, line 56: Missed this col

andylassiter: Fixed

johnflavin:
Everything looks good. Ready for review. (Though I think @kathrynalpert might have to approve since everyone else contributed on this branch.)

orchestration/temporal-python/src/temporalpy/hl7extractor/deltalake.py
```diff
@@ -92,68 +124,169 @@ def read_hl7_input(hl7input: Iterable[str]) -> Iterator[MessageData]:
...
 
-    df = pd.DataFrame.from_records(
-        asdict(message) for message in read_hl7_input(hl7_input) if message is not None
-    ).astype(dtype="string[pyarrow]")
-
-    log.info(f"Extracted data from {len(df)} HL7 messages")
+    s3a_endpoint = os.environ.get("AWS_ENDPOINT_URL")
+    s3a_access_key = os.environ.get("AWS_ACCESS_KEY_ID")
+    s3a_secret_key = os.environ.get("AWS_SECRET_ACCESS_KEY")
+    s3a_region = os.environ.get("AWS_REGION", "us-east-1")
```
kathrynalpert, line 137: Ok for now, but this should be a variable, too - we pass it through Ansible.

orchestration/temporal-python/src/temporalpy/hl7extractor/deltalake.py
```diff
@@ -92,68 +124,169 @@ def read_hl7_input(hl7input: Iterable[str]) -> Iterator[MessageData]:
...
+        .withColumn("year", F.year("message_dt"))
+        .withColumn("updated", F.current_timestamp())
+        .join(patient_id_df, "message_control_id", "left")
+    )
+
+    # Create table if it doesn't yet exist
+    delta_table = delta_table.replace("s3://", "s3a://")
+
+    dt = (
+        DeltaTable.createIfNotExists(spark)
```
kathrynalpert, line 272: Here we still have a race condition, right?

kathrynalpert (approved):

orchestration/temporal-python/src/temporalpy/hl7extractor/deltalake.py
johnflavin: With delta lake on S3 (or compatible), we will always have race conditions unless we limit to a single writer at a time.

orchestration/temporal-python/src/temporalpy/hl7extractor/deltalake.py
johnflavin: I'll make that change in my follow-up branch.

PR #46, opened 2025-03-12
SCOUT-61 Continue as new

johnflavin:
## Type of change
- [ ] New feature
- [X] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

### Product
Large ingest workflows will be split into smaller batches.

### Technical
The ingest workflow will ingest only a certain (configurable) number of log files before starting a new workflow using the Continue-As-New mechanism which will continue processing the rest.

##### Authorization
N/A

##### Appsec
N/A

### Performance
The intent of this change is to enable the temporal ingest workflow to run larger ingests without running into temporal limits. Performance should be similar to before, except that what would have been a single large workflow before will be batched into smaller workflows.

### Data
N/A

### Backward compatibility
N/A

## Testing
I set an artificially low limit on the number of child workflows (50) and ran the test dataset. Verified that it correctly chunks the log files and continues a new workflow.

I ran other data sets as well, and they seemed to also work fine, but they took much longer as I'm running tests locally and performance is an issue.

## Note for reviewers
N/A

## Checklist

### DIFF
--- a/orchestration/temporal-java/src/main/java/edu/washu/tag/temporal/activity/FindHl7LogsActivityImpl.java
+++ b/orchestration/temporal-java/src/main/java/edu/washu/tag/temporal/activity/FindHl7LogsActivityImpl.java
@@ -112,8 +112,13 @@ public class FindHl7LogsActivityImpl implements FindHl7LogsActivity {
         // Do we continue again or are we at the end?
         ContinueIngestWorkflow continued = null;
         int nextIndex = input.nextIndex() + maxChildWorkflows;
-        if (nextIndex < logFiles.size()) {
-            continued = new ContinueIngestWorkflow(manifestFilePath, logFiles.size(), nextIndex);
+        if (nextIndex < input.numLogFiles()) {
+            logger.info("WorkflowId {} ActivityId {} - We will continue the workflow as new starting with index {}",
+                activityInfo.getWorkflowId(), activityInfo.getActivityId(), nextIndex);
+            continued = new ContinueIngestWorkflow(manifestFilePath, input.numLogFiles(), nextIndex);
+        } else {
+            logger.info("WorkflowId {} ActivityId {} - nextIndex {} >= numLogFiles {}. This batch is the end of the log files.",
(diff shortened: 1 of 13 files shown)


orchestration/temporal-java/src/main/java/edu/washu/tag/temporal/activity/FindHl7LogsActivityImpl.java
```diff
@@ -46,7 +63,63 @@ public FindHl7LogFileOutput findHl7LogFiles(FindHl7LogFileInput input) {
...
+        // Get the next batch of log files to process
+        List<String> logFiles = Arrays.stream(manifest.split("\n"))
+            .skip(input.nextIndex())  // Skip forward to the next index
+            .limit(maxChildWorkflows) // Return the next maxChildWorkflows files to process
+            .toList();
+
+        // Do we continue again or are we at the end?
+        ContinueIngestWorkflow continued = null;
+        int nextIndex = input.nextIndex() + maxChildWorkflows;
+        if (nextIndex < input.numLogFiles()) {
```
kathrynalpert, line 114: This reads correctly to me, but did you confirm we're not off by one?

kathrynalpert (approved):

orchestration/temporal-java/src/main/java/edu/washu/tag/temporal/activity/FindHl7LogsActivityImpl.java
johnflavin: I didn't do anything to directly confirm we aren't off by one. I'm not really sure how I would do that. Maybe this is where unit testing our activities would come in.

I'm confident this is correct but that doesn't mean a whole lot.

kathrynalpert: Yeah, we could unit test or just manually run with , e.g., 11 dates and max size 10. If you are confident and it looks right to me, that's probably good enough, though.

kathrynalpert: I confirmed it - we're good!

PR #55, opened 2025-03-27
Split ingest workflow into two: split hl7 log to hl7 files and ingest hl7 files to delta lake

johnflavin:
## Type of change
- [ ] New feature
- [X] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

### Technical
Created a new temporal workflow (java sdk) that takes a manifest file of HL7 file paths (or an s3 directory which will be recursively search for HL7 file paths) and runs the activity where spark ingests them into delta lake.

### Performance
The performance of the two-workflow system should be similar to the single-workflow system, but it should enable better developer experience. If the HL7 transformation and upload works fine but the spark ingest fails, now we can launch the second workflow again using the same manifest file and not have to split and transform the HL7 logs again.

### Data
Workflow's data handling is essentially unchanged

### Backward compatibility
N/A

## Testing
Manually tested on big-03.
- Ran the workflow through from the beginning with an HL7 log
- Ran just the spark ingest workflow using the manifest file generated from the previous two-workflow run
- Ran another spark ingest workflow run using one directory of HL7 files

## Checklist


orchestration/temporal-python/src/temporalpy/ingesthl7worker.py
```diff
@@ -20,15 +21,20 @@ async def run_worker(
...
         worker = Worker(
             client,
             task_queue=TASK_QUEUE_NAME,
-            activities=[ingest_hl7_files_activity_wrapper(mapping_file_path)],
+            activities=[ingest_hl7_files_activity.ingest_hl7_files_to_delta_lake],
             activity_executor=pool,
+            shared_state_manager=SharedStateManager.create_from_multiprocessing(
+                multiprocessing.Manager()
+            ),
+            max_concurrent_activities=1,
```
andylassiter, line 34: I tested with:
```
  max_cached_workflows=1,
  max_concurrent_workflow_tasks=1,
  max_concurrent_workflow_task_polls=1,
  max_concurrent_activities=1,
  max_concurrent_activity_task_polls=1,
```
And the memory usage is now resetting between activities instead of growing endlessly.

andylassiter (approved):
Do you have an example `ktemporal` command you could share for running the workflows separately?

orchestration/temporal-python/src/temporalpy/ingesthl7worker.py
johnflavin: Do you want me to make those changes here or would you rather put them in a separate PR?

andylassiter: Include `max_concurrent_activity_task_polls=1`, the default was 5 and there shouldn't be more pollers than activities. I'll include the others in a separate PR.

PR #59, opened 2025-04-01
Fix Single-User Jupyter Notebook Pod PV/PVC Binding Issue

andylassiter:
# Fix Single-User Jupyter Notebook Pod PV/PVC Binding Issue

## Type of change
- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Documentation update
- [ ] Test update

### Product
- Fix issue with single-user notebook storage.

### Technical
When [updating storage classes and PVs](https://github.com/washu-tag/scout/pull/50), the single-user pods are unable to bind to the pv/pvc shared with the Hub. But single-user pods probably shouldn't share a pv/pvcC with the Hub. This gives the hub one pv/pvc and all single-user pods share a pv/pvc.

### Backward compatibility
Fixes issue introduced in https://github.com/washu-tag/scout/pull/50. The behavior is the same from a user perspective, but instead of each single-user pod getting it's own pv/pvc, all single-user pods share a pv/pvc. I tried to setup a rancher/local-path-provisioner specifically for Jupyter pods but was not able to do so quickly. This unblocks Jupyter and is okay for now

## Testing
Tested on big-02. I was able to log out/in and my notebooks were persisted.

## Checklist

### DIFF
--- a/ansible/playbooks/jupyter.yaml
+++ b/ansible/playbooks/jupyter.yaml
@@ -10,34 +10,17 @@
       include_tasks: tasks/storage_setup.yaml
       vars:
         storage_definitions:
-          - name: jupyter
+          - name: jupyter-hub
             size: '15Gi'
             path: '{{ jupyter_dir }}/hub'
 
-    - name: Create single-user storage directory
-      file:
-        path: '{{ jupyter_dir }}/singleuser'
-        state: directory
-        mode: '0777'
-
-    - name: Create PersistentVolume for single-user storage
-      kubernetes.core.k8s:
-        state: present
-        definition:
-          apiVersion: v1
-          kind: PersistentVolume
-          metadata:
-            name: jupyter-singleuser-pv
-            namespace: jupyter
-          spec:
-            capacity:
-              storage: 250Gi
-            accessModes:
-              - ReadWriteOnce
-            persistentVolumeReclaimPolicy: Retain
-            storageClassName: manual
-            hostPath:
-              path: '{{ jupyter_dir }}/singleuser'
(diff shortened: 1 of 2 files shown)


johnflavin (commented):
I'm not sure about this approach. Why define the PV and PVC manually? Seems like you could have fewer changes and lean into the existing patterns by just adding another item to the `storage_definitions` list in the `Setup storage` task to define a notebook PV and storage class, and in `hub_values.yaml` set `storage.dynamic.storageClass` to whatever the name of the storage class would be. (Probably also need to set `hub.dynamic.storageClass` to `jupyter-storage` to keep the hub and notebook PVs separate.)

andylassiter:
The existing pattern doesn't quite fit for the singleuser PV and PVCs. JH will create claims named `claim-user1`, `claim-user2`, etc. But once a user's claim binds to the volume, no other claim can bind to the volume. The hubs `dynamic` storage can create PVCs but not new PVs. Rancher had previously been deploying the needed PVs. To use the hubs `static` option we need to manually create the PVC, the helm chart won't accept a storage class for `static` storage. I updated the PR to use the existing `Setup Storage` task to create the PV. The storage class it creates isn't needed but it cuts down on code changes. We do still need to create the PVC.

johnflavin (commented):
> Rancher had previously been deploying the needed PVs

Did something change to make this no longer the case?

andylassiter:
> > Rancher had previously been deploying the needed PVs
> 
> Did something change to make this no longer the case?

https://github.com/washu-tag/scout/pull/50

PR #60, opened 2025-04-02
Add python worker health check for spark

johnflavin:
## Type of change
- [X] New feature
- [ ] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

## Description

This adds a health check to the python worker, so that if spark dies or whatever we can signal to kubernetes that it needs to restart the pod.

There are two main pieces
1. Add a small FastAPI server with a `/healthz` API endpoint. When the API is hit it checks the contents of a temp file; if there is anything in the file it reports that the pod is unhealthy. We configure the helm chart so kubernetes will periodically check this API for health.
2. In the main worker code, if certain spark-y exceptions are caught, write the error message to the health check temp file.

##### Authorization
N/A

##### Appsec
I don't think there are any security implications here.
- I did initially hard-code the path to the temp health check file, but I thought that might open some kind of attack (though I'm not sure how it would actually work). I switched to using the `tempfile.mkstemp` function ([docs](https://docs.python.org/3.10/library/tempfile.html#tempfile.mkstemp)) to create the file, which should be more secure.
- The `/healthz` API is open, but it is a fairly cheap operation.

### Performance
N/A

### Data
N/A

### Backward compatibility
N/A

## Testing
I deployed this to a local k8s cluster and manually tested a couple scenarios.
- Manually write a message to the health check temp file (using a dev API that I did not check in)
- While a workflow was running, manually exec to the container and kill the spark process
In both cases the `/healthz` endpoint did report the pod was unhealthy and it was restarted.

I haven't been able to reproduce the error case where spark is OOM killed, but I suspect it would be similar to my "kill spark" test case.

## Note for reviewers
N/A

## Checklist

### DIFF
--- a/orchestration/temporal-python/src/temporalpy/healthapi.py
+++ b/orchestration/temporal-python/src/temporalpy/healthapi.py
@@ -30,12 +30,13 @@ async def healthz():
 
         # Read the contents of the temporary file
         contents = SPARK_HEALTH_TEMP_FILE.read_text()
-        if contents != HEALTHY:
-            log.warning("Health check file reports failure: %s", contents)
-            return unhealthy_json_response(f"Spark reported status: {contents}")
+        if not contents:
+            # No health information available. Assume healthy.
+            log.debug("No health information available (assume healthy)")
+            return HEALTHY_JSON_RESPONSE
 
-        log.info("Health check succeeded")
-        return HEALTHY_JSON_RESPONSE
+        log.warning('Health check file reports failure: "%s"', contents)
+        return unhealthy_json_response(f"Spark reported status: {contents}")
     except Exception as e:
         log.error("Health check failed with exception", e)
         return unhealthy_json_response(str(e))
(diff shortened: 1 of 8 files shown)


orchestration/temporal-python/src/temporalpy/healthapi.py
```diff
@@ -0,0 +1,59 @@
...
+        log.error("Health check failed with exception", e)
+        return unhealthy_json_response(str(e))
+
+
+def unhealthy_json_response(
+    reason: str, messages: list[str] | None = None
+) -> JSONResponse:
+    return JSONResponse(
+        status_code=503,
+        content={"status": UNHEALTHY, "reason": reason, "messages": messages or []},
```
cmoore01, line 52: Code scanning is complaining about this line. I know you said in the description that the API is open, but by whom and when would it be reachable? Could there be anything sensitive leaking into it?

johnflavin: The API is accessible to anyone with internal access to the cluster and its services, as far as I know. It isn't exposed out.

As far as the contents, the `reason` and `messages` here contain whatever was just read from the health check temp file. The only thing that gets written there are exception messages. I'm pretty confident (though not 100% sure) that spark doesn't put anything from the data it is processing into its error messages.

cmoore01 (approved):

PR #66, opened 2025-04-10
Remove stack trace from REST API response

kathrynalpert:
# Remove stack trace from REST API response

## Type of change
- [ ] New feature
- [x] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

### Product
Python worker health API 500 error due to exception in health check method itself returns generic error message instead of stack trace.

### Technical
Log stack trace rather than passing it through to user

##### Authorization
N/A

##### Appsec
Should improve security posture by not sharing stack trace info with end user.

### Performance
N/A

### Data
N/A

### Backward compatibility
N/A

## Testing
No testing, this is a blind code change.

## Checklist

### DIFF
--- a/orchestration/temporal-python/src/temporalpy/healthapi.py
+++ b/orchestration/temporal-python/src/temporalpy/healthapi.py
@@ -14,7 +14,7 @@ SPARK_HEALTH_TEMP_FILE = Path(tempfile.mkstemp()[1])
 HEALTHY = "healthy"
 UNHEALTHY = "unhealthy"
 HEALTHY_JSON_RESPONSE = JSONResponse(status_code=200, content={"status": HEALTHY})
-
+HEALTH_CHECK_EXCEPTION_MESSAGE = "Health check failed with exception"
 
 app = FastAPI()
 
@@ -40,8 +40,8 @@ async def healthz():
         log.warning('Health check file reports failure: "%s"', reason)
         return unhealthy_json_response(reason, messages)
     except Exception as e:
-        log.exception(e)
-        return unhealthy_json_response("Health check failed with exception")
+        log.exception(HEALTH_CHECK_EXCEPTION_MESSAGE)
+        return unhealthy_json_response(HEALTH_CHECK_EXCEPTION_MESSAGE)
 
 
 def unhealthy_json_response(
--- a/orchestration/temporal-python/src/temporalpy/healthapi.py
+++ b/orchestration/temporal-python/src/temporalpy/healthapi.py
@@ -40,8 +40,8 @@ async def healthz():
         log.warning('Health check file reports failure: "%s"', reason)
         return unhealthy_json_response(reason, messages)
     except Exception as e:
-        log.error("Health check failed with exception", e)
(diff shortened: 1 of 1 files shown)


orchestration/temporal-python/src/temporalpy/healthapi.py
```diff
@@ -40,8 +40,8 @@ async def healthz():
         log.warning('Health check file reports failure: "%s"', reason)
         return unhealthy_json_response(reason, messages)
     except Exception as e:
-        log.error("Health check failed with exception", e)
-        return unhealthy_json_response(str(e))
+        log.exception(e)
```
johnflavin, line 43: I'd prefer to keep both the message and the stack trace in the log message. I just did it wrong before. In java you just stick the exception object into the logger and it knows what to do, but in python you use the keyword arg `exc_info=e` to pass the exception.
```
log.exception("Health check failed with exception", exc_info=e)
```

kathrynalpert: You'll still get the actual exception message here, plus the stack trace, but if you want to include the string we control, we'd need to do:
`log.exception("Health check failed with exception: %s", str(e))`, which I thought was a little gross (and my sense was that we'd be able to find this issue pretty easily in the logs without needing to grep for a particular string... we could even just search by the healthapi file name itself).

kathrynalpert: Actually, maybe I can just do `log.exception("Health check failed with exception")`

kathrynalpert: Java coders in python :) So, yeah, it looks like the stack trace prints the error message (at the bottom), so we can include a custom message and get both. 
`log.exception("Health check failed with exception")` seems to be the same as `log.error("Health check failed with exception", exc_info=True)` where [sys.exc_info()](https://docs.python.org/3/library/sys.html#sys.exc_info) is called to get the exception info.

johnflavin: But if you already have an exception object, you can just pass that as the `exc_info` value to skip calling `sys.exc_info()`.

From the [logging docs](https://docs.python.org/3/library/logging.html#logging.Logger.debug)
> If `exc_info` does not evaluate as false, it causes exception information to be added to the logging message. If an exception tuple (in the format returned by [sys.exc_info()](https://docs.python.org/3/library/sys.html#sys.exc_info)) _**or an exception instance**_ is provided, it is used; otherwise, [sys.exc_info()](https://docs.python.org/3/library/sys.html#sys.exc_info) is called to get the exception information.
(Emphasis mine)

kathrynalpert: Right, but when I use the `log.exception` convenience method, I don't have to explicitly do either. I debated whether passing the `e` and using `log.error("Health check failed with exception", exc_info=e)` would be more efficient, but they made the method for a reason I assume and this is so non-consequential either way.

johnflavin (approved):

PR #71, opened 2025-04-16
Add postgres db for ingest file records

johnflavin:
## Type of change
- [X] New feature
- [ ] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

### Product
Write log file and HL7 file status information to a database as they are processed, which enables us to implement a processing dashboard (yet to be done).

### Technical
- Add an ansible task to deploy a postgres db, and include the task in the orchestrator install. (Note that I'm not sure this is still the right way to install postgres if we're going to use it for other things like superset.)
- Add a barebones db service to the temporal java worker
- Use that db service to write success/failure records for all the log files and HL7 files that are processed through the ingest workflow
- Use flyway to configure db tables (I'm excited about this)

### Performance
N/A

### Data
No new data processing considerations, but this should make it easier to detect when data problems arise.

### Backward compatibility
No backward compatibility changes. But note that we are adding a new db schema, which will mean we need to be mindful of future backward incompatible changes to that data. Luckily I think flyway will take care of schema migrations.

## Testing
Mostly GH CI testing

## Note for reviewers
- This doesn't currently include any error handling / db writing from the pyspark part of the ingest. I intend to follow this up with that. 
- ~~Keeping this in draft mode for now because there is an error with the postgres PV/PVCs on big-03. I think they're not being provisioned correctly.~~ Fixed. This was an error redeploying the ansible playbook with a different namespace without removing the old namespace, and PVCs were conflicting.

## Checklist


.github/ci_resources/inventory.yaml
```diff
@@ -14,22 +14,27 @@ k3s_cluster:
...
-    base_dir: /var/lib
-    scout_repo_dir: WORK_DIR
     kubeconfig_group: 'root'
-    orchestration_data_dir: '{{ scout_repo_dir }}/tests/staging_test_data'
     temporal_java_image: 'docker.io/local/washu-tag/temporal-java'
     temporal_python_image: 'docker.io/local/washu-tag/temporal-python'
     explorer_image: 'docker.io/local/washu-tag/explorer'
+    postgres_user: 'scout'
+    postgres_password: 'scout123'
+    postgres_cluster_namespace: 'cloudnative-pg'
```
kathrynalpert, line 23: I had moved this https://github.com/washu-tag/scout/blob/postgres/ansible/playbooks/orchestrator.yaml#L162 so you didn't have to specify the namespace in the inventory.

ansible/inventory.example.yaml
```diff
@@ -25,13 +25,17 @@ k3s_cluster:
     github_client_secret: $(echo $CLIENT_SECRET | ansible-vault encrypt_string --vault-password-file vault/pwd.sh)
     github_organization: 'org'
     jupyter_metrics_api_token: $(openssl rand -hex 32  | ansible-vault encrypt_string --vault-password-file vault/pwd.sh)
+    postgres_user: 'scout'
+    postgres_password: $(openssl rand -hex 32 | ansible-vault encrypt_string --vault-password-file vault/pwd.sh)
+    postgres_cluster_namespace: 'cloudnative-pg'
```
kathrynalpert, line 30: Same comment, maybe just merge in the latest changes from the `postgres` branch

ansible/playbooks/tasks/postgres.yaml
```diff
@@ -0,0 +1,72 @@
+- name: Add Postgres Helm repository
+  kubernetes.core.helm_repository:
+    name: cnpg
+    repo_url: https://raw.githubusercontent.com/cloudnative-pg/charts/refs/heads/gh-pages/
+
+- name: Install/Upgrade Postgres Operator
+  kubernetes.core.helm:
+    name: cnpg
+    chart_ref: cnpg/cloudnative-pg
+    chart_version: 0.23.2
```
kathrynalpert, line 10: I was hesitant to allow version "fuzzy matching" here bc of the implications of a db version upgrade, but maybe allowing patch updates would be ok?

orchestration/temporal-java/src/main/java/edu/washu/tag/temporal/activity/SplitHl7LogActivityImpl.java
```diff
@@ -60,20 +68,32 @@ public SplitAndTransformHl7LogOutput splitAndTransformHl7Log(SplitAndTransformHl
...
         URI destination = URI.create(input.hl7OutputPath());
-        List<String> hl7Paths;
+        List<Hl7File> segmentResults;
         try {
-            hl7Paths = processLogFile(input.logPath(), destination);
+            segmentResults = processLogFile(input.logPath(), destination, activityInfo.getWorkflowId(), activityInfo.getActivityId());
+
+            ingestDbService.insertLogFile(LogFile.success(input.logPath(), activityInfo.getWorkflowId(), activityInfo.getActivityId()));
         } catch (IOException e) {
+            ingestDbService.insertLogFile(LogFile.error(input.logPath(), e.getMessage(), activityInfo.getWorkflowId(), activityInfo.getActivityId()));
```
kathrynalpert, line 77: Do we want to log stack trace here?

orchestration/temporal-java/src/main/java/edu/washu/tag/temporal/activity/SplitHl7LogActivityImpl.java
```diff
@@ -85,7 +105,7 @@ public SplitAndTransformHl7LogOutput splitAndTransformHl7Log(SplitAndTransformHl
         try {
             uploadedList = uploadHl7PathList(hl7Paths, hl7ListFileUri);
         } catch (IOException e) {
-            throw ApplicationFailure.newFailureWithCause("Failed to upload log file list to {}" + hl7ListFileUri, "type", e);
+            throw ApplicationFailure.newFailureWithCause("Failed to upload log file list to " + hl7ListFileUri, "type", e);
```
kathrynalpert, line 108: Do we want to log stack trace here?

orchestration/temporal-java/src/main/java/edu/washu/tag/temporal/activity/SplitHl7LogActivityImpl.java
```diff
@@ -223,57 +241,61 @@ private List<String> processLogFile(String logFile, URI destination) throws IOEx
...
-        String sourceInfo = logFilePath + " (section " + hl7Count + ")";
+    private Hl7File transformToHl7AndUpload(String logFile, List<String> lines, URI destination, int segmentNumber, String workflowId, String activityId) {
         try {
-            String outputPath = transformAndUpload(lines, destination, sourceInfo);
-            outputPaths.add(outputPath);
-        } catch (IOException | FileFormatException e) {
-            logger.error("Unable to transform to HL7 {}", sourceInfo, e);
+            return transformAndUpload(logFile, lines, destination, segmentNumber);
+        } catch (IOException e) {
+            return Hl7File.error(logFile, segmentNumber, destination.toString(), "Unable to transform to HL7: " + e.getMessage(), workflowId, activityId);
```
kathrynalpert, line 264: Do we want to log stack trace here?

orchestration/temporal-java/src/main/java/edu/washu/tag/temporal/db/IngestDbServiceImpl.java
```diff
@@ -0,0 +1,74 @@
...
+                logger.warn("DataSource is not an instance of HikariDataSource. Connection parameters may not be accessible.");
+            }
+        } catch (Exception e) {
+            logger.error("Failed to log database connection parameters", e);
+        }
+    }
+
+    @Override
+    public void insertLogFile(LogFile logFile) {
+        logger.info("WorkflowId {} ActivityId {} - Inserting log file {} record into database", logFile.workflowId(), logFile.activityId(), logFile.filePath());
```
kathrynalpert, line 40: Should this be debug level?

orchestration/temporal-java/src/main/resources/db/migration/V2__Add_indexes.sql
```diff
@@ -0,0 +1,9 @@
+-- Indexes for log_files
```
kathrynalpert, line 1: I don't know if these are the right things to index, either. I wonder if some testing with Grafana or Superset might answer that question.

kathrynalpert (approved):

orchestration/temporal-java/src/main/resources/db/migration/V2__Add_indexes.sql
andylassiter: +1

andylassiter (approved):

.github/ci_resources/inventory.yaml
johnflavin: I wanted to put it here in anticipation of needing to install postgres independently of the orchestrator. Plus, I just don't like having to modify a playbook if you do want to change some value.

ansible/inventory.example.yaml
johnflavin: I cherry-picked everything from the `postgres` branch into this one, but I made my own tweaks after that. (Unless you have more changes on `postgres` that you haven't pushed up.)

kathrynalpert: Nope, nothing more locally.

ansible/playbooks/tasks/postgres.yaml
johnflavin: I also don't know how I feel about this. I looked at examples in `orchestration.yaml`, and we're kind of all over the place. 

- The pg version here is pinned to an exact chart release.
- For `cass-operator` we don't specify a version, meaning we're always using the `latest` chart (though later we do specify a `serverVersion` which I think pins the version of cassandra).
- For temporal we have `chart_version: ^0.56.0`. But I don't know the semantics of the `^` operator. Will this take _any_ version ≥ `0.56.0`? Or only bumps to minor and patch but not major?

Broadening to other files: with `k3s` we're using their install script that we just pulled off the web, so who knows what we're getting there.

Anyway, back to this example. I could add a `^` before the version. I don't know exactly what versions that would allow or not, but it's likely safer than some of the looser versioning practices we are using elsewhere. But maybe this is pointing to the fact that we would benefit from taking a pass through all our versions and being consistent on our practices.

kathrynalpert: See https://github.com/Masterminds/semver?tab=readme-ov-file#caret-range-comparisons-major. I was questioning whether we should use `^0.23.2` for >=0.23.2 and <0.24.0. I think it may be impossible to be fully consistent in this, since different services may have different sensitivity to version changes.

orchestration/temporal-java/src/main/resources/db/migration/V2__Add_indexes.sql
andylassiter: Ok with me, too.

ansible/playbooks/tasks/postgres.yaml
johnflavin: Updated to `^0.23.2`

PR #74, opened 2025-04-22
Add "Analytics" service

kathrynalpert:
# Add Analytics service (Superset + Trino) and swap to using Hive Metastore Service catalog with Spark

## Type of change
- [x] New feature
- [ ] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

### Product
Integrate Apache Superset into the Scout stack for no-code visualizations and SQL lab query environment. Due to technical limitations (that may soon be addressed on Superset side - PRs as recent as this week), Superset cannot run on a subpath. So, the root path will now take the user to Superset, and the "whitelabeled" logo within superset will bring the user to the "Launchpad," which was formerly at the root.

In the next PR, I would like to add a default dashboard and have the user land there vs the default Superset home page.

### Technical
To integrate Apache Superset with Delta Lake, we need to set up Trino and Hive Metastore (with Thrift RPC service) as intermediary services. The Delta Lake storage format exposes parquet files and a `_delta_log` directory, but no SQL endpoint for Superset to query. Trino, a distributed query engine, can add SQL support to Delta Lake using its Delta Lake connector, which, in turn, needs a Hive‑compatible metastore to discover schema and table locations.

>A database engine handles a lot of the custom interactions between various databases and maps them to the interface that Superset understands. If certain concepts are missing in a certain database, like time granularity or SQL syntax, the database engine for that database indicated to Superset that this is not available. As a result the option does not show in Superset, or a concise error message is reported. By default, database engines use the [base.py](https://github.com/apache/superset/blob/master/superset/db_engine_specs/base.py) methods, but each engine, like Trino, add the custom mappings with a specific engine implementation, [trino.py](https://github.com/apache/superset/blob/master/superset/db_engine_specs/trino.py).

[Source](https://trino.io/episodes/12.html)

> Trino allows you to query many different data sources with SQL statements. You enable that by creating a catalog that contains the configuration to connect to a specific data source. The data source can be a relational database, a NoSQL database, and an object storage. A connector is the translation layer that maps the concepts in the data source to the Trino concepts of schema, tables, rows, columns, data types and so on. The connector needs to know how to retrieve the data itself from data source, and also how to interact with the metadata.
> 
> Here are some examples metadata questions to answer:
> 
> What are the available tables in schema `xyz`?
> What columns does table `abc` have and what are the data types?
> What file format is used by the storage for table `efg`?
> A connector unlocks the data to all your SQL analytics powered by Trino, and the underlying data source doesn’t even have to support SQL.

[Source](https://trino.io/episodes/34.html)

**Query Flow**

* Superset sends SQL queries to Trino
* Trino uses the Delta connector to
    * Interface with Hive Metastore to get table metadata
    * Read Delta Lake transaction logs to determine the current table state
    * Access the actual Parquet files in the data lake
* Trino returns results to Superset for visualization

##### Authorization
No auth for HMS or Trino. Superset has one default admin user. Others can be added via the UI.

##### Appsec
* Superset user currently has write access to the Delta Lake to allow uploading a csv as a new table for JOINs. 
* Jupyter users can access the HMS (needed for Spark SQL)

### Performance
I ran an ingest of the `8115k_larger_report_data_sample` test data with HMS instead of the default Spark catalog. The ingest workflows completed in ~6hrs and the delta lake workflows completed an hour later, so ~7hrs total for the ingest & store. I believe this is comparable to previous results.

### Data
I haven't extensively tested with Superset, I'm sure there are some data issues. One known issue is within SQL Lab, Superset complains: `trino error: TrinoUserError(type=USER_ERROR, name=COLUMN_NOT_FOUND, message="line 5:3: Column 'partition' cannot be resolved", query_id=20250422_185955_00078_mhst8)`

### Backward compatibility
NOT BACKWARD COMPATIBLE. Prior ingests need to be rerun to populate the Hive metastore. (You can create a new metastore by running a `SELECT *` on the previous lake via Spark SQL, but we'd want to explore implications before doing so.)

## Testing
I ran an ingest of the `8115k_larger_report_data_sample` test data with HMS instead of the default Spark catalog. I've played around with Superset manually and queried the lake via Trino CLI.

## Note for reviewers
* I'm hoping to follow this PR with changes to include a default dashboard
* Right now, I'm hard-coding the table name. Should this be a variable with a default?

## Checklist


ansible/playbooks/services/hive.yaml
```diff
@@ -0,0 +1,68 @@
...
+          value: ''
+        - name: AZURE_ABFS_OAUTH_SECRET
+          value: ''
+        - name: AZURE_ABFS_OAUTH_ENDPOINT
+          value: ''
+        - name: AZURE_WASB_ACCESS_KEY
+          value: ''
+
+- name: Wait for Hive to be ready
+  command: kubectl -n hive wait --for=condition=Available --timeout=300s deployments/hive-metastore
```
johnflavin, line 66: Need to use `-n {{ hive_namespace }}`

helm/jupyter/hub/hub-values.yaml
```diff
@@ -78,3 +78,9 @@ singleuser:
                 kubernetes.io/metadata.name: minio-scout
         ports:
           - port: 9000
+      - to:
+          - namespaceSelector:
+              matchLabels:
+                kubernetes.io/metadata.name: hive
```
johnflavin, line 84: This should be `{{ hive_namespace }}`, but you'd need to have ansible read it for that template variable to be resolved.

kathrynalpert: Good thing I have a model for how to do this :)

ansible/playbooks/services/grafana.yaml
```diff
@@ -0,0 +1,104 @@
...
+        path: '{{ grafana_dir }}'
+
+- name: Add Grafana Helm repository
+  kubernetes.core.helm_repository:
+    name: grafana
+    repo_url: https://grafana.github.io/helm-charts
+
+- name: Create Loki namespace
+  kubernetes.core.k8s:
+    name: '{{ loki_namespace }}'
```
andylassiter, line 19: The loki namespace is hardcoded in `helm/grafana/values.yaml`. That's an existing issues and I don't think should be fixed in this PR. 

@johnflavin You'll be moving `helm/grafana/values.yaml` into Ansible when you add a postgres datasource to grafana, right? I think this can be easily fixed then.

andylassiter (approved):

ansible/playbooks/services/grafana.yaml
johnflavin: Yes, I will almost certainly need to pull that file into ansible and could parameterize the namespace when I do.

johnflavin (approved):

ansible/playbooks/services/grafana.yaml
kathrynalpert: I actually differentiated these in `inventory.example.yaml` in a "DO NOT CHANGE" section. Will be great to get rid of it. 

https://github.com/washu-tag/scout/blob/2e7d3e9628e7c7f28a24d5c47ba0312318f885de/ansible/inventory.example.yaml#L63

PR #76, opened 2025-04-23
Unparsable HL7 file error

johnflavin:
## Type of change
- [ ] New feature
- [X] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

### Product
Add error information for HL7 files that were written to object store but can't be ingested into delta lake

### Technical
- Detect HL7 file parse errors as a null value for message control ID `MSH-10`
    - Write error information for these rows
    - Filter these rows out before writing to delta lake
- Add `psycopg` dependency to python worker to enable connecting to postgres
- Add ansible configuration to pass postgres connection info to python worker

## Testing
- Added a file to the test data set that is a valid hl7-ish log file and contains valid headers but the resulting HL7 file will fail to parse. (Contributed by Charlie)
- Ran the ingest workflow on the test data set. Verified that an error record is written to the db for this file

## Note for reviewers
N/A

## Checklist


orchestration/temporal-python/src/temporalpy/db.py
```diff
@@ -0,0 +1,85 @@
...
+        )
+    return _connection
+
+
+def write_errors(
+    hl7_files: list[str], error_message: str, workflow_id: str, activity_id: str
+) -> None:
+    """Write an error message to the database for a list of HL7 files."""
+    activity.logger.info("Writing errors to database for %d HL7 files", len(hl7_files))
+    conn = connect_to_db()
```
kathrynalpert, line 36: Do we need to worry about closing this connection?

orchestration/temporal-python/src/temporalpy/hl7extractor/deltalake.py
```diff
@@ -84,6 +84,26 @@ def import_hl7_files_to_deltalake(
             .withColumn("source_file", F.input_file_name())
         )
 
+        # Filter out rows from empty / unparsable HL7 files
+        message_control_id = segment_field("MSH", 10)
+        error_paths_df = (
+            df.filter(message_control_id.isNull()).select("source_file").collect()
+        )
+        error_paths = [
+            row.source_file.replace("s3a://", "s3://") for row in error_paths_df
```
kathrynalpert, line 93: This is a bit tangential to this PR, but the delta lake `source_file` column uses filenames with "s3a" prefix. Seems like it'd be nice to standardize on one or the other.

kathrynalpert (approved):

orchestration/temporal-python/src/temporalpy/db.py
johnflavin: I was thinking that we don't need to worry about it. But now that you point it out, I think you’re probably right. I will change this to close it.

orchestration/temporal-python/src/temporalpy/hl7extractor/deltalake.py
johnflavin: You’re right both that it's a bit tangential, but also that we should standardize the scheme of the paths. IMO they should all be `s3` since
1. That's what a user would be most familiar with and
2. They could take a path out and pass it to various tools and it would most likely work (...except for spark/hadoop).

kathrynalpert: Would it make sense to update this code in this PR so that we're consistent. If you feel `s3` is best, can you update the other location (which writes to the lake) to also replace the `s3a`?

kathrynalpert: Doh! I see it now. Thanks!

PR #81, opened 2025-04-25
Remove problematic image tag from docs

kathrynalpert:
This works differently in GH vs RTD, and it's not important, just remove it


andylassiter (approved):
These doc previews are nice.

johnflavin (approved):

PR #85, opened 2025-04-29
First pass at postgres status tests

cmoore01:
# First pass at postgres status tests

## Type of change
- [X] New feature
- [ ] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [X] Test update

### Product
N/A

### Technical
This adds a new class (both in the Java sense and the categorical sense) that connects to the postgres instance in scout containing statuses about the import process. We check that successful imports and represented properly, along with the failure scenarios in the README of the test data. The following entities are checked:
1. `log_files` (table)
2. `recent_log_files` (view)
3. `hl7_files` (table)
4. `recent_hl7_files` (view)

##### Authorization
N/A

##### Appsec
N/A

### Performance
N/A

### Data
Failure scenarios already in the test data.

### Backward compatibility
N/A

## Testing
Successful run on big-01 and successful run on GHA.

## Note for reviewers
The tests in this PR assume the ingestion process has been run exactly once, and for only the test data. A future update will remove the script for launching the ingestion and instead launch and monitor it from the tests. I also plan on documenting the individual tests to explain them a bit more.

## Checklist


kathrynalpert (approved):

tests/src/test/java/edu/washu/tag/BaseTest.java
```diff
@@ -1,7 +1,11 @@
 package edu.washu.tag;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 public class BaseTest {
 
+    private static final Logger log = LoggerFactory.getLogger(BaseTest.class);
```
johnflavin, line 8: Nit: Is this used?

tests/src/test/java/edu/washu/tag/tests/TestStatusDatabase.java
```diff
@@ -0,0 +1,264 @@
...
+            logRowWithRetries,
+            logRowWithRetries
+        );
+
+        runLogTest(
+            SqlQuery.logViewQuery("20240102"),
+            logRowWithRetries
+        );
+
+        final Hl7FileRow repeatedFailingHl7Message = Hl7FileRow.failure(0, null, "Split content has fewer than 3 lines");
```
johnflavin, line 70: Thinking aloud:

Seeing this makes me think that maybe this kind of error shouldn't be in the HL7 table at all. Like, if the segment of the log file is not valid HL7 then does that error message belong in the table of HL7 files? On the other hand, if a log file has many segments and only one of them is invalid, it doesn't make sense to write an error to the table where rows represent the whole log file.

Do you have any thoughts on this?

tests/src/test/java/edu/washu/tag/tests/TestStatusDatabase.java
```diff
@@ -0,0 +1,264 @@
...
+        private final String date;
+        private final String errorMessage;
+
+        private LogRow(String status, String date, String errorMessage) {
+            this.status = status;
+            this.date = date;
+            this.errorMessage = errorMessage;
+        }
+
+        private static LogRow success(String date, String errorMessage) {
```
johnflavin, line 192: Why have the `errorMessage` arg? Isn't it always `null`?

johnflavin (approved):

tests/src/test/java/edu/washu/tag/BaseTest.java
cmoore01: Technically it isn't used in this PR. This was something used in the upcoming work that I missed when ripping stuff out for this smaller PR. I'll take it out.

tests/src/test/java/edu/washu/tag/tests/TestStatusDatabase.java
cmoore01: This error is for a log file that is entirely garbage, so yes, I was originally expecting the error to be reflected in `log_files` instead.

PR #90, opened 2025-05-02
Ingest job orchestrated by tests directly

cmoore01:
# Ingest job orchestrated by tests directly

## Type of change
- [ ] Work behind a feature flag
- [ ] New feature
- [X] Improvement
- [ ] Bug fix
- [ ] Refactor (code improvement with no functional changes)
- [X] Documentation update
- [X] Test update

### Product
N/A

### Technical
This removes the previous requirement of scout only having ingested data a single time by having the tests connect to temporal directly to launch and monitor the ingest process. When launching the ingest, the tests will track the workflow ID for the overall ingest job and the workflow IDs of all of the delta lake ingest child workflows. By filtering on these IDs, we can look at only the newly relevant rows in the `ingest` tables.

##### Authorization
N/A

##### Appsec
N/A

### Performance
N/A

### Data
N/A

### Backward compatibility
The structure of the JSON for configuring the tests has been updated so previous `local.json` files will likely need to be tweaked. `deltaLakeUrl` has been removed because it is redundant with `temporalConfig.ingestJobInput.deltaLakePath`

## Testing
Tests passed on dev VM and GHA.

## Note for reviewers
See backwards compatibility.

## Checklist
- [X] My code adheres to the coding and style guidelines of the project.
- [X] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have added or updated user documentation, if appropriate.
- [X] I have added or updated technical documentation, including an architectural decision record, if appropriate.
- [ ] I have added unit tests, unless this is a test code PR.
- [X] I have added end-to-end tests, unless this is a documentation-only PR.

PR #92, opened 2025-05-02
Default params

johnflavin:
## Type of change
- [X] New feature
- [ ] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

### Product
- Admin can set default workflow params in ansible inventory
- User can run workflow with no input params to use all defaults, or override specific params as required

### Technical
- Added properties to example and test `inventory.yaml` for default paths and table names
- Added infrastructure to `hl7log-extractor` to read default param properties and provide them to the workflow

### Security 
N/A

### Performance
N/A

### Data
N/A

### Backward compatibility
This should be backwards compatible with previous ways of launching the workflow that included all the params in the launch request.

## Testing
Deployed to big-03 and ran workflows with some subset of the args (including no args) on the test data.

## Note for reviewers
1. ~~I haven't updated the test code that launches workflows because I know @cmoore01 is working on it. Once #90 is merged I can rebase on it and change the test launching code.~~ Done
2. ~~I know there was a discussion on slack about removing the explicit delta lake path and letting hive control that. I have not implemented this yet; I am not sure how much time it would take to implement and test so I thought it prudent to put up the PR with the simpler change first so it could be reviewed. I would like to get it into this PR if there is adequate time to test it.~~ Done
3. I was just going through the checklist below and realized that if there is any user documentation about how to launch the workflow I should update that.

## Checklist


.github/ci_resources/inventory.yaml
```diff
@@ -76,7 +90,7 @@ k3s_cluster:
 
     # computed values, used across playbooks
     s3_endpoint: 'http://minio.{{ minio_tenant_namespace }}'
-    hive_warehouse: s3a://lake/delta # Hadoop requires the s3a prefix, it is used just like s3://
+    hive_warehouse: 's3a://{{ lake_bucket }}/delta' # Hadoop requires the s3a prefix, it is used just like s3://
```
kathrynalpert, line 93: [This also applies to `inventory.example.yaml`, but you didn't change the line there]

I think we should either have just this variable or just `delta_lake_path`. If you remove the `location` specification from the Extractor, then nothing needs `delta_lake_path`, and Hive's "s3a" path is all that's used. However, I like calling it `delta_lake_path` vs `hive_warehouse` as that's more clear about what it is.

extractor/hl7-transformer/src/hl7scout/hl7extractor/deltalake.py
```diff
@@ -256,23 +257,23 @@ def import_hl7_files_to_deltalake(
...
         activity.logger.info(
-            "Creating Delta Lake table %s if it does not exist", delta_table
+            "Creating Delta Lake table %s if it does not exist", delta_lake_path
         )
         dt = (
             DeltaTable.createIfNotExists(spark)
-            .tableName("default.reports")
-            .location(delta_table)
+            .tableName(f"default.{report_table_name}")
+            .location(delta_lake_path)
```
kathrynalpert, line 269: I'm suggesting we remove this line completely, and thus the need for `delta_lake_path` at all in the extractor. Hive manages this for us.

kathrynalpert (approved):

PR #104, opened 2025-05-07
Postgres ingest db schema changes

johnflavin:
This is a bit of an experimental PR. I'm proposing a big change by writing the documentation for it before actually implementing the change. The idea is for us to review the proposal and come to an agreement/understanding of what the change will be before I spend a bunch of time doing it and locking in changes that the team has never seen.

Once we review this proposal... I guess I could merge the doc and then do the work on a different branch / different PR? Or maybe I don't merge this yet, push the implementation changes up to this branch, then re-review this PR? I'm not sure, I haven't thought that part through.

The particular change I'm proposing is a redesigned schema for the postgres ingest db. See the doc for details.


docs/source/ingest_status_db.md
```diff
@@ -0,0 +1,42 @@
...
+| `file_statuses` | Contains records of the status of each file as it is processed. This table is append-only, meaning we will write multiple rows for a single file as its status is updated. File records in this table can be for HL7 files as well as the "HL7-ish" log files from which the HL7 files are extracted. |
+| `hl7_files` | Contains a correspondence between an HL7 file and the "HL7-ish" log file that it was extracted from. |
+
+## file_statuses Table
+The `file_statuses` table contains the following columns:
+| Column Name | Data Type | Description |
+|-------------|-----------|-------------|
+| `id` | integer | The unique identifier for the file status record. |
+| `file_path` | string | The path to the file. HL7 file paths will be S3 URIs, HL7-ish log file paths will be local paths as seen from within the container. |
+| `type` | string | The type of file. Possible values are: `HL7` and `Log`. |
```
kathrynalpert, line 21: I don't necessarily object to combining logs and HL7s into one audit table, but I'm not sure what we gain from doing so either. I guess just one schema to maintain vs two nearly identical ones? At the cost of requiring our views read from a really large table?

kathrynalpert (approved):

kathrynalpert (commented):
I guess I should not have approved this, as you might add code to it. I approve the schema changes :)

kathrynalpert (changes requested):
Requesting changes to remove approval

docs/source/ingest_status_db.md
```diff
@@ -0,0 +1,42 @@
...
+| `activity_id` | string | The ID of the activity that processed the file. |
+| `processed_at` | timestamp | The timestamp when the file status record was created. |
+
+## hl7_files Table
+The `hl7_files` table contains the following columns:
+| Column Name | Data Type | Description |
+|-------------|-----------|-------------|
+| `hl7_file_path` | string | The path to the HL7 file. |
+| `log_file_path` | string | The path to the "HL7-ish" log file. |
+| `segment_number` | integer | The segment of the HL7-ish log file from which the HL7 file was extracted. |
```
cmoore01, line 34: If we're revisiting the schema, what do we think about picking a different term than `segment` here? The part I find potentially confusing is that "segment" is the name for sections within HL7, so it feels odd to duplicate it. It makes me think more of where a parsing issue happened within an HL7 message.

docs/source/ingest_status_db.md
```diff
@@ -0,0 +1,42 @@
...
+|------------|-------------|
+| `file_statuses` | Contains records of the status of each file as it is processed. This table is append-only, meaning we will write multiple rows for a single file as its status is updated. File records in this table can be for HL7 files as well as the "HL7-ish" log files from which the HL7 files are extracted. |
+| `hl7_files` | Contains a correspondence between an HL7 file and the "HL7-ish" log file that it was extracted from. |
+
+## file_statuses Table
+The `file_statuses` table contains the following columns:
+| Column Name | Data Type | Description |
+|-------------|-----------|-------------|
+| `id` | integer | The unique identifier for the file status record. |
+| `file_path` | string | The path to the file. HL7 file paths will be S3 URIs, HL7-ish log file paths will be local paths as seen from within the container. |
```
kathrynalpert, line 20: Instead of "HL7-ish logs" you could say "HL7 listener/subscriber logs", as that's what they really are - logs from the HL7 listener application.

docs/source/ingest_status_db.md
johnflavin: > I guess just one schema to maintain vs two nearly identical ones?

That was my idea, yes. The schemas were already almost identical before, and once I removed the `log_file_path` and `segment_number` columns from the HL7 file table they were identical.

> At the cost of requiring our views read from a really large table?

The HL7 file table was already going to be huge (every day we'll write a few thousand files * 3–5 status updates per file). Adding a few statuses for one log file per day is nothing compared to all those HL7 files.

This is making me think that maybe we could periodically consolidate it to remove all but the most recent status for each file? Or maybe I should be doing updates instead of inserts only?

docs/source/ingest_status_db.md
```diff
@@ -0,0 +1,42 @@
...
+| `log_file_path` | string | The path to the "HL7-ish" log file. |
+| `segment_number` | integer | The segment of the HL7-ish log file from which the HL7 file was extracted. |
+| `date` | date | The date of the HL7-ish log file. |
+
+## Database Views
+The Ingest Status DB also contains the following views:
+| View Name | Description |
+|-----------|-------------|
+| `recent_log_file_statuses` | A view that shows the most recent status of each log file. This view is useful for monitoring the progress of the ingestion process. |
+| `recent_hl7_file_statuses` | A view that shows the most recent status of each HL7 file. This view is useful for monitoring the progress of the ingestion process. |
```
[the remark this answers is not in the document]
johnflavin: I'm not sure if that join should be part of the view itself or something that happens only in queries that need it. Maybe we leave this one as-is but add another view that joins in `date` and the other cols from the files table.

docs/source/ingest_status_db.md
johnflavin: I'm open to it, but I don't know what the alternate term would be. (Consulting a thesaurus...) Section? Division? Entry? Partition?

docs/source/ingest_status_db.md
kathrynalpert: I think we should figure out how to manage a huge table. I vote leave it as you have it.

docs/source/ingest_status_db.md
kathrynalpert: What about calling it an HL7 message?

docs/source/ingest_status_db.md
```diff
@@ -0,0 +1,44 @@
+# Ingest Status DB
+The Ingest Status DB is a database that stores the status of the HL7 report ingestion process. It contains information about
+the status of each report, including whether it has been successfully ingested, failed, or is in progress. This information
+is used to monitor the progress of the ingestion process and to troubleshoot any issues that may arise.
+ 
+The Ingest Status DB is populated by the {ref}`Extractor service <extractor_ref>` as it processes HL7 files. The process
```
cmoore01, line 6: If this document gets moved from readthedocs to internal, this reference will need to be removed or modified.

docs/source/ingest_status_db.md
johnflavin: How about
```md
| `message_id` | integer | ID of the segment of the log file from which the HL7 file was extracted. |
```

kathrynalpert: I think `message_count` or `message_number` would work

kathrynalpert:
I think we could live without date for `log_file` errors, since these are usually like file not found or not readable things, and we can discern date from the file path as well as the java code can.

I would argue if we go this route, we should make a separate table for log file error messages though.

tests/src/test/java/edu/washu/tag/tests/TestStatusDatabase.java
johnflavin, line 1: I'm still in progress converting these tests for the new schema. First two are done, the rest are commented out so the tests can compile and run while I continue working on the rest.

extractor/hl7log-extractor/src/main/java/edu/washu/tag/extractor/hl7log/activity/SplitHl7LogActivityImpl.java
```diff
@@ -220,10 +243,11 @@ private String parseAndValidateTimestamp(String headerStr) throws FileFormatExce
      * @return List of Hl7File instances containing generated output file paths or error messages
      * @throws IOException If an I/O error occurs
      */
-    private List<Hl7File> processLogFile(String logFile, URI destination, String workflowId, String activityId) throws IOException {
+    private Pair<List<FileStatus>, LocalDate> processLogFile(String logFile, URI destination, String workflowId, String activityId) throws IOException {
         Path logFilePath = Paths.get(logFile);
 
-        List<Hl7File> results = new ArrayList<>();
+        LocalDate date = dateFromLogFilePath(logFile);
```
kathrynalpert, line 249: Why not just compute this in the calling method (then we wouldn't have to return a Pair)? Or... since we have it here, why not add it to the FileStatus?

extractor/hl7log-extractor/src/main/java/edu/washu/tag/extractor/hl7log/activity/SplitHl7LogActivityImpl.java
```diff
@@ -291,38 +315,49 @@ private Hl7File validateWriteAndUploadHl7(String logFile, List<String> lines, St
         try {
             timestamp = extractTimestamp(headerLine);
         } catch (FileFormatException e) {
-            logger.warn("WorkflowId {} ActivityId {} - Unable to extract timestamp for message {}: {}", workflowId, activityId, segmentNumber, e.getMessage());
-            return Hl7File.error(logFile, segmentNumber, "Unable to extract timestamp for message: " + e.getMessage(), workflowId, activityId);
+            logger.warn("WorkflowId {} ActivityId {} - Unable to extract timestamp for segment {}: {}", workflowId, activityId, messageNumber, e.getMessage());
```
kathrynalpert, line 318: segment -> message

extractor/hl7log-extractor/src/main/java/edu/washu/tag/extractor/hl7log/activity/SplitHl7LogActivityImpl.java
```diff
@@ -291,38 +315,49 @@ private Hl7File validateWriteAndUploadHl7(String logFile, List<String> lines, St
...
+            return FileStatus.failed(
+                createPlaceholderHl7FilePath(logFile, messageNumber),
+                FileStatusType.HL7,
+                "Unable to extract timestamp for segment: " + e.getMessage(),
```
kathrynalpert, line 322: segment -> message

extractor/hl7log-extractor/src/main/java/edu/washu/tag/extractor/hl7log/activity/SplitHl7LogActivityImpl.java
```diff
@@ -291,38 +315,49 @@ private Hl7File validateWriteAndUploadHl7(String logFile, List<String> lines, St
...
         } catch (IOException e) {
-            logger.error("WorkflowId {} ActivityId {} - Could not write message {} to HL7 file", workflowId, activityId, segmentNumber, e);
-            return Hl7File.error(logFile, segmentNumber, "Could not write message to HL7 file: " + e.getMessage(), workflowId, activityId);
+            logger.error("WorkflowId {} ActivityId {} - Could not write message {} to HL7 file", workflowId, activityId, messageNumber, e);
+            return FileStatus.failed(createPlaceholderHl7FilePath(logFile, messageNumber), FileStatusType.HL7, "Could not write message to HL7 file: " + e.getMessage(), workflowId, activityId);
         }
     }
 
+    private String createPlaceholderHl7FilePath(String logFile, int segmentNumber) {
+        return logFile + "_" + segmentNumber;
```
kathrynalpert, line 337: Should we name all the HL7 files this way? Saves us some parsing, should be unique & constant... Feels scary (like messageNumber could change), but I don't _think_ it could.

Also, swap from segment -> message.

extractor/hl7log-extractor/src/main/java/edu/washu/tag/extractor/hl7log/activity/SplitHl7LogActivityImpl.java
```diff
@@ -332,17 +367,17 @@ private Hl7File writeAndUpload(String logFile, List<String> lines, String timest
...
                 String outputPath = fileHandler.putWithRetry(outputStream.toByteArray(), relativePath, destination);
-                return Hl7File.success(logFile, segmentNumber, outputPath, workflowId, activityId);
+                return FileStatus.staged(outputPath, FileStatusType.HL7, workflowId, activityId);
             } catch (IOException e) {
-                logger.error("WorkflowId {} ActivityId {} - Failed to upload segment {} HL7 file {}/{}", workflowId, activityId, segmentNumber, destination,
-                    relativePath, e);
-                return Hl7File.error(logFile, segmentNumber, "Failed to upload HL7 file", workflowId, activityId);
+                logger.error("WorkflowId {} ActivityId {} - Failed to upload segment {} HL7 file {}/{}",
+                    workflowId, activityId, messageNumber, destination, relativePath, e);
+                return FileStatus.failed(createPlaceholderHl7FilePath(logFile, messageNumber), FileStatusType.HL7, "Failed to upload HL7 file", workflowId, activityId);
```
kathrynalpert, line 380: We have the actual HL7 file name, but not it's actual path on S3 (since it didn't get uploaded to S3). I sort of see what we're doing - local path when we have an error vs S3 path for success - but the local path doesn't actually correspond to a file, so I think it's perhaps a bit confusing.

extractor/hl7log-extractor/src/main/java/edu/washu/tag/extractor/hl7log/activity/SplitHl7LogActivityImpl.java
```diff
@@ -365,4 +400,14 @@ private byte[] convertListToByteArray(List<String> hl7FilePaths) {
...
         return content.getBytes(StandardCharsets.UTF_8);
     }
+
+    private static LocalDate dateFromLogFilePath(String filePath) {
+        String fileName = Paths.get(filePath).getFileName().toString();
+        Matcher m = DATE_PATTERN.matcher(fileName);
+        if (m.find()) {
+            return LocalDate.parse(m.group(), DATE_FORMATTER);
+        }
+        return null;
```
kathrynalpert, line 410: Are uses of this null safe? Should we use `@Nullable`?

extractor/hl7log-extractor/src/main/java/edu/washu/tag/extractor/hl7log/activity/SplitHl7LogActivityImpl.java
```diff
@@ -68,25 +79,37 @@ public SplitAndTransformHl7LogOutput splitAndTransformHl7Log(SplitAndTransformHl
...
         URI destination = URI.create(input.hl7OutputPath());
-        List<Hl7File> segmentResults;
+        List<FileStatus> segmentResults;
+        LocalDate date;
         try {
-            segmentResults = processLogFile(input.logPath(), destination, workflowId, activityId);
-
-            ingestDbService.insertLogFile(LogFile.success(input.logPath(), workflowId, activityId));
+            Pair<List<FileStatus>, LocalDate> processedResults = processLogFile(input.logPath(), destination, workflowId, activityId);
+            segmentResults = processedResults.getLeft();
```
kathrynalpert, line 86: Maybe swap from `segmentResults` to `messageUploadStatuses` ? I don't know :)

extractor/hl7log-extractor/src/main/java/edu/washu/tag/extractor/hl7log/db/DbUtils.java
```diff
@@ -51,7 +91,17 @@ private static <T> String generateInsertSql(Class<T> recordClass) {
             .limit(columnNames.size())
             .collect(Collectors.joining(", "));
 
-        return String.format("INSERT INTO %s (%s) VALUES (%s)", tableName, columns, placeholders);
+        String insertSql = String.format("INSERT INTO %s (%s) VALUES (%s)", tableName, columns, placeholders);
```
kathrynalpert, line 94: Any length-limit concerns here (size of placeholders)

johnflavin: No, the number of placeholders here is the number of components in the record class (= number of cols in the db table). None of our tables have so many columns that it would cause an error.

extractor/hl7log-extractor/src/main/java/edu/washu/tag/extractor/hl7log/activity/SplitHl7LogActivityImpl.java
johnflavin: This `date` isn't used for anything in the code, it is just written directly into the `hl7_files` table as-is. It is fine for that to be `null` at least in the sense that it won't cause any errors in the code.

extractor/hl7log-extractor/src/main/java/edu/washu/tag/extractor/hl7log/activity/SplitHl7LogActivityImpl.java
johnflavin: Yeah, I went back and forth on whether I should write this placeholder file path or just leave it `null`. I'm still not sure I made the correct choice, but I can tell you my reasoning. If you leave a `null` for the file path, then when you see a `"failed"` status row you aren't able to identify which log file or message within that file caused the error. Having the fake path here lets you join back to the `hl7_files` table and identify the log file and message number. (Or you could just extract those values from the `file_path` value itself if you know its structure.)

extractor/hl7log-extractor/src/main/java/edu/washu/tag/extractor/hl7log/activity/SplitHl7LogActivityImpl.java
johnflavin: > Why not just compute this in the calling method (then we wouldn't have to return a Pair)?
That makes sense, I can do that.

extractor/hl7log-extractor/src/main/java/edu/washu/tag/extractor/hl7log/activity/SplitHl7LogActivityImpl.java
johnflavin: > Should we name all the HL7 files this way? Saves us some parsing, should be unique & constant... 

IMO we want to put the HL7 files into subdirectories down to the hour to avoid having thousands of files in a single date directory, so we'd still need to parse out the higher-precision timestamp from the header. And once we have that timestamp we may as well use it, we're guaranteed to never have collisions so long as it takes more than 100µs for the upstream server to write its messages into the log.

> Feels scary (like messageNumber could change), but I don't think it could.

Now that you mention it... the message number might change. Or at least the message number that we have here in this line (which is derived from keeping a count of message chunks in the log file) isn't necessarily the same as the message number that we write to the `hl7_files` table (which is based on the index of the `FileStatus` message in the list that gets passed to `batchInsertNewHl7FileStatuses`). I'm confident that right now those two are the same because every time we increment the counter we make a `FileStatus` object. But if we ever make some change to this code that drops some message chunk without creating a `FileStatus` object, then those two message counts could in principle get out of sync.

extractor/hl7log-extractor/src/main/java/edu/washu/tag/extractor/hl7log/activity/SplitHl7LogActivityImpl.java
kathrynalpert: Yeah, I realized that after I posted. I wonder if we should modify the path such that it's obvious it's a `DUMMY`. Probably ok as is for now.

extractor/hl7log-extractor/src/main/java/edu/washu/tag/extractor/hl7log/activity/SplitHl7LogActivityImpl.java
kathrynalpert: Got it. To be more clear, I wasn't meaning to suggest we change the path structure, just the file name itself (bc I liked tying the file name to the message number). But yeah, seems too risky somehow.

kathrynalpert (approved):
Approved pending small changes discussed above

PR #111, opened 2025-05-12
Add heartbeats to split and upload activity for fine-grained timeout and proper cancellation

kathrynalpert:
# Add heartbeats to split and upload activity for fine-grained timeout and proper cancellation

## Type of change
- [ ] New feature
- [x] Improvement
- [x] Bug fix
- [ ] Documentation update
- [ ] Test update

### Product
When a temporal user cancels the `IngestHl7LogWorkflow`, the running split & upload activities now get cancelled, and this is reflected in the UI (first cancelation is requested, once the activities stop, it's marked as cancelled).

This also adds a heartbeat timeout, which I've defaulted to 30m but made adjustable via Ansible and as part of the input. The idea here is to have a more fine-grained read on whether an activity is still working (versus having to wait the whole `startToClose` timeout). As such, I've bumped up the default `startToClose` timeout for the split & upload activity to 8hrs, though this can still be adjusted via Ansible and as an input parameter.

### Technical
Added heartbeat calls. Cancellation causes the heartbeat method to throw an exception, which culminates in the activities all cancelling.

##### Authorization
N/A

##### Appsec
N/A

### Performance
This may have some impact on temporal performance, as now we're hitting the heartbeat API with a lot of information. I could heartbeat less often (right now, once per parsed message, plus around the aggregate activities like writing the manifest file), but then cancelation takes longer to "take effect". I will do some experimenting to see if I can find a good balance. Temporal does throttle heartbeat traffic to ameliorate this issue.

### Data
N/A

### Backward compatibility
No change

## Testing
Running CI, running 2023 from 8k test data on `big-04` and cancelling while tailing the logs. Then restarting 2023 and looking at Temporal metrics to see if we're seeing increased latencies due to the additional network traffic.

## Note for reviewers
N/A

## Checklist
- [x] I have added unit tests, unless this is a test code PR.

### DIFF
--- a/docs/source/ingest.md
+++ b/docs/source/ingest.md
@@ -18,6 +18,9 @@ corresponding Ansible variables.
 - `hl7OutputPath`: path to write HL7 files. Note that this is _not_ the path to the resulting delta lake. Ansible equivalent: `hl7_path`.
 - `splitAndUploadTimeout`: timeout in minutes for the activity that splits the HL7 listener log files and uploads the component HL7 messages to MinIO. 
 Ansible equivalent: `hl7log_extractor_timeout`
+- `splitAndUploadHeartbeatTimeout`: timeout in minutes for **heartbeats** during the activity that splits the HL7 listener log files and uploads 
+the component HL7 messages to MinIO. This is roughly the time to parse and upload one HL7 message, but it also includes steps like writing the manifest 
+file and updating the database. Ansible equivalent: `hl7log_extractor_heartbeat_timeout`
 - `splitAndUploadConcurrency`: number of HL7 listener log files to process concurrently. Ansible equivalent: `hl7log_extractor_concurrency`
 - `modalityMapPath`: path to read modality map file, which is the source of the `modality` column in the Delta Lake table.
    Ansible equivalent: `modality_map_path`.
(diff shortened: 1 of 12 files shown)


docs/source/ingest.md
```diff
@@ -18,6 +18,9 @@ corresponding Ansible variables.
 - `hl7OutputPath`: path to write HL7 files. Note that this is _not_ the path to the resulting delta lake. Ansible equivalent: `hl7_path`.
 - `splitAndUploadTimeout`: timeout in minutes for the activity that splits the HL7 listener log files and uploads the component HL7 messages to MinIO. 
 Ansible equivalent: `hl7log_extractor_timeout`
+- `splitAndUploadHeartbeatTimeout`: timeout in minutes for **heartbeats** during the activity that splits the HL7 listener log files and uploads 
+the component HL7 messages to MinIO. This is roughly the time to parse and upload one HL7 message, but it also includes steps like writing the manifest
```
johnflavin, line 22: Kind of a nit: 

> This is roughly the time to parse and upload one HL7 message

This is more like the minimum time. Our worker will call the temporal client's `heartbeat` method this frequently, sure. But the temporal client only sends the heartbeats to the server periodically; in between sends it queues up heartbeats internally and could coalesce or drop heartbeats if they are too frequent.

johnflavin (approved):

PR #123, opened 2025-05-23
Refactor spark for performance

johnflavin:
## Type of change
- [ ] New feature
- [X] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

### Product
Speed up workflow that ingests HL7 to delta lake by optimizing spark operations

### Technical
- Inline some python functions and constants to attempt to avoid spark serializing python module. (Though this may not have actually been happening, the refactors are still probably improvements.)
- Pull HL7 segments and fields out of smolder objects earlier so the objects can be GCed ASAP
- Avoid python lambdas in favor of using pyspark SQL string expressions which contain lambdas (ensuring no python variables get captured in the lambda context)
- Add some spark config settings intended to speed up computation
    - I'm still running a test to see if these are actually necessary

##### Authorization
N/A
##### Appsec
N/A
### Performance
Improves active execution time (start-to-close) of ingest HL7 to delta lake workflow from ~1 hour 20 minutes to ~20 minutes

### Data
N/A

### Backward compatibility
No changes

## Testing
Ran several tests on big-04 and big-03 using 15m test data set. Batches of 150 log files expanded to ~700k HL7 files which were ingested via spark workflow in ~20 minutes.

## Note for reviewers
None

## Checklist


extractor/hl7-transformer/src/hl7scout/hl7extractor/deltalake.py
```diff
@@ -92,6 +58,12 @@ def import_hl7_files_to_deltalake(
         activity.logger.info("Creating Spark session")
         spark = (
             SparkSession.builder.appName("IngestHL7ToDeltaLake")
+            # Recommended setting in
+            # https://spark.apache.org/docs/latest/sql-performance-tuning.html#coalescing-post-shuffle-partitions
+            .config("spark.sql.adaptive.coalescePartitions.parallelismFirst", "false")
+            # Serialization settings for large tasks. Recommended in
+            # https://spark.apache.org/docs/latest/tuning.html#data-serialization
+            .config("spark.serializer", "org.apache.spark.serializer.KryoSerializer")
```
johnflavin, line 66: These config changes may not be required. I'm currently running a test without them. If the execution time is comparable and logs look similar, I'll remove these.

kathrynalpert (approved):
I do like the idea of moving the postgres writes outside the Spark block (also so we don't restart the pod for a postgres insert failure).

johnflavin:
> I do like the idea of moving the postgres writes outside the Spark block (also so we don't restart the pod for a postgres insert failure).

I forgot to mention that in the description, but I do still intend to test that out. Wanted to get the rest up for review before ending the day today, though.

PR #129, opened 2025-05-30
Test additions & CI tweak

cmoore01:
# Test additions & CI tweak

## Type of change
- [ ] New feature
- [X] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [X] Test update

### Product
N/A

### Technical
1. Modifies the ci configuration so that branches starting with `ci-` will have CI launched on them automatically on pushes.
2. Adds Flavin's test for no null message dt elements in a simpler form.
3. Adds a new test to check many more of the columns in the delta lake.
4. Adds a new test to repeat a full ingest request to make sure the delta lake does not get duplicate rows.

##### Authorization
N/A

##### Appsec
N/A

### Performance
N/A

### Data
N/A

### Backward compatibility
N/A

## Testing
Tests fully passing on CI.

## Checklist
- [X] I have added end-to-end tests, unless this is a documentation-only PR.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -170,9 +170,9 @@ jobs:
           cp .github/ci_resources/test_config_template.json tests/src/test/resources/config/local.json
           sed -i "s:WORK_DIR:$(pwd):" .github/ci_resources/tests-job.yaml
           sudo -E kubectl apply -f .github/ci_resources/tests-job.yaml
-          sudo -E kubectl wait --for=condition=complete --timeout=300s job/ci-tests -n extractor &
+          sudo -E kubectl wait --for=condition=complete --timeout=600s job/ci-tests -n extractor &
           completion_pid=$!
-          sudo -E kubectl wait --for=condition=failed --timeout=300s job/ci-tests -n extractor && exit 1 &
+          sudo -E kubectl wait --for=condition=failed --timeout=600s job/ci-tests -n extractor && exit 1 &
           failure_pid=$!
           wait -n $completion_pid $failure_pid
       - name: Test Logs
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,7 +4,9 @@ on:
   pull_request:
     types: [opened, reopened, edited, synchronize]
   push:
-    branches: [main]
+    branches:
+      - main
+      - 'ci/**'
 
 env:
   JAVA_DIST: 'zulu'
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,7 +6,7 @@ on:
   push:
     branches:
       - main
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yaml
```diff
@@ -168,9 +170,9 @@ jobs:
           cp .github/ci_resources/test_config_template.json tests/src/test/resources/config/local.json
           sed -i "s:WORK_DIR:$(pwd):" .github/ci_resources/tests-job.yaml
           sudo -E kubectl apply -f .github/ci_resources/tests-job.yaml
-          sudo -E kubectl wait --for=condition=complete --timeout=300s job/ci-tests -n extractor &
+          sudo -E kubectl wait --for=condition=complete --timeout=600s job/ci-tests -n extractor &
```
kathrynalpert, line 173: Do the tests actually take this long now? Ok if so, but if this was added bc of the python worker SDK issue, let's pull it back out to fail fast(er).

tests/src/test/java/edu/washu/tag/tests/TestScoutQueries.java
```diff
@@ -41,15 +40,17 @@ public Object[][] knownQueries() {
...
-            .stream()
-            .filter(testQuery -> testQuery.getId().equals(queryId))
-            .findFirst()
-            .orElseThrow(RuntimeException::new);
+        runTest(queryId);
+    }
 
-        logger.info("Performing query with spark: {}", query.getSql());
-        query.getExpectedQueryResult().validateResult(spark.sql(query.getSql()));
+    @Test(dependsOnGroups = BaseTest.RELAUNCH_PRECURSOR, alwaysRun = true)
```
kathrynalpert, line 46: Does this mean this test doesn't run until all tests within the `BaseTest.RELAUNCH_PRECURSOR` group have completed?

.github/workflows/ci.yaml
cmoore01: They don't take quite this long, but 5 minutes is definitely not safe as-is. One of the runs on GHA from this branch took 6 minutes and 9 seconds for the test pod to be done. The primary reason the time jumped up is the test to make sure we don't get duplicated rows. That test has to launch and wait for another full run of the whole ingest job. We could speed that test up a bit by launching on a smaller subset of data, but that would be at the expense of the quality of the checks (if some data caused duplicate rows but not others).

cmoore01: Oh, and actually the timeout here was only the relevant one just _barely_ with the python worker SDK issue. There's a wait here for 5 minutes, but also within the tests there's a wait of 5 minutes for the launched ingest job to complete. This one gets triggered first because it takes a few seconds before the one within the tests starts polling. So, even if we had a similar issue, we wouldn't go much above 5 minutes here anyway.

PR #139, opened 2025-06-25
Fixes for multinode cluster: taint control plane and make directories on all nodes

johnflavin:
## Type of change
- [ ] New feature
- [X] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

### Product
No user-facing changes

### Technical
Changed ansible playbooks so scout can be deployed on a multinode cluster.
- Added a `PreferNoSchedule` taint to the control plane (aka server) node.
  - Because this is `PreferNoSchedule` and not `NoSchedule` the deploy still works on a single node cluster. If there is only one node available, it'll be used.
  - Also added a toleration for this taint to the minio tenant so it can be scheduled on the control plane. We don't have minio replicas yet, but this is one step towards that.
- Fixed a problem with the directory creation for volumes.
  - Before this change directories would be created and given 777 permissions on the control plane node before mounting, but not the worker node(s). Then when the pods started on the worker nodes and mounted the volumes the directories would get created on those nodes but owned by root with 755 permissions, meaning the process running in the pod didn't have permission to write there.
  - After this change the directories will be created and given 777 permissions on all nodes, whether or not any pod will actually mount them

##### Authorization
N/A

##### Appsec
N/A

### Performance
This is the first step to enabling performance improvements by splitting the pieces of the application across multiple nodes. As we get more specialized node types (i.e. with GPUs) we will need to tweak parts of this to ensure pods are running where they are supposed to.

But even with only this change we should expect some stability improvements on a multinode cluster by protecting resources for kubernetes itself.

### Data
N/A

### Backward compatibility
This should be backwards compatible. If a scout was already deployed on a single-node cluster you can use this to update it.

## Testing
- Dev + deploy on a two-node cluster (big-03 + big-04)
- CI testing is run on a single-node cluster

## Note for reviewers
I'll add a couple of line comments with notes

## Checklist

### DIFF
--- a/.github/ci_resources/inventory.yaml
+++ b/.github/ci_resources/inventory.yaml
@@ -18,6 +18,7 @@ k3s_cluster:
     kubeconfig_group: root
 
     # Minio config, ensure passwords meet complexity requirements (must be >=8 characters)
+    minio_storage_size: 5Gi
     s3_username: minio
     s3_password: minio123
     s3_region: us-east-1
@@ -28,9 +29,12 @@ k3s_cluster:
     s3_loki_writer: loki-writer
     s3_loki_writer_secret: lokiwriter123
 
+    # Monitor config
+    prometheus_storage_size: 1Gi
+
     # Jupyter config
-    jupyter_hub_storage_size: 10Gi
-    jupyter_singleuser_storage_size: 10Gi
+    jupyter_hub_storage_size: 1Gi
+    jupyter_singleuser_storage_size: 1Gi
 
     # Postgres config
     postgres_user: scout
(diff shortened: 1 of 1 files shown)


ansible/playbooks/vars/jupyter_storage.yaml
```diff
@@ -0,0 +1,7 @@
+storage_definitions:
+  - name: jupyter-hub
+    size: '15Gi'
```
johnflavin, line 3: As with minio, I don't like having all these volume size settings buried so deep. These are particular to your deploy and might need to be changed based on the infrastructure or requirements of a site. I think it would be better to move them into the inventory.

andylassiter (approved):

PR #143, opened 2025-07-16
Extract postgres init sql for services into separate files

johnflavin:
## Type of change
- [ ] New feature
- [ ] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

## Description

I took out the hard-coded SQL init script from the postgres playbook and turned it into separate per-service SQL init files which get loaded dynamically into the postgres playbook. The reason I wanted to do this is because I have some additional users and tables that I wanted to add, and it seemed weird to keep adding more and more into the postgres playbook. This way if we need another service to set up some db stuff we simply add a new file.

There are some requirements/assumptions about the names of files and variables which I added to a README in the `postgres_init` directory containing the files.

### Backward compatibility
No incompatibility that I can tell.

## Testing
I tried deploying scout via the ansible playbook with superset disabled, then deploying again with superset enabled (simulating adding a new service). It seemed to work without errors.

## Note for reviewers
I'm a little nervous about doing database schema stuff with an init script like this. It seems like it would be safer to use a schema migration tool (like I have with flyway in the java hl7 extractor). But I don't know how to set that up cluster-wide.

In any case, the way I've done it in this change is no worse than what we currently have, and allows easy expansion.

## Checklist

### DIFF
--- a/ansible/playbooks/templates/postgres_init/hive_init_sql.yaml
+++ b/ansible/playbooks/templates/postgres_init/hive_init_sql.yaml
@@ -1,4 +1,4 @@
 ---
-base_init_sql:
+hive_init_sql:
   - "CREATE USER {{ hive_postgres_user }} WITH PASSWORD '{{ hive_postgres_password }}';"
   - 'CREATE DATABASE hive OWNER {{ hive_postgres_user }};'


ansible/playbooks/postgres.yaml
```diff
@@ -16,6 +16,60 @@
...
+          vars:
+            module_name: "{{ item.path | basename | regex_replace('_init_sql\\.yaml$', '') }}"
+            module_enabled: "{{ module_name + '_enabled' }}"
+            module_init_sql: "{{ module_name + '_init_sql' }}"
+          loop: '{{ sql_init_files.files }}'
+          loop_control:
+            label: '{{ item.path | basename }}'
+          no_log: true
+
+    - name: Remove temporary directory
```
kathrynalpert, line 68: Nit: Any concern about this not running if there's a failure on a prior step?

kathrynalpert (approved):
I take your point re flyway, but I think we can have a policy that schema is not managed this way: we only use the init SQL for creating tables & users.

ansible/playbooks/postgres.yaml
johnflavin: I don't think it is super concerning, but it also doesn't seem hard to make it always run.

PR #165, opened 2025-09-08
Adds documentation on new extraction columns

cmoore01:
# Adds documentation on new extraction columns

## Type of change
- [ ] New feature
- [ ] Improvement
- [ ] Bug fix
- [X] Documentation update
- [ ] Test update

### Product
N/A

### Technical
N/A

##### Authorization
N/A

##### Appsec
N/A

### Performance
N/A

### Data
N/A

### Backward compatibility
N/A

## Testing
Not tested yet, documentation only

## Checklist

### DIFF
--- a/docs/source/dataschema.md
+++ b/docs/source/dataschema.md
@@ -5,42 +5,43 @@ and manage HL7 radiology reports and other data and metadata. There is one main 
 Below is a description of the report table schema and the mapping of HL7 fields to the report table columns. Note that 
 some fields are derived from others, and some fields may not be directly mapped to HL7 fields.
 
-| Column Name                        | HL7 Report Field | Data Type      | Nullable | Description/Notes                                                                                                                    |
-|------------------------------------|------------------|----------------|----------|--------------------------------------------------------------------------------------------------------------------------------------|
-| `source_file`                      |                  | string         | No       | The location of the report file.                                                                                                     |
-| `updated`                          |                  | timestamp      | No       | Timestamp of the last update to the report in the Delta Lake.                                                                        |
-| `message_control_id`               | MSH-10           | string         | Yes      | Unique identifier for the HL7 message.                                                                                               |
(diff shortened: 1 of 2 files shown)


docs/source/dataschema.md
```diff
@@ -63,4 +64,27 @@ An example of the JSON representation is shown below:
...
+
+To create a rudimentarily "parsed" version of the report text, we can infer some structure to the reports from the value of the observation ID suffix defined in section 7.2.4 of the HL7 (v2.7) standard.
+For this process, we:
+1. Filter the `OBX` segments to retain those with a non-empty `OBX-3.1.2` value corresponding to this suffix.
+2. Maintaining the order of the filtered segments, group the segments by suffix.
+3. For each suffix group, join the segments by newlines and store them in the following columns:
+
+| Observation ID Suffix             | Data Lake Column Name       |
+|-----------------------------------|-----------------------------|
+| `ADT`                             | `inferred_addendum_note`    |
```
johnflavin, line 87: I think having a prefix for all of these column names is good but I'm not sure I like `inferred_*`. What are you communicating with that prefix?

What do you think about `report_segment_*`? Just spitballing, we could  probably find something better.

cmoore01: I was thinking it communicates a bit less confidence on our end on the accuracy of these columns. That seems fair because:
1. We can't parse the 2.4 reports like this.
2. There are scenarios where something labeled "ADDENDUM - ..." will get categorized under impression based on OBX-3.

johnflavin: What if we just name them by the segment identifier rather than a human-interpretable name? I think that could convey that these values are straight from the HL7.

So the column names could be `<lowercase abbreviation>_segment`, like `imp_segment`. That doesn't say that this _is_ the impression segment, just that it is whatever was labeled `IMP`.

PR #171, opened 2025-09-23
External ansible

johnflavin:
## Type of change
- [ ] New feature
- [X] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

## Description

We can now configure the ansible playbooks to deploy remotely.
- **Entirely** - run the playbook on a machine where scout will not be installed, like your laptop,
- Or **partially** - run the playbook from a control node

These situations require different inventory configuration; as such I've added some explanatory comments to the inventory example.

Additionally you can run the playbook as a non-privileged user and it will become root. You need to first encrypt your password using `ansible-vault` and add the encrypted password to the inventory.

##### Appsec
There is a security consideration that I want to bring up for discussion. I had to disable TLS cert verification on the connection to the remote `k3s` API.

When installing remotely we use a local kubeconfig file to authenticate against the remote `k3s`; we copy that kubeconfig from the remote after `k3s` is installed. The config contains cert information that can verify the remote server's certificate; that certificate only validates when using the full FQDN (like `tagdev-control-03.nrg.wustl.edu`). However, our tailscale DNS does not work on the full FQDN, it only works when using a simplified domain (like `tagdev-control-03`). Therefore to enable the remote `k3s` connection I disabled certificate verification.

### Backward compatibility
This is backwards compatible with the previous installation workflow, no changes in inventory required.

## Testing
- Wiped `03` cluster
- Ran playbook from my laptop, installed Scout to `03` cluster
- Wiped `03` cluster
- Ran playbook from `tagdev-control-03`, installed Scout to `03` cluster

## Note for reviewers
The main changes are annotating that every task that uses `k8s` or `helm` should run locally, i.e. I added `delegate_to: localhost`. I'm sure there is a better and more efficient way to have done this but I don't know what it is.

There are a few other changes around how we handle looking for files in postgres and superset installs, but those are just streamlining and are not functionally any different.

I also added boolean params `install_helm` and `clone_scout_repo` so you can skip those installs if you need to, especially for a dev install. I didn't document them though.

## Checklist

### DIFF
--- a/ansible/inventory.example.yaml
+++ b/ansible/inventory.example.yaml
@@ -4,7 +4,7 @@ server:
   hosts:
     FQDN-leader.edu:
       ansible_connection: local # set if running ansible on this node. if connecting via ssh omit.
-      ansible_host: leader # set if connecting over ssh AND can connect to different host name than FQDN, else omit
+      ansible_host: leader # Host name to use for ssh connection; omit if `ansible_connection: local` OR ssh can connect using FQDN
       ansible_python_interpreter: /usr/bin/python3
       k3s_control_node: true
       external_url: alt.fqdn.edu # omit to use FQDN-leader.edu
@@ -13,13 +13,13 @@ workers:
   hosts:
     FQDN-worker-1.edu:
       ansible_connection: local # set if running ansible on this node. if connecting via ssh omit.
-      ansible_host: worker-1 # set if connecting over ssh AND can connect to different host name than FQDN, else omit
+      ansible_host: worker-1 # Host name to use for ssh connection; omit if `ansible_connection: local` OR ssh can connect using FQDN
       ansible_python_interpreter: /usr/bin/python3
 gpu_workers:
   hosts:
     FQDN-gpu-1.edu:
       ansible_connection: local # set if running ansible on this node. if connecting via ssh omit.
-      ansible_host: gpu-1 # set if connecting over ssh AND can connect to different host name than FQDN, else omit
+      ansible_host: gpu-1 # Host name to use for ssh connection; omit if `ansible_connection: local` OR ssh can connect using FQDN
       ansible_python_interpreter: /usr/bin/python3
 # The agents group contains nodes that run only the k3s agent.
 # In this example it only includes the workers and gpu_workers groups.


ansible/playbooks/tasks/set_up_kubeconfig.yaml
```diff
@@ -0,0 +1,53 @@
...
+        modified_kubeconfig: '{{ k3s_kubeconfig | combine(cluster_update, recursive=True) }}'
+      vars:
+        cluster_update:
+          clusters:
+            - name: '{{ k3s_kubeconfig.clusters[0].name }}'
+              cluster:
+                server: 'https://{{ ansible_host | default(inventory_hostname) }}:6443'
+                # certificate-authority-data: '{{ k3s_kubeconfig.clusters[0].cluster["certificate-authority-data"] }}'
+                # NOTE: The certificates we have only validate for hostnames that we can't use to reach the machines, so we cannot TLS validate the connection
+                insecure-skip-tls-verify: true
```
kathrynalpert, line 32: We may not need this if the staging server is within the WashU env vs connected via TS, could make it conditional or wait until that's the case to do so.

ansible/inventory.example.yaml
```diff
@@ -3,20 +3,23 @@
 server:
   hosts:
     FQDN-leader.edu:
-      ansible_connection: local
+      ansible_connection: local # set if running ansible on this node. if connecting via ssh omit.
+      ansible_host: leader # set if connecting over ssh AND can connect to different host name than FQDN, else omit
```
kathrynalpert, line 7: Could you clarify this comment? 
Do you mean "Host name to use for ssh connection if different from FQDN; omit if `ansible_connection: local`"?

kathrynalpert (approved):
All the `delegate_to` does feel a bit messy, would probably be ideal to refactor into roles or plays that can run on one node or another, but I think this is a downstream effort.

I'm debating if we should allow user to specify "staging" node and `delegate_to: staging` (vs assuming localhost), but again, I think we can make this change downstream.

ansible/inventory.example.yaml
johnflavin: > Host name to use for ssh connection if different from FQDN; omit if `ansible_connection: local`

This is somewhat better, but the `if different from FQDN` is a sneaky second reason to omit. For full clarity, even though I think you already know this, by default ansible will try to ssh to the host using the FQDN; if that works, you don't need to set `ansible_host`. 

How about
> Host name to use for ssh connection; omit if `ansible_connection: local` OR `ssh` can connect using FQDN

ansible/playbooks/tasks/set_up_kubeconfig.yaml
johnflavin: > We may not need this if the staging server is within the WashU env vs connected via TS

Is the implication that, if we're inside the WashU network but not using the TS network, we would be able to route to the node using the FQDN?

I'm fine adding a boolean to the inventory to control this.

kathrynalpert: Right. Though I don't know for sure that it'll work, so perhaps we should just wait until we're there to add the complexity.

PR #180, opened 2025-10-03
Ansible roles refactor part 1: postgres

johnflavin:
## Type of change
- [ ] New feature
- [ ] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

## Description

Refactor the ansible postgres playbook into a role structure. This can serve as an example of how the rest of the playbooks can be refactored into roles in the future.

## Impact

Right now, not much. But the goal is to lean into ansible best practices to increase maintainability over time.

### Backward compatibility
As far as I can tell this should be backwards compatible with existing scout installs and inventory files. 

## Testing
- I ran the `postgres` playbook by itself on an existing scout install on the `03` cluster
- I wiped the `03` cluster and ran the whole `main` playbook
- Added a small unit test to CI for the `postgres` role

## Note for reviewers
I had a lot of help from Claude Code on this one, but I do think I have a pretty good handle on all the changes here. (Partly because the changes were initially much more wider in scope and I walked them back to something that A. worked and B. was comprehensible to a human reviewer.)

## Checklist
- [X] I have added unit tests, unless this is a test code PR.


andylassiter (approved):

ansible/roles/postgres/templates/postgres_init/README.md
```diff
@@ -0,0 +1,9 @@
+# Postgres Init SQL
```
kathrynalpert, line 1: These are only run when postgres is first installed, so it's probably better to perform postgres db setup within the service's playbook. I don't know if we want to add this caveat here. It's sort of out of scope of this PR.

kathrynalpert (approved):

PR #337, opened 2026-04-07
Remove timestamp header line requirement from HL7 log extractor

kathrynalpert:
### Product

HL7 log files from some sources don't include timestamp lines between messages: they go straight from `<EB>` to `<SB>`. The extractor now handles these files by deriving naming from the log filename date instead of requiring per-message timestamp headers.

### Technical

Removed the timestamp header line parsing from `SplitHl7LogActivityImpl`. HL7 files are now named `{logFileNameNoExtension}_{messageNumber}.hl7` (e.g., `20250814_0.hl7`) and organized in ZIP directories as `YYYY/MM/DD/` instead of `YYYY/MM/DD/HH/`. Deleted `extractTimestamp`, `parseAndValidateTimestamp`, `previousLine` tracking, and related constants (`HEADER_LENGTH`, `MIN_TIMESTAMP_LENGTH`, etc.). The `Hl7LogEntry` record  no longer carries a `headerLine` field.

## Type of change
- [ ] New feature
- [x] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization

N/A

##### Appsec

N/A

### Performance

N/A - slightly less work per message (no timestamp extract/parse)

### Data

- **New files** will use the new naming format (`YYYY/MM/DD/{logFileNameNoExtension}_{num}.hl7`).                                                                                                                                               
- **Existing files** in MinIO retain the old format (`YYYY/MM/DD/HH/{timestamp}_{num}.hl7`). Each file is tracked by its full S3 path in the database, so no reconciliation is needed.                                        
- Messages without timestamp headers (previously marked FAILED) will now be ingested successfully.                                                                                                                            
- The downstream hl7-transformer is unaffected, it treats filenames as opaque and extracts all metadata from HL7 message content (`MSH-7`, etc.).  

### Backward compatibility

**S3 path format change**: new ingestions produce different file paths than previous runs. No migration *required* since paths are stored as full strings in the database, but a reingest might be nice for consistency.

## Testing

- Updated integration test expectations in `TestStatusDatabase.java` to reflect the new path format
- Updated `merge_base.json` and `merge_additions.json` test fixtures with new `source_file` paths
- The `testStatusDbHl7MessageRepeatError` test (renamed to `testStatusDbHl7MessageRepeat`) now expects both duplicate messages to succeed instead of the second failing
- Added `20250814_PORU.log` to extraction test data, a second log file for the same date as `20250814.log`. The `testCurationMerges` test validates the report survives end-to-end through the transformer without filename collision (HL7 files use `20250814_PORU_0.hl7` prefix, not `20250814_0.hl7`). Count-based query fixtures updated accordingly.                                                                                                     
  - Added `20000216_PORU.log` to postgres test data with new `testStatusDbSuffixedLogFileName` test verifying the extractor records the correct suffixed filename in the ingest database 
- Tests pass in CI

## Checklist


johnflavin (approved):
This approach still has a possibility of HL7 collision if there are two log files with the same name in different directories. But, to be fair, that's also true on `main`, not new to this branch. So I think it's a good change.

kathrynalpert:
> This approach still has a possibility of HL7 collision if there are two log files with the same name in different directories. 

Ah, fair point. When we see this, though, they actually _are_ the same file, so I'm going to call this good in that regard.

tests/ingest/staging_test_data/postgres/2000/20000216_PORU.log
```diff
@@ -0,0 +1,10 @@
...
+OBX|1|TX|976487||MICHAEL MARKS, M.D.~~~~********Final Report********~~~ ||||||F<R>
+OBX|2|TX|976487||ACC#  Date Time  Exam~~7795083412 Feb 16 2000 02:30PM BED70260 70260 Skull complete min 4 views~~EXAMINATION:~Skull X-ray, complete series~~FINDINGS:~The complete skull series demonstrates normal cranial vault contour~and thickness. No fracture lines or bony erosions are identified.~The sella turcica is normal in size and configuration. The paranasal~sinuses are clear. The mastoid air cells are well-pneumatized~bilaterally.~~IMPRESSION:~Normal complete skull series. No acute intracranial abnormality~detected.||||||F<R>
+OBX|3|TX|976487||Requested By: Donovan,William C. M.D.~~Dictated By: Michael Marks, M.D. on Feb 16 2000 02:30PM~~This document has been electronically signed by: Michael Marks, M.D. on~Feb 16 2000 02:30PM~ ~~~ ~~~ ||||||F<R>
+OBX||RP|976487|4|7795083412^UK^ZB||||||F<R>
+<EB>
```
cmoore01, line 10: Is the ending intentionally inconsistent with `20250814_PORU.log` ? That one ends in `<EB><R>` versus here it's just `<EB>` . I think the delimiters here look different if it's the last message or not, but for both of these log files it's a single message each.

tests/ingest/src/test/resources/spark_queries.json
```diff
@@ -96,8 +96,8 @@
           },
           "MG": {
             "male_count": 0,
-            "female_count": 9,
```
cmoore01, line 99: Hmm, direct changes to these files means I can no longer regenerate test data for these tests if we needed to. It's possible we wouldn't need to again, but the design here is that the data-generator project is what produced the data for these tests and that project used the data it was making to produce the expected outputs here. Hence, if we need to generate new data, we can automatically produce a new `spark_queries.json` automatically. That would no longer be possible.

kathrynalpert: Would it be better for me to update the data generator to produce the PORU files? Or drop this test?

cmoore01: The approach I've been going with for tests around a more targeted feature is to have the log files in their own directory (e.g. the `patient_ids` directory in the test data). Then the individual test can launch with only the relevant log file(s), and everything stays independent. The main downside with this approach is that it increases the runtime of the tests because you're not just relying on the "generic" ingest that several of the tests all depend on, but I think I would argue that's a shortcoming of how we spin spark up ephemerally and that's the real issue we should fix.

However, given you probably want to check the extractor handling both regular log files and PORU log files at the same time, it makes sense to probably keep it as you have it. I just wanted to point out what we are losing and that it would be harder to update the tests if needed.

cmoore01 (approved):

PR #350, opened 2026-04-23
fix(launchpad): make sign-out actually sign the user out

andrey-embarklabs:
## Summary

Clicking "Sign out" in Launchpad destroyed the NextAuth session and the Keycloak realm session, but the user landed back on `/` still appearing signed in. Root cause was three compounding issues:

1. **HomeClient auto-re-federated** — `HomeClient.tsx` unconditionally calls `signIn('keycloak')` whenever there's no NextAuth session. So the moment the user returned from Keycloak's logout endpoint, launchpad immediately bounced them back to Keycloak, which silently re-federated via the upstream IdP (GitHub session is still alive — we can't kill it).

2. **No `id_token_hint` in the logout URL** — the `/api/auth/signout` route just returned `OAUTH2_PROXY_SIGN_OUT_URL` unchanged. Without `id_token_hint`, Keycloak shows a "Do you want to sign out?" confirmation page (RP-initiated logout spec). Minor UX wart.

3. **Id token wasn't being captured** — the NextAuth `jwt` callback only stored `access_token` and `refresh_token`, so the signout route had no id token to work with.

## Changes

| File | What changed |
|---|---|
| `launchpad/src/lib/auth.ts` | jwt callback also stores `account.id_token` |
| `launchpad/src/types/next-auth.d.ts` | JWT interface gains optional `idToken?: string` |
| `launchpad/src/app/api/auth/signout/route.ts` | Reads the JWT server-side via `getToken()`; appends `id_token_hint` to the logout URL when available |
| `launchpad/src/app/HomeClient.tsx` | Reads a `justLoggedOut` flag from `sessionStorage`; when set, skips the auto-`signIn` and renders a "You are signed out" landing page with an explicit Sign in button |
| `launchpad/src/components/UserDropdown.tsx` | Sets `sessionStorage.setItem('justLoggedOut','1')` before invoking `signOut()` so the flag survives the round-trip to Keycloak |

## Why `sessionStorage` vs a URL query param?

- Survives the full `launchpad → Keycloak → launchpad` browser redirect within the same tab
- Doesn't leak into analytics or bookmarked URLs (like `?loggedOut=1` would)
- Cleared on first read so subsequent page refreshes behave normally (auto-login kicks back in)

## Test plan

- [ ] Sign in via GitHub IdP, click Sign out → land on launchpad with a "You are signed out" page, **no automatic redirect**
- [ ] Click "Sign in" button → normal OIDC flow, user can re-authenticate
- [ ] Refresh the "You are signed out" page → auto-login resumes (flag is single-use)
- [ ] `kcadm get events -r scout -q type=LOGOUT` shows a LOGOUT event for `launchpad-client` after sign-out
- [ ] With Keycloak 18+ and `id_token_hint` passed, the Keycloak confirmation page is skipped (direct redirect)
- [ ] Dev mode (`NEXT_PUBLIC_SKIP_AUTH=true`) still works unchanged

## Related

Depends on the gitops side already setting `OAUTH2_PROXY_SIGN_OUT_URL` to Keycloak's `end_session_endpoint` and the `launchpad-client` having `post.logout.redirect.uris` configured (done in embarklabs-ai/gitops).

### DIFF
--- a/launchpad/src/app/HomeClient.tsx
+++ b/launchpad/src/app/HomeClient.tsx
@@ -447,12 +447,45 @@ export default function HomeClient({ enableChat, enablePlaybooks }: HomeClientPr
   const skipAuth =
     process.env.NODE_ENV === 'development' && process.env.NEXT_PUBLIC_SKIP_AUTH === 'true';
 
-  // Auto-login: redirect to sign in if not authenticated
+  // After an explicit sign-out we suppress auto-login so the user stays signed out
+  // — otherwise the upstream IdP (e.g. GitHub) silently re-federates the session.
+  const [justLoggedOut, setJustLoggedOut] = useState(false);
   useEffect(() => {
-    if (!skipAuth && status !== 'loading' && !session) {
+    if (typeof window === 'undefined') return;
+    if (sessionStorage.getItem('justLoggedOut') === '1') {
+      setJustLoggedOut(true);
+      sessionStorage.removeItem('justLoggedOut');
+    }
+  }, []);
+
+  // Auto-login: redirect to sign in if not authenticated (unless user just signed out)
+  useEffect(() => {
+    if (!skipAuth && !justLoggedOut && status !== 'loading' && !session) {
       signIn('keycloak');
     }
-  }, [status, session, skipAuth]);
+  }, [status, session, skipAuth, justLoggedOut]);
+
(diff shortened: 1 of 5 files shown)


andylassiter:
Closing, not an error, need to get OA2P working in your dev env.

PR #353, opened 2026-04-30
Add native S3 read support to HL7 log extractor

kathrynalpert:
### Product

Scout can now ingest HL7 log files directly from S3-compatible object storage — both AWS S3 and the existing in-cluster MinIO — by passing an `s3://` URI as the ingestion workflow's `logsRootPath`. Previously, the extractor could only read logs from a filesystem path, so the HL7 logs had to be present on a disk mounted into the extractor pod. Now, HL7 logs can reside in S3, which paves the path for an event-driven workflow that writes HL7 to S3 for Scout to consume. As with existing HL7 logs on disk, operators can point ingestion at either a single log object or a prefix of many. **Existing on-prem MinIO deployments are unaffected and need no configuration changes.**

### Technical

The extractor previously treated `logsRootPath` as a local filesystem path. This PR makes it scheme-aware and adds a native S3 read path that targets MinIO and AWS S3 through the same code:

- `FindHl7LogsActivityImpl` dispatches on the URI scheme. For `s3://` inputs, it uses a new `FileHandler.isFile` (an S3 `HeadObject`) to decide whether the URI is a single key (fast path) or a prefix to list; filesystem inputs keep their existing listing behavior.
- `SplitHl7LogActivityImpl` serves both filesystem and S3 sources: the log is read through a `BufferedReader` (ISO-8859-1, matching existing HL7 parsing) so the whole file is never held in memory, and the Temporal activity heartbeats after each parsed HL7 message. For S3, closing the stream returns the connection to the SDK pool. 
- `S3Config` makes `s3.endpoint` optional and adds `s3.path-style-access`. With no endpoint configured, the SDK uses the standard regional AWS endpoint; `path-style-access` defaults to `false` (AWS virtual-hosted style). The Ansible `extractor` role sets `path-style-access` from `aws_deployment`, so MinIO deployments get `true` and continue to work unchanged on upgrade.
- Adds the AWS STS SDK dependency, required at runtime so the default credentials provider chain can use `WebIdentityTokenCredentialsProvider` (IRSA) on AWS.

## Type of change
- [x] New feature
- [x] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization

No user roles are added or modified, and nothing changes about the data Scout users can see or the actions they can take.

On AWS, the extractor obtains S3 credentials via IRSA (a scoped IAM role assumed through `WebIdentityTokenCredentialsProvider`), which supports least-privilege, bucket-scoped access instead of long-lived static keys; on MinIO the existing static credentials (from `s3-secret`) are used unchanged. `logsRootPath` is an admin-settable input to the ingestion workflow (vs end-user-controllable value), so the S3 read path is not an end-user path-traversal or data-exposure vector).

##### Appsec

- Object lookups fail safe: `isFile` treats both `NoSuchKeyException` and a generic `S3Exception` carrying HTTP 404 as "not a single object" and falls back to prefix listing, so a malformed or missing key degrades cleanly instead of throwing. 
- Stream reads use a fixed ISO-8859-1 charset, matching prior parser behavior. 
- Streaming reads reduce data-at-rest exposure: HL7 logs are read from object storage and parsed in memory.

### Performance

HL7 log files can be tens of MB at production scale. Streaming through a `BufferedReader` reads in chunks and never materializes the full file in memory or on disk, and the per-message heartbeat keeps the Temporal activity alive while parsing large files. S3 access reuses the configured connection pool (`s3.max-connections`, default 50), and each stream is closed as soon as its file is parsed so the connection returns to the pool.

### Data

Input edge cases considered: a single-object URI versus a prefix is disambiguated by `HeadObject`; 404s on the head request (both `NoSuchKeyException` and a generic `S3Exception` with status 404) fall back to listing; leading-slash keys are normalized consistently so writes never create `/`-prefixed objects on AWS S3; HL7 bytes are decoded as ISO-8859-1 to match the existing parser. The date-based log filter is preserved identically across the filesystem and S3 paths.

### Backward compatibility

No data-model or REST API changes. Configuration: `s3.endpoint` becomes optional (it was previously hard-coded to the in-cluster MinIO URL), and a new `s3.path-style-access` property is added with a default of `false`; the Ansible `extractor` role derives `path-style-access` from `aws_deployment`, so existing MinIO deployments receive `true` and behave exactly as before.

## Testing

Unit tests added/updated in the `hl7log-extractor` module:

- `FindLogFileTest` covers S3 prefix listing, the date filter, the single-file fast path, a prefix supplied without a trailing slash, and the `isFile` 404 fallbacks for both `NoSuchKeyException` and a generic `S3Exception` with status 404.
- `FileHandlerImplTest` covers `normalizeS3Key` behavior across put/get/read/delete/list.

The extractor is additionally exercised end-to-end by the existing CI jobs (`build-and-upload (hl7log-extractor)`, `deploy-and-test`, `smoke-test`). 

Manual verification against a live AWS S3 bucket with IRSA.

## Note for reviewers
This should have zero impact on existing WashU Scout installations.

## Checklist
- [ ] My code adheres to the coding and style guidelines of the project (and I've run `pre-commit run --all-files`)


johnflavin (approved):
LGTM

PR #357, opened 2026-05-06
Parameterize spark.sql.shuffle.partitions; reduce to 8 for CI

cmoore01:
This came up as an optimization to make in the tests when adding a new test for patient IDs. By overriding `spark.sql.shuffle.partitions` for the very small test data used in CI, we get significantly faster performance. The default stays at 200 (no production behavior change).

### Product
N/A

### Technical
Trivial ansible template tweak

## Type of change
- [ ] New feature
- [X] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
N/A

##### Appsec
N/A

### Performance
Should be a no-op change in production, but enable faster CI.

### Data
N/A

### Backward compatibility
N/A

## Testing
Previous CI tests on branch with bigger scope gave promising results. Test run when submitting this PR will be the real test.

## Checklist

### DIFF
--- a/ansible/roles/scout_common/templates/spark-defaults.conf.j2
+++ b/ansible/roles/scout_common/templates/spark-defaults.conf.j2
@@ -15,6 +15,7 @@ spark.sql.catalog.spark_catalog org.apache.spark.sql.delta.catalog.DeltaCatalog
 spark.hadoop.fs.s3a.path.style.access {{ 'true' if not aws_deployment else 'false' }}
 spark.hadoop.hive.metastore.uris {{ spark_hive_metastore_uri | default(hive_metastore_endpoint_readonly) }}
 spark.sql.warehouse.dir {{ delta_lake_path }}
+spark.sql.shuffle.partitions {{ spark_sql_shuffle_partitions | default(200) }}
 spark.driver.extraJavaOptions -Divy.cache.dir=/tmp -Divy.home=/tmp
 spark.executor.memory {{ spark_memory | default('1G') }}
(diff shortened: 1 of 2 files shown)


kathrynalpert (approved):
~~Is 200 an appropriate default for the non-CI-test workflows we run?~~ (Sorry, I'm moving this to the LOC where it applies)

ansible/roles/scout_common/templates/spark-defaults.conf.j2
```diff
@@ -15,6 +15,7 @@ spark.sql.catalog.spark_catalog org.apache.spark.sql.delta.catalog.DeltaCatalog
 spark.hadoop.fs.s3a.path.style.access {{ 'true' if not aws_deployment else 'false' }}
 spark.hadoop.hive.metastore.uris {{ spark_hive_metastore_uri | default(hive_metastore_endpoint_readonly) }}
 spark.sql.warehouse.dir {{ delta_lake_path }}
+spark.sql.shuffle.partitions {{ spark_sql_shuffle_partitions | default(200) }}
```
kathrynalpert, line 18: Is 200 an appropriate default for the non-CI-test workflows we run?

cmoore01: That's a good question. From some cursory reading I did around this, it sounds like the real way to answer that question is to measure it rather than logically deduce it. That's not something we've done in production. That said, the default spark provides is 200, so maybe it would be clearer here to omit rather than default to spark's default value if the ansible var isn't specified.

PR #360, opened 2026-05-07
Add trino-rw RW Trino instance and harden Trino secret handling

cmoore01:
### Product
Changes nothing from user perspective.

### Technical
Two related changes:

1. Stand up a small read-write Trino release (trino-rw) in scout-extractor so the hl7-transformer (in a future PR) can issue write DDL like CREATE OR REPLACE VIEW. The user-facing Trino in scout-analytics is read-only by design (delta.security=READ_ONLY, read-only metastore endpoint, reader S3 creds), so the transformer can't reuse it.

   - Single-coordinator/1-worker Helm release reusing s3_lake_writer and the writable Hive metastore endpoint — no new credentials.
   - fullnameOverride keeps Service names matching the release name.
   - Catalog config injects S3 credentials via envFrom + ${ENV:VAR} substitution, so the rendered ConfigMap holds only references.
   - Chart-native networkPolicy.enabled gates ingress: hl7-transformer in scout-extractor (port 8080) and Prometheus server in scout-monitoring (port 5556). Coordinator <-> worker traffic is allowed automatically by the chart.
   - Pre-wires TRINO_HOST/PORT/USER/CATALOG/SCHEME env vars on the transformer pod. Inert until the add_epic_views Trino integration ships in a follow-up PR.

2. Apply the same Secret-based S3 credential handling to the existing user-facing Trino in scout-analytics. Previously its catalog config inlined s3_lake_reader_secret directly, which the chart rendered into a ConfigMap (not encrypted at rest, broader RBAC visibility than Secrets). Now reader creds also flow through a trino-s3 Secret via envFrom + ${ENV:VAR}, matching the new RW deployment.

CI tests (tests/network/network-policy-tests.sh):
- Verifies the trino-rw NetworkPolicy: pods in scout-analytics or mislabeled pods in scout-extractor are denied; transformer-labeled pods in scout-extractor are allowed. Uses kubectl run + retry-loops to handle kube-router's ipset reconciliation lag, runs in parallel for ~25s total.

## Type of change
- [X] New feature
- [X] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [X] Test update

##### Authorization
N/A

##### Appsec
This is the part of this change that is most significant to think about. This adds a new trino instance that can be used to make arbitrary writes, without auth. However, it should be locked down by NetworkPolicy such that only prometheus and hl7-transformer can talk to it. I ran some `kubectl` commands manually to verify that which then became the new network tests to verify it in CI.

### Performance
Usage of new trino instance is trivial in scale.

### Data
N/A

### Backward compatibility
Should be backwards compatible

## Testing
Tested manually with upcoming work.

## Note for reviewers
Please double check the security aspect here.

## Checklist

### DIFF
--- a/ansible/roles/scout_common/defaults/main.yaml
+++ b/ansible/roles/scout_common/defaults/main.yaml
@@ -102,6 +102,10 @@ extractor_namespace: '{{ scout_extractor_namespace }}'
 
 # Analytics & user services
 trino_namespace: '{{ scout_analytics_namespace }}'
+trino_rw_namespace: '{{ scout_extractor_namespace }}'
+trino_rw_release_name: trino-rw
+trino_rw_endpoint_host: '{{ trino_rw_release_name }}.{{ trino_rw_namespace }}'
+trino_rw_endpoint_port: 8080
 superset_namespace: '{{ scout_analytics_namespace }}'
 jupyter_namespace: '{{ scout_analytics_namespace }}'
 chatbot_namespace: '{{ scout_analytics_namespace }}'
@@ -182,7 +186,10 @@ ingest_postgres_table_name: ingest
 hive_metastore_instance: hive-metastore
 hive_metastore_instance_readonly: hive-metastore-readonly
 
-# Trino - this is a dummy username since all Trino access is RO
+# Trino RO username — the user-facing Trino has no authN; writes are blocked
+# at the connector level (delta.security=READ_ONLY) and by the read-only
+# metastore + S3 credentials. The RW Trino instance (trino-rw) accepts
+# writes and is gated to the hl7-transformer pod via NetworkPolicy.
 trino_user: trino
 
 # Shared environment for Kubernetes operations on cluster nodes
--- a/ansible/roles/trino/templates/values.yaml.j2
+++ b/ansible/roles/trino/templates/values.yaml.j2
@@ -10,6 +10,14 @@ server:
 serviceAccount:
   create: false
   name: '{{ lake_reader_service_account_name }}'
+{% else %}
+# Inject S3 credentials from a Secret. The Secret is created by deploy.yaml
+# before this chart is installed. The catalog config below references these
+# via Trino's ${ENV:VAR} substitution syntax, so the rendered catalog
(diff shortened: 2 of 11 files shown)


ansible/roles/scout_common/defaults/main.yaml
```diff
@@ -102,6 +102,10 @@ extractor_namespace: '{{ scout_extractor_namespace }}'
 
 # Analytics & user services
 trino_namespace: '{{ scout_analytics_namespace }}'
+trino_rw_namespace: '{{ scout_extractor_namespace }}'
```
johnflavin, line 105: Do we want this rw trino to be in a different namespace from the ro trino? I guess it doesn't matter, really. This way they're nearby the consumer.

Not sure how to feel about this, but I wanted to flag.

ansible/roles/trino/templates/values.yaml.j2
```diff
@@ -71,8 +79,8 @@ catalogs:
     fs.native-s3.enabled=true
     s3.region={{ s3_region }}
 {% if not aws_deployment %}
-    s3.aws-access-key={{ s3_lake_reader }}
-    s3.aws-secret-key={{ s3_lake_reader_secret }}
+    s3.aws-access-key=${ENV:S3_ACCESS_KEY}
+    s3.aws-secret-key=${ENV:S3_SECRET_KEY}
```
johnflavin, line 83: What does this do? Is it saying we read the s3 values from the pod's environment variables, which get set from the `trino-s3` secret above? That seems better than putting them into the helm values like we were doing.

johnflavin (approved):

ansible/roles/trino/templates/values.yaml.j2
cmoore01: Right, previously the way they were baked in meant they got created in a ConfigMap

ansible/roles/scout_common/defaults/main.yaml
cmoore01: Good question. It gives us some extra defense in making sure a notebook can't access it, so that's a nice little benefit.

PR #362, opened 2026-05-09
Orthanc aws cnpg

spetrick:
### Product
This change allows usage of Orthanc for testing and development purposes with a Postgres backend using CNPG. It also exposes LoadBalancerClass and service annotations for the Dicom service to allow deploying on Cloud infrastructure.

### Technical
Added a template to the Orthanc helm chart for deploying a CNPG cluster
Added a switch to the Orthanc Ansible role to allow choosing between CNPG and SQLite
Updated config and values in the Orthanc Ansible role to support the new switch.
Added a configuration sample to the example inventory

## Type of change
- [x] Improvement
- [ ] Bug fix

##### Authorization

There are no impacts to user roles. Orthanc is intended for development and testing and has weak authentication defaults that can be overridden in Ansible.

##### Appsec

This change allows deployment of Orthanc in Cloud Hosted environments. The biggest security risk involves allowing public access to DIMSE endpoints.

### Performance

I expect Orthanc to be used for testing and development and have deployed with 3 replicas and a reasonable sample of data.

### Backward compatibility

SQLite remains the default mode so any existing deployments using Orthanc shouldn't see any unexpected changes.
## Testing

I deployed the Orthanc Helm chart using both the built-in SQLite and the new CNPG backend and ensured the installation was successful and the service accessible within a cloud environment on AWS.
## Checklist


ansible/roles/orthanc/templates/orthanc-conf.json.j2
```diff
@@ -6,6 +6,20 @@
...
     "IngestTranscodingOfCompressed" : false,
+{%- if orthanc_postgres_mode != 'disabled' %}
+    "PostgreSQL": {
+        "EnableIndex": true,
+        "EnableStorage": false,
+        "Host": "orthanc-cnpg-rw.{{ orthanc_namespace }}.svc.cluster.local",
+        "Port": {{ orthanc_postgres_port }},
+        "Database": {{ orthanc_postgres_database | to_json }},
+        "Username": {{ orthanc_postgres_user | to_json }},
+        "Password": {{ orthanc_postgres_password | to_json }},
```
johnflavin, line 17: Why the use of `to_json`? Here, these are surely just strings. You could wrap them in quotes, like `"{{ orthanc_postgres_database }}"`.

ansible/roles/orthanc/templates/values.yaml.j2
```diff
@@ -25,7 +25,10 @@ securityContext: {}
 service:
   type: ClusterIP
   http: 8042
-  dicom: {{ orthanc_dicom_port }}
+  dicom:
+    port: {{ orthanc_dicom_port }}
+    annotations: {{ orthanc_dicom_service_annotations | to_json }}
```
johnflavin, line 30: This use of `to_json` is somewhat more defensible, as you are getting an object here. But wouldn't the value already be yaml, and could just be left as yaml?

ansible/roles/orthanc/templates/values.yaml.j2
```diff
@@ -69,8 +72,34 @@ volumeMounts:
     readOnly: false
     mountPropagation: HostToContainer
 
+postgres:
+  cluster:
+    affinity: {{ orthanc_postgres_cluster_affinity | to_json }}
+    annotations: {{ orthanc_postgres_cluster_annotations | to_json }}
```
johnflavin, line 78: Again, lots of uses of `to_json` here and elsewhere in this file that I'm not sure about.

ansible/roles/orthanc/templates/values.yaml.j2
```diff
@@ -69,8 +72,34 @@ volumeMounts:
...
+    parameters:
+      max_connections: {{ orthanc_postgres_max_connections }}
+      shared_buffers: {{ orthanc_postgres_shared_buffers }}
+    enableSuperuserAccess: true
+    credentials:
+      # Always use an external secret for production deployments.
+      externalSecret: orthanc-cnpg-app-credentials
+      database: "{{ orthanc_postgres_database | to_json }}"
+      username: {{ orthanc_postgres_user | to_json }}
+      password: {{ orthanc_postgres_password | to_json }}
```
johnflavin: This values file hard-codes the name of the external secret we've already created in the ansible but also passes in the username and password that the chart uses if we haven't created an external secret. We don't need to do both. You should leave off the username and password here.

ansible/roles/orthanc/tasks/deploy.yaml
```diff
@@ -39,6 +39,30 @@
...
 
+- name: Validate Orthanc Postgres credentials
+  when: orthanc_postgres_mode == 'cnpg'
+  ansible.builtin.assert:
+    that:
+      - orthanc_postgres_password is defined
+      - orthanc_postgres_password | length > 0
+    fail_msg: 'orthanc_postgres_password is required when orthanc_postgres_mode == cnpg'
+
+- name: Create Orthanc CNPG credentials secret
```
johnflavin, line 50: This secret gets created in the helm chart as well, in `helm/orthanc/templates/cnpg_cluster.yaml`. (I've got a few comments about this, but just noting it here.)

helm/orthanc/templates/cnpg_cluster.yaml
```diff
@@ -0,0 +1,57 @@
...
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: orthanc-cnpg-app-credentials
+type: kubernetes.io/basic-auth
+stringData:
+  username: {{ .Values.postgres.cluster.credentials.username }}
+  password: {{ .Values.postgres.cluster.credentials.password }}
+{{- end }}
```
johnflavin: This is where the secret gets created if we didn't provide an external secret. (I've got a few comments about this, but just noting it here.)

helm/orthanc/templates/cnpg_cluster.yaml
```diff
@@ -0,0 +1,57 @@
...
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+
+spec:
+  instances: {{ .Values.postgres.cluster.instances }}
+  {{- with .Values.postgres.cluster.affinity }}
+  affinity:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  imageName: ghcr.io/cloudnative-pg/postgresql:{{ .Values.postgres.cluster.version }}
```
johnflavin: Would be nice to have the image configurable as well. I believe scout's ansible already has a variable for this somewhere in scout_common.

ansible/roles/orthanc/templates/orthanc-conf.json.j2
spetrick: My understanding is that, passwords specifically, it's easy to have characters that would break a simple quote, but would be properly escaped by using to_json. For the others it's less necessary, but still catches things like null values properly.

ansible/roles/orthanc/templates/values.yaml.j2
spetrick: to_yaml | indent would work just fine as well, but to_json is safer because it handles special characters and null values better.

ansible/roles/orthanc/templates/values.yaml.j2
spetrick: Same as above in that it handles null values better than the alternatives. to_yaml | indent would probably work just fine as well, but to_json is safer.

helm/orthanc/templates/cnpg_cluster.yaml
spetrick: I don't see a CNPG image override anywhere, but making this configurable is simple enough.

helm/orthanc/templates/cnpg_cluster.yaml
spetrick: I did this out of habit for creating helm charts with functional defaults and the ability to properly secure resources. I'm okay pulling it out of here since that's not really how the helm charts in Scout work. Would that be your preference?

helm/orthanc/templates/cnpg_cluster.yaml
johnflavin: Hmm, you’re right that it isn't configured anywhere. We just specify the version of the operator chart. 

But that makes me wonder whether you need to have this here at all. Is there a reason you want to specify the image or (more likely) the version rather than leave it empty and let the operator apply its defaults?

spetrick: It's probably not necessary, especially for Orthanc. Just habit on my end to have control over which images are deployed for auditing purposes.

kathrynalpert (approved):

PR #366, opened 2026-05-12
Launchpad UI improvements and dependency updates

kathrynalpert:
### Product

The Launchpad landing page has been redesigned to remove the marketing "Welcome to Scout" headline.

The visual treatment shifts from the heavier gradient cards / colored borders to a calmer slate palette: neutral white cards with subtle colored icon tints and hover accents, lower-contrast typography, and unified iconography. Admin card content gains short descriptions, and the Chat card now renders first when enabled. The header is restructured: a left-aligned Scout brand strip with an environment label (e.g. "Scout / synthetic") sits opposite the existing TopBar. The footer is rewritten to credit Washington University with a link to the GitHub repo and an optional "Deployed by {institution}" line for downstream deployers.

Screenshot 2026-05-12 at 12 40 25 PM

### Technical

- Standard styling updates
- New ENV vars for `scout_env` and `deployer_name`
- Dependency updates so `npm audit` reports `found 0 vulnerabilities`

## Type of change
- [ ] New feature
- [x] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization

New `NEXT_PUBLIC_DEV_ADMIN` bypass in `AdminSection` is explicitly gated on `NODE_ENV === 'development'`, so it is impossible to enable in a production build regardless of how the env var is set. Server-rendered Next.js bundles in production will not include this code path. (Note: this was included to allow local `npm run dev` testing of launchpad)

##### Appsec

All external links use `target="_blank"` with `rel="noopener noreferrer"`. `deployerName` is rendered through React's default text interpolation (not `dangerouslySetInnerHTML`), so any institution name supplied via inventory is automatically escaped. No new outbound network calls, no new cookies, no new auth surfaces.

This PR also adds an `overrides.postcss: "^8.5.10"` entry in `launchpad/package.json` that patches GHSA-qx2v-qp2m-jg93 (PostCSS XSS via unescaped `</style>` in CSS stringify output). The vulnerable PostCSS appeared transitively in two places — bundled by `next@16.2.6` (8.4.31) and by `@tailwindcss/postcss@4.1.16` (8.5.3); the override forces both to dedupe to 8.5.14. The Next-side fix has only landed on `next@16.3.0-canary.6+` with no stable release yet, so this override is the intended interim mitigation; an inline `//overrides` comment in `package.json` notes that it should be removed once a stable Next release ships the patch. `npm audit` reports `found 0 vulnerabilities` after the override.

### Performance

N/A

### Data

N/A

### Backward compatibility

Compatible

## Testing

Deployed to `dev04` and run locally.

## Checklist


andylassiter (approved):

ansible/roles/launchpad/defaults/main.yaml
```diff
@@ -8,6 +8,15 @@ launchpad_helm_chart_ref: '{{ scout_repo_dir }}/helm/launchpad'
 # Custom Launchpad image (override from inventory for CI/CD)
 launchpad_image_repository: ghcr.io/washu-tag/launchpad
 launchpad_image_tag: latest
+launchpad_image_pull_policy: IfNotPresent
```
johnflavin, line 11: I understand using this `IfNotPresent` for dev but I'm not sure about setting it as the default. Typically we would want the `latest` image to always pull updates, which is the k3s default. 

This inspired my [slack message in #scout](https://cil-wustl.slack.com/archives/C07AM78J7PT/p1778676461813389) about pushing dev images to harbor.

johnflavin (approved):

ansible/roles/launchpad/defaults/main.yaml
kathrynalpert: I waffled on this, too. 

My thinking is that cloud dev and/or gitops is likely to push us away from `latest` tags, which has always felt a bit like an antipattern. In dev, you're already modifying the inventory to change the repo/tag, so swapping from `IfNotPresent` to `Always` is an easy step.

But it doesn't have to change in this PR, I can adapt my dev process. I'll see how the Slack discussion goes and make a call.

Update: the helm chart currently sets `IfNotPresent`, so if we want to use the `latest` --> `Always` I'll need to remove that.

kathrynalpert: Based on the conversation in Slack, I think we might want to make a more repo-wide decision. Given the current state of the helm chart, I'm going to proceed with this PR as-is (adding the param, which means no functional change from prior to the param since the helm default was already `IfNotPresent`).

PR #368, opened 2026-05-13
Migrate log collection from promtail to alloy

johnflavin:
### Product

No user-facing changes. Log collection source was changed, but should be transparent to users. (And only admins can see logs at all.)

### Technical

- Replaces deprecated grafana promtail with grafana alloy for log collection.
- Adds a new `alloy` Ansible role to deploy alloy. (Promtail installation had been embedded in the `loki` role.)
- Alloy is configured to read pod logs via the kubelet pod-logs API (`loki.source.kubernetes`). This is a change from how promtail read logs via hostPath mounts to `/var/log/pods` and `/var/lib/docker/containers`. 
- Alloy config is set to reproduce promtail's label set (`namespace`, `pod`, `container`, `app`, `instance`, `component`, `node_name`, `job=<ns>/<app>`), which means we don't need to change any of our grafana dashboards
- Adds a migration uninstall task in the `alloy` role to clean up legacy promtail releases (with a comment saying we can remove this later). It's idempotent, i.e. ok to run more than once.
- Extends `scout_common/deploy_helm_chart` task with a `helm_chart_state` parameter so helm install and uninstall route through the same shared task, which takes care of the air-gapped delegate/kubeconfig handling.

## Type of change
- [ ] New feature
- [x] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization

N/A

##### Appsec

N/A

### Performance

Should be similar to promtail, which is basically not noticable. Using the kubelet API for log tailing rather than hostPath mounts is a slight increase in load, but likely to be negligible at Scout's size.

### Data

Verified that the labels are the same as with promtail. During dev I had both promtail and alloy running together to check consistency, and loki merged them both into a single log stream.

A caveat from Claude: The `loki.source.kubernetes` startup replay produces a brief one-time burst of `too_far_behind` rejections for old history already shipped by Promtail; no data loss.

Translation: using the kubelet API as the log source means when pods restart we will get some duplicate / old log messages flowing into loki, but loki rejects them so we should be fine.

### Backward compatibility

Included a migration task to remove existing promtail installations.

## Testing

Used Claude to verify deployment on dev03, this is its summary:
1. Audited live Loki labels and dashboard queries to define the contract.
2. Parallel-deployed alloy alongside promtail and confirmed all River components healthy.
3. Verified byte-identical label parity across multiple workloads via `/loki/api/v1/series`.
4. Removed promtail and confirmed continued log flow with `~0` rejection rate. 

## Note for reviewers

N/A

## Checklist


kathrynalpert (approved):

ansible/group_vars/all/versions.yaml
```diff
@@ -99,11 +99,11 @@ ollama_image_tag: 0.22.1
...
 prometheus_helm_chart_version: ~27.52.0
 
-# Monitoring & Observability - Loki & Promtail
+# Monitoring & Observability - Loki & Alloy
 # renovate: datasource=helm registryUrl=https://grafana.github.io/helm-charts depName=loki
 loki_helm_chart_version: ~6.49.0
-# renovate: datasource=helm registryUrl=https://grafana.github.io/helm-charts depName=promtail
-promtail_helm_chart_version: ~6.17.0
+# renovate: datasource=helm registryUrl=https://grafana.github.io/helm-charts depName=alloy
+alloy_helm_chart_version: ~1.5.1
```
andylassiter: Is this the latest version? I see a [1.8.1](https://artifacthub.io/packages/helm/grafana/alloy).

johnflavin: Good catch. Bumped.

PR #369, opened 2026-05-13
Scout dashboards via helm chart

kathrynalpert:
### Product

Ships Scout's Superset dashboards as a versioned, gitops-friendly Helm chart with site-configurable bundles.

- The Scout core dashboard installs by default. **Quality / TAT** and **Follow-up Detection** dashboards are now opt-in per inventory.
- The Scout core dashboard gains a diagnosis table and "HL7 messages", next to the existing "Count" tile so operators can see raw HL7 throughput alongside unique-exam scale.
- Local Superset admin user is renamed from "Superset Admin" to "Scout Admin" on fresh installs.

Screenshot 2026-05-13 at 12 32 59 PM
Screenshot 2026-05-13 at 12 34 06 PM
Screenshot 2026-05-13 at 12 33 50 PM
Screenshot 2026-05-13 at 12 33 39 PM

### Technical

Moves Superset dashboard assets out of `analytics/superset/` and into a new `helm/scout-dashboards/` chart. The chart owns the `dashboard-config` ConfigMap and a post-install/upgrade Job that runs `superset import-dashboards` plus a force-overwrite Python pass (because the importer skips existing UUIDs). The Job's `checksum/config` annotation is keyed to the rendered ConfigMap, so any analytics-file edit or value flip causes Helm to recreate the Job and re-run the import (no Ansible-side change detection).

Dashboard assets are partitioned into bundles by subdirectory:

```
files/analytics/
  charts/{core,quality,followup}/*.yaml
  dashboards/{core,quality,followup}/*.yaml
  datasets/Scout_Data_Lake/{core,followup}/*.yaml
```

And values control what ships:
```yaml
bundles:
  enabled: [core]          # built-in bundles to include
```
(wired through Ansible inventory as `scout_dashboard_bundles`)

The upgrade path includes a one-time migration task that deletes a pre-Helm `dashboard-config` ConfigMap if it lacks Helm release annotations.

Dataset hygiene during this PR:
- `diagnoses_flat` (virtual `CROSS JOIN UNNEST` against raw `reports`) from old demos (not in IaC) → `reports_dx` (physical view from the lake, built on `reports_latest`).
- `reports` (raw curated, every HL7 version per exam) → `reports_latest` (one row per exam) for every existing chart; UUID kept stable so the swap is a pure `table_name` edit on the existing dataset row.
- A separate new dataset pointing at raw `reports` backs the "HL7 messages" scale tile.

dev04 inventory was updated to opt into all three bundles: `scout_dashboard_bundles: [core, quality, followup]` for testing.

## Type of change
- [ ] New feature
- [x] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization

N/A

##### Appsec

N/A

### Performance

- One extra Job pod, negligible in our install and much better than depending on the core Superset job (we control when it runs, TTL, etc)
- Dataset switch from `reports` → `reports_latest` reduces per-query work for core chart

### Data

N/A

### Backward compatibility

- Existing dashboard / chart / dataset rows in Superset's metadata DB are updated in place by UUID (via the force-overwrite Python pass) — no orphan rows on upgrade.
- Old Superset clusters get a one-time migration step in the `superset` role that deletes the legacy `dashboard-config` ConfigMap when it lacks Helm release annotations

## Testing

Manual on `dev04` ~~this was a bit iterative and I didn't wipe clean and start fresh.~~ I did decide to do an uninstall/reinstall, as well.

## Note for reviewers

I snuck in a couple unrelated changes to specify resources in CI for voila and grafana so `smoke-tests` pass.

## Checklist


helm/scout-dashboards/templates/import-job.yaml
```diff
@@ -0,0 +1,99 @@
...
+              cd /app
+              python /app/dashboard-config/import.py /app/dashboard-config/analytics
+          resources:
+            {{- toYaml .Values.resources | nindent 12 }}
+          volumeMounts:
+            - name: superset-config
+              mountPath: /app/pythonpath
+              readOnly: true
+            - name: dashboard-config
+              mountPath: /app/dashboard-config
```
johnflavin, line 63: I think this path gets mounted read-only since it's from a configmap even though it's not explicitly marked `readOnly: true`; see https://github.com/kubernetes/kubernetes/issues/62099. That means the steps in the script above which cd to this directory and then copy in any extra configmaps would fail. Should probably prepare these built-in configmaps + the extra in an emptydir mount.

helm/scout-dashboards/templates/import-job.yaml
```diff
@@ -0,0 +1,99 @@
...
+            secretName: {{ .Values.supersetRelease.configSecret }}
+        - name: dashboard-config
+          configMap:
+            name: dashboard-config
+            items:
+              {{- include "scout-dashboards.configMapItems" . | nindent 14 }}
+        {{- range $i, $cm := .Values.extraConfigMaps }}
+        - name: extra-{{ $i }}
+          configMap:
+            name: {{ $cm }}
```
johnflavin: I believe mounting the extra configmaps this way won't get them to the right place. (This comment really isn't just on this part, it touches where these configmaps are mounted and copied, but I am just putting it all here.)

This mounts the contents of each configmap named in `extraConfigMaps` to `/app/dashboard-config-extras/<i>/<key>`. The key can't contain any `/`. From https://kubernetes.io/docs/concepts/configuration/configmap/:

> Each key under the `data` or the `binaryData` field must consist of alphanumeric characters, `-`, `_` or `.`.

Then in the script when they get copied, they would land in
```
analytics/<key>
```
But superset won't find them there. It expects them to be in something like
```
analytics/chart/<key>
```
The comment above the script says to name them something like `chart/foo.yaml` but that can't happen in a configmap key.

You could give these keys a more complex path here when defining the volume, something like
```
{{- range $i, $cm := .Values.extraConfigMaps }}
- name: extra-{{ $i }}
  configMap:
    name: {{ $cm.name }}
    {{- with $cm.items }}
    items:
      {{- toYaml . | nindent 6 }}
    {{- end }}
{{- end }}
```
But that's only if the `extraConfigMaps` were defined not just as a configmap name but an object that specified for each key in the cm what path it should go to. For example:
```
extraConfigMaps: []
# - name: site-overrides
#   items:
#     - key: my_chart.yaml          # flat key in the ConfigMap
#       path: charts/my_chart.yaml  # where it lands in the volume
#     - key: my_dashboard.yaml
#       path: dashboards/my_dashboard.yaml
#     - key: my_dataset.yaml
#       path: datasets/Scout_Data_Lake/my_dataset.yaml
```

helm/scout-dashboards/README.md
```diff
@@ -0,0 +1,177 @@
+# Scout Dashboards
```
andylassiter, line 1: Naming nit, Scout Dashboards (`scout-dashboards`) is ambiguous. We also have Grafana dashboards that we will need to move into a helm chart to make GitOps friendly. Maybe Scout Analytics Dashboards (`scout-analytics-dashboards`) ?

helm/scout-dashboards/templates/import-job.yaml
kathrynalpert: Yes indeed... but I decided that we should wait on that "extra configmaps" feature until we need it. I got it fixed and was testing, but that required making a manual cm, and it just felt premature.

helm/scout-dashboards/README.md
kathrynalpert: I disagree on this. I think Scout Dashboards are the user-facing, analytics dashboards that show Scout data. The Grafana Dashboards are admin-only and can have the qualified name.

ansible/roles/superset/tasks/deploy.yaml
```diff
@@ -150,3 +98,51 @@
...
+# YAML edit or family toggle change changes the hash, Helm replaces the Job
+# (per before-hook-creation delete policy), and the import re-runs. No
+# Ansible-side change detection.
+- name: Deploy scout-dashboards chart
+  ansible.builtin.include_role:
+    name: scout_common
+    tasks_from: deploy_helm_chart
+  vars:
+    helm_chart_name: scout-dashboards
+    helm_chart_ref: '{{ scout_repo_dir }}/helm/scout-dashboards'
```
johnflavin: Nit: In the other local helm chart installations (launchpad, voila) we have a `helm_chart_ref` variable so the location can be overridden if the need arises.

johnflavin (approved):

PR #370, opened 2026-05-13
Patient Mapping Table

cmoore01:
### Product
Provides a new derivative table to provide a mapping from a report to a scout patient ID to enable longitudinal queries. Users, including developers, have access to new views in trino for the curated, latest, and dx tables that will provide "resolved" EPIC MRNs and MPIs where we can infer them, even if they are not defined in the individual report.

### Technical
New view is added via trino-rw instance from #360 . Algorithm for constructing this mapping table is rather complex, but is documented in the internal patient IDs document.

## Type of change
- [X] New feature
- [ ] Improvement
- [ ] Bug fix
- [X] Documentation update
- [X] Test update

##### Authorization
N/A

##### Appsec
N/A

### Performance
A lot of work has gone into addressing the performance even with the full prod data set, and I've gotten it down to on the order of a few days for the entire dataset.

### Data
The algorithm is centered around the weirdness of the data.

### Backward compatibility
Not expected to be backwards compatible. Will likely require a full ingest, but this is expected to complete in a reasonable amount of time.

## Testing
I've added an automated test, including for the provenance in the history table. I've also gone through the full production data and done some spot checks, but this is fundamentally a complex feature that is hard to validate. I've done the best I can to not make any accuracy mistakes, but I can't promise one didn't sneak through.

## Checklist
- [X] I have added or updated user documentation, if appropriate


tests/ingest/src/test/java/edu/washu/tag/model/ReportPatientMappingHistoryEntry.java
```diff
@@ -0,0 +1,19 @@
+package edu.washu.tag.model;
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategies;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public class ReportPatientMappingHistoryEntry extends ReportPatientMappingEntry {
+
+    private String previousScoutId;
```
johnflavin: The python writes a `previous_scout_patient_id`. Is this supposed to map to the same id column? Because it looks to me like this would (with `SnakeCaseStrategy`) map to a `previous_scout_id`.

extractor/hl7-transformer/src/hl7scout/hl7extractor/mappingtableextractor.py
```diff
@@ -0,0 +1,607 @@
...
+            ).select(
+                F.expr("uuid()").alias("scout_patient_id"),
+                "primary_report_identifier",
+                "mpi",
+                "epic_mrn",
+                F.lit(True).alias("consistent"),
+            )
+        )
+
+        self.merge_to_dt(reports_with_no_existing_match_df)
```
johnflavin: (Sorry if this is wrong, it's a lot to try to follow.) Don't you need to `recache_existing_mapping` here so stage 4 can see any new ids that were just created?

extractor/hl7-transformer/src/hl7scout/hl7extractor/mappingtableextractor.py
```diff
@@ -0,0 +1,607 @@
...
+                on="primary_report_identifier",
+                how="left_anti",
+            ).select(
+                F.expr("uuid()").alias("scout_patient_id"),
+                "primary_report_identifier",
+                "mpi",
+                "epic_mrn",
+                F.lit(True).alias("consistent"),
+            )
+        )
```
johnflavin, line 396: This is a claude finding but it seems plausible to me.

- **Claim:** `reports_with_no_existing_match_df` calls `F.expr("uuid()").alias("scout_patient_id")` per row. Two stage-3 inputs that share the same single ID — e.g. (mpi=A, null) and (mpi=A, null') — pass through `process_single_id_reports` finding no existing match and each receive a distinct UUID, but they describe the same patient. The same is true for two `(null, X)` reports that came through stage 2 because `epic_mrn_count > 1`.
- **Evidence:** Stage 1 only dedupes rows with identical `(mpi, epic_mrn)` (Window partition on both). Stage 2 forwards anything with `mpi_count > 1 OR epic_mrn_count > 1` to stage 3, so multiple single-id rows sharing one ID can coexist in `reports_with_single_id_df`. The `uuid()` expression is evaluated independently per row.
- **Suggested fix:** Before assigning a UUID in stage 3, group by the present id (`mpi` or `epic_mrn`) and pick one UUID per group; or feed these rows into stage 4 instead.

tests/ingest/src/test/java/edu/washu/tag/model/ReportPatientMappingHistoryEntry.java
cmoore01: Absolutely; the history stuff was a very late addition here, so I hadn't actually run this section of the test. I was going to let CI run it, and it indeed failed on exactly this.

extractor/hl7-transformer/src/hl7scout/hl7extractor/dataextraction.py
```diff
@@ -63,3 +96,66 @@ def process_derivative_data(spark: SparkSession, report_table_name: str):
...
+            CREATE OR REPLACE VIEW {view_name} AS
+            SELECT
+                r.*,
+                m.scout_patient_id,
+                MAX(m.epic_mrn) OVER (PARTITION BY m.scout_patient_id) AS resolved_epic_mrn,
+                MAX(m.mpi)      OVER (PARTITION BY m.scout_patient_id) AS resolved_mpi
+            FROM {source_table} r
+            JOIN {mapping_table_name} m
+                ON r.primary_report_identifier = m.primary_report_identifier
+            WHERE m.consistent = true
```
kathrynalpert, line 137: As I'm looking at adding RBAC -- For Trino, we probably want this view to have `SECURITY INVOKER`. This warrants a bit more research/reading, so we can merge as-is and add this downstream. I do want to flag that if we don't add it now, we may need to recreate the view to get RBAC working. (We may end up with a full reingest anyway if we decide to lean on schemas or partitions more deeply, which doesn't appear to be required but could be worthwhile).

extractor/hl7-transformer/src/hl7scout/hl7extractor/mappingtableextractor.py
cmoore01: I don't think I'm really following here.
> Evidence: Stage 1 only dedupes rows with identical (mpi, epic_mrn) (Window partition on both).

Right, but that would include these:
> Two stage-3 inputs that share the same single ID — e.g. (mpi=A, null) and (mpi=A, null')

Those would have already been handled in stage 1. Here's an example:
```
from pyspark.sql import SparkSession
from pyspark.sql import functions as F
from pyspark.sql.types import StructType, StructField, StringType
from pyspark.sql.window import Window

spark = SparkSession.builder.appName("Stage1DedupDemo").getOrCreate()

schema = StructType([
    StructField("primary_report_identifier", StringType()),
    StructField("mpi", StringType()),
    StructField("epic_mrn", StringType()),
])

incoming = spark.createDataFrame(
    [
        ("r1", "A", None),
        ("r2", "A", None),
    ],
    schema,
)

incoming_report_dupe_id_window = Window.partitionBy("mpi", "epic_mrn").orderBy(
    F.monotonically_increasing_id()
)
remaining_reports_ranked = incoming.withColumn(
    "_rank", F.row_number().over(incoming_report_dupe_id_window)
)

unique_ids_incoming_reports = remaining_reports_ranked.filter(
    F.col("_rank") == 1
).drop("_rank")
duplicate_ids_incoming_reports = remaining_reports_ranked.filter(
    F.col("_rank") > 1
).drop("_rank")


print("\n=== Ranks within each (mpi, epic_mrn) partition ===")
remaining_reports_ranked.orderBy("mpi", "epic_mrn", "_rank").show(truncate=False)

print("=== Reports that PROCEED to Stage 2+ (rank == 1) ===")
unique_ids_incoming_reports.orderBy("primary_report_identifier").show(truncate=False)

print("=== Reports DEFERRED by Stage 1 (rank > 1) ===")
duplicate_ids_incoming_reports.orderBy("primary_report_identifier").show(truncate=False)
```

gives:

```
=== Ranks within each (mpi, epic_mrn) partition ===
+-------------------------+---+--------+-----+
|primary_report_identifier|mpi|epic_mrn|_rank|
+-------------------------+---+--------+-----+
|r1                       |A  |NULL    |1    |
|r2                       |A  |NULL    |2    |
+-------------------------+---+--------+-----+

=== Reports that PROCEED to Stage 2+ (rank == 1) ===
+-------------------------+---+--------+
|primary_report_identifier|mpi|epic_mrn|
+-------------------------+---+--------+
|r1                       |A  |NULL    |
+-------------------------+---+--------+

=== Reports DEFERRED by Stage 1 (rank > 1) ===
+-------------------------+---+--------+
|primary_report_identifier|mpi|epic_mrn|
+-------------------------+---+--------+
|r2                       |A  |NULL    |
+-------------------------+---+--------+
```

The partitionBy here does actually treat nulls as equal, which is what we want for this particular case.

docs/source/dataschema.md
```diff
@@ -278,3 +279,69 @@ ID for the report and an index starting from 0 per report, separated by an under
 the latest table with new diagnostic information may overwrite and/or delete rows in this table when looking by id. Additionally,
 each row contains a diagnosis with the columns `diagnosis_code`, `diagnosis_code_text`, and `diagnosis_code_coding_system`, matching
 the earlier structs.
+
+(mapping_ref)=
+### reports_report_patient_mapping
```
kathrynalpert, line 284: Do you want to include anything on `reports_report_patient_mapping_history` in the docs?

extractor/hl7-transformer/src/hl7scout/hl7extractor/mappingtableextractor.py
johnflavin: You’re correct, this was not a valid finding. Thanks for helping me understand what's going on there.

extractor/hl7-transformer/src/hl7scout/hl7extractor/mappingtableextractor.py
cmoore01: Actually...I think you are correct. I'm confused why the new test wouldn't have caught this though. Let me dig in and see what's going on.

docs/source/dataschema.md
cmoore01: Hmm, I'm not sure. There is something in the sense that it's described in the original patient IDs proposal that I added to the docs. They're not something I would expect an end user to ever query though.

kathrynalpert (approved):

PR #371, opened 2026-05-14
Declarative configuration of Chat via Helm-hooked bootstrap Job

### LINKED ISSUE #358: Knowledge files not injected into LLM context after Open WebUI v0.7.0 upgrade
When we bumped the Open WebUI Helm chart from 13.3.1 → 14.2.0 (PR #354), knowledge file injection broke. In v0.7.0+, OWUI switched to a native function calling path that exposes `list_knowledge` / `query_knowledge_files` as built-in tools instead of auto-injecting knowledge content into the system prompt. In practice the LLM never calls those tools because nothing in the prompt tells it to.

On preprod we dropped the knowledge files and folded the schema/knowledge content directly into each model's system prompt.

kathrynalpert:
### Product

The Chat UI's model picker now offers three Scout Explorer entries. I did my best to [have Claude] describe them, but I'm very open to changes.
- **Scout Explorer (Gemma 4 31B)** — recommended default. Fast and reliable for cohort queries, filters, and free-text searches.
- **Scout Explorer (Qwen 3.6 35B)** — reasoning-focused alternative. Thinks step-by-step on every question.
- **Scout Explorer (GPT-OSS 120B)** — largest alternative. Broadest medical knowledge; slower first response.

The radiology-tuned suggestion-prompt cards on the new-chat screen (Pulmonary nodule cohort, Turnaround time, Exam volume trends, Report addendum rate, Pediatric CT) are now consistent across deployments. (Also very open to changes.)

### Technical

Replaces the plethora of manual steps to set up Chat chain with an in-cluster Job deployed via a new Helm chart at `helm/open-webui-bootstrap/`. The chart's Job runs as a `helm.sh/hook: post-install,post-upgrade` so it executes after OWUI is installed/upgraded, and re-runs on every Helm upgrade thanks to a `checksum/config` annotation over the rendered ConfigMap.

**Job phases** (`helm/open-webui-bootstrap/files/bootstrap.py`):
1. **Password migration** — rewrites the bootstrap user's bcrypt hash to match the current `open_webui_bootstrap_password`. Idempotent; no-op on a clean DB. The Job runs the same container image as OWUI (image tag discovered at deploy time from the running Pod), so its `import open_webui.internal.db` / `open_webui.models.auths` calls execute against the same SQLAlchemy models and bcrypt version the OWUI Pod uses.                    
2. **Mint admin JWT** — `/signin` first; falls back to `/signup` on a truly empty DB (gated by `ENABLE_INITIAL_ADMIN_SIGNUP=true`). Retries to absorb pod-not-quite-ready races.
3. **PersistentConfig** — tool servers, `DEFAULT_MODELS`, `TASK_MODEL` updated via REST. OWUI stores these in the `config` Postgres table as PersistentConfig, where env-var changes after first launch are silent no-ops, so we re-POST on every deploy to keep them inventory-driven.
4. **Filter functions** — idempotent GET-create-or-update one entry per `open_webui_filter_functions[]` (`Link Sanitizer`, `Context Summarization`, diagnostic `Tool Result Attribute Migrator`); sets valves and toggles `is_global`/`is_active` to the desired state.
5. **Custom models** — one ModelForm payload per `scout_models[].ui` entry plus hide-base overrides for the raw Ollama tags.

The chart owns all payload rendering: `configmap.yaml` transforms raw inventory shapes (`filterFunctions`, `scoutModels`, `scoutDefaults`) into the JSON keys `bootstrap.py` reads from `/app/config/`. Filter source code, the system prompt, and the profile image live in `helm/open-webui-bootstrap/files/payloads/` and are globbed into separate CM keys.

**Other architectural notes/shifts:**
- **Image lookup**: `configure_admin.yaml` discovers the running OWUI Pod's image via `kubernetes.core.k8s_info` and passes it to the bootstrap chart's `image.ref` (single field, handles `:tag` and `@sha256:digest`). Renovate bumps the OWUI chart → chart appVersion drives both the OWUI Pod and the bootstrap Job automatically.
- **Multi-model context handling**: `scout_models` is now a list. Each entry creates a Modelfile-derived variant on top of `base` with extended `num_ctx`/`num_keep`/`num_predict` baked in (Ollama's runtime default is 4096 regardless of native max). `preload: true` warms the model into VRAM via the values.yaml lifecycle hook; `preload: false` cold-loads on demand to manage VRAM headroom.
- **Image cleanup**: `gpt-oss-scout-query-prompt.md` → `scout-system-prompt.md` (now used by all three variants); legacy `rag-prompt.md` removed (Scout Explorer uses `function_calling: native`, which bypasses OWUI's RAG auto-injection — schema/charting docs are inlined into the system prompt instead).
- **Version bumps** OWUI chart: `~14.2.0` (app 0.9.2) → `~14.5.0` (app 0.9.5), Ollama 0.22.1 → 0.23.4

## Type of change
- [ ] New feature
- [x] Improvement
- [ ] Bug fix
- [x] Refactor (code improvement with no functional changes)
- [x] Documentation update
- [ ] Test update

##### Authorization

A persistent admin user `scout-deploy@scout-deploy.local` is created at the OWUI application layer. The bootstrap email is `.local` (RFC 6762 reserved, never delivered to) so it can't collide with a real user. Note: I initially used OWUI's `WEBUI_AUTH_TRUSTED_*_HEADER` SSO feature for the bootstrap so as not to require an admin account. This passed `X-Scout-Bootstrap-*` headers in-cluster, with a Traefik middleware stripping them from external traffic. I had to abandon this approach because I was intermittently seeing error messages in the frontend about missing headers. If we are uncomfortable with the admin user, we could dive into this more deeply.

No changes to end-user Keycloak roles (`open-webui-user` / `open-webui-admin`) or to how OAuth2 Proxy enforces the `scout-user` group gate. 

The capability flags on Scout Explorer models disable `web_search`, `image_generation`, `code_interpreter`, `terminal`, and `vision` (no internet, no PHI exfil via LLM-generated URLs, no arbitrary code execution).

##### Appsec

- **Secret hygiene** — `DATABASE_URL` is now sourced from the `open-webui-secrets` Secret via `extraEnvFrom` (OWUI workload) and `secretKeyRef` (bootstrap Job). Setting the chart's `databaseUrl: ''` skips its conditional `env: - name: DATABASE_URL value: <literal>` block. Postgres password no longer renders into the Helm release Secret or `kubectl describe pod` output for either chart. `WEBUI_SECRET_KEY` and `OAUTH_CLIENT_SECRET` are read from the same Secret (unchanged from prior).

### Performance

Mostly N/A for these changes. `gemma4-31b-long` and `qwen3.6-35b-long` can run concurrently on the 80GB A100.

### Data

OWUI's `config` table (Postgres) is the source-of-truth for runtime settings; this PR explicitly designates four fields as **declarative** (re-pushed on every deploy):
- `tool_server_connections` (overwritten — handler unconditionally assigns)
- `DEFAULT_MODELS` (GET-merge-POST to preserve siblings)
- `TASK_MODEL` (GET-merge-POST)
- Filter function rows + their valves + global/active toggles

**Anything an operator changes in OWUI Admin UI for these fields gets overwritten on the next deploy.** Documented in role README's Bootstrap & Migration section; escape hatch is `open_webui_admin_setup_enabled: false` to skip the Job entirely.

`RAG_TEMPLATE` is intentionally **not** pushed — Scout Explorer runs with `function_calling: native`, which bypasses OWUI's RAG auto-injection path. The old knowledge files are inlined into the system prompt instead (closing issue #358).

Seed-only knobs (`WEBUI_URL`, `ENABLE_EVALUATION_ARENA_MODELS`, `ENABLE_SIGNUP`, `ENABLE_LOGIN_FORM`, `ENABLE_COMMUNITY_SHARING`) stay env-var-only: set once at first launch, expected stable thereafter (aka attempts to update these fields after first install will not be detected). README documents the SQL-surgery workaround if one of these ever needs flipping post-launch.

### Backward compatibility

Existing OWUI deployments with other users but no `scout-deploy@scout-deploy.local`** will require manual action. OWUI's `has_users` check blocks `/signup` once any user exists; the Job exits non-zero with a diagnostic pointing to the README's one-time manual step (create the bootstrap user via Admin UI with any password; migration rewrites it).

Single-model defaults migrate to the three-model list automatically via the new `scout_models` shape. Inventory overrides for the old `scout_base_model` / `scout_model_name` scalars no longer apply — operators with such overrides need to migrate to the `scout_models:` list shape. Defaults / README / `inventory.example.yaml` updated.

**Note**: the diagnostic `tool_result_attr_filter` works around an OWUI 0.9.5 rendering regression — when bumping past `~14.5.0`, verify the tool-call Output panel renders correctly without the filter and remove the filter source + inventory entry if upstream has fixed it.

## Testing

**Automated:**
- New unit tests for the `tool_result_attr_filter` Function (`helm/open-webui-bootstrap/tests/test_tool_result_attr_filter.py`): idempotency, body→attribute rewrite, non-tool-call blocks untouched.
- Existing Link Sanitizer and Context Summarization filter tests moved to `helm/open-webui-bootstrap/tests/` alongside the filter sources.
- `helm template` renders cleanly with default values; verified `DATABASE_URL` is sourced only via `envFrom` (no literal value emission) and the bootstrap Job's `image: <repo>:<tag>` references the OWUI ref passed from Ansible.

**Manual:**
- [x] Fresh deploy: `make install-chat` from clean
- [x] Re-deploy on a fresh cluster after editing inventory (model name + description): re-run picks up changes
- [x] Sanity check Scout Explorer behavior: select each variant, send "How many CT reports are there?", confirm the Trino MCP tool is invoked natively and results render in the Output panel (validates the diagnostic filter for 0.9.5).
- [x] Air-gapped path: confirm model pulling still works on staging cluster + NFS mount; lifecycle postStart preload on production pod start.

## Checklist


helm/open-webui-bootstrap/files/bootstrap.py
```diff
@@ -0,0 +1,322 @@
...
+    for fn in load_config("filters.json") or []:
+        fn_id = fn["id"]
+        payload = {
+            "id": fn_id,
+            "name": fn["name"],
+            "content": load_text(fn["content_file"]),
+            "meta": {"description": fn.get("description", "")},
+        }
+
+        exists_code, _ = http("GET", f"/api/v1/functions/id/{fn_id}", token=token)
```
johnflavin: ~~I don't know where `filters.json` comes from or what its contents are, but~~ does this `fn_id` need to be url-encoded? (Also applies to a few lines below.)

I traced through the logic and I think it would be good to url-encode the `fn_id`.
- filters.json comes from a configmap
- The contents of that configmap are built from `.Values.filterFunctions`.
- `filterFunctions` is set from `open_webui_filter_functions`
- `open_webui_filter_functions` is set in the openwebui role defaults and is overridable in the inventory

It is in principle user-settable. Currently everything in `open_webui_filter_functions` has an `id` that is ok to use in a url, but it would be a nice safety measure to url-encode to protect against future user error.

johnflavin (approved):

ansible/roles/open-webui/defaults/main.yaml
```diff
@@ -100,3 +142,260 @@ open_webui_resources_default:
...
+  - title: ['Pulmonary nodule cohort', 'Chest CT findings']
+    content: 'Find chest CTs showing a pulmonary nodule.'
+  - title: ['Turnaround time', 'Order to final report']
+    content: "What's the typical turnaround time from order to final report, by modality?"
+  - title: ['Exam volume trends', 'Monthly, by modality']
+    content: 'How has exam volume changed over time, by modality?'
+  - title: ['Report addendum rate', 'Quality signal']
+    content: 'How often are reports amended with an addendum, by modality?'
+  - title: ['Pediatric CT', 'Patients under 18']
+    content: 'How many pediatric CT exams have we done, broken down by year?'
```
andylassiter, line 356: I'm trying to think of questions Dan may be interested in. Maybe something neuroimaging, stroke, or Alzheimer's related?

This was an interesting one from Claude:
```
  - title: "Ischemic vs hemorrhagic stroke"
    content: "How many reports mention ischemic stroke versus hemorrhagic stroke in the impression?"
```
> The ischemic vs hemorrhagic split is a classic epidemiology query — roughly 85/15 nationally, interesting to see if your population matches. 

A couple others:
```
- title: "Long-term neuroimaging patients"
   content: "How many patients have brain MRI reports spanning more than five years?"
- title: "Dementia progression imaging"
   content: "How many patients have more than one brain MRI report mentioning dementia?"
```

helm/open-webui-bootstrap/files/payloads/scout-system-prompt.md
```diff
@@ -0,0 +1,301 @@
...
+### Frequently-queried columns
+
+| Column | Type | Notes |
+|---|---|---|
+| `year` | int | **Partition column** — always include in WHERE for performance. Derived from `message_dt`. |
+| `message_dt` | timestamp | When the HL7 message was created. |
+| `requested_dt` | timestamp | When the order was placed. Preferred TAT start (more reliably populated than `observation_dt`). |
+| `observation_dt` | timestamp | Fallback TAT start. |
+| `results_report_status_change_dt` | timestamp | Report finalized. TAT end. |
+| `modality` | string | Derived (CT, MR, US, MG, NM, PT, etc.). |
```
andylassiter: I've been noticing gemma4 keeps trying to use `MRI` for modality instead of `MR`. Maybe because a lot of our service names start with `MRI ...`? Addressing this is probably out of scope of this PR but wanted to call it out and see if you've noticed any issues.

andylassiter (approved):

helm/open-webui-bootstrap/files/bootstrap.py
kathrynalpert: Fixed!

ansible/roles/open-webui/defaults/main.yaml
kathrynalpert: Added

helm/open-webui-bootstrap/files/payloads/scout-system-prompt.md
kathrynalpert: Added some prompt direction for this, though I haven't seen it myself.

PR #373, opened 2026-05-14
Unvendor cass operator and bump cass-operator + cassandra versions

johnflavin:
### Product
N/A

### Technical
In #330 I vendored in the cass-operator chart because k8ssandra's web hosting was having issues. Those have been resolved, so I have un-vendored the chart, and bumped the version while I'm at it.

## Type of change
- [ ] New feature
- [X] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
N/A

##### Appsec
N/A

### Performance
N/A

### Data
N/A

### Backward compatibility
This should apply on top of a Scout with either the older k8ssandra chart or the vendored chart.

## Testing
Deployed to dev03 as an upgrade of the vendored chart.

## Checklist


ansible/group_vars/all/versions.yaml
```diff
@@ -24,11 +24,8 @@ elasticsearch_version: 9.2.3
 # Databases & Data Storage - Cassandra
 # renovate: datasource=helm registryUrl=https://charts.jetstack.io depName=cert-manager
 cert_manager_version: ~1.19.0
-# Chart vendored at vendored/helm/cass-operator-0.60.0.tgz due to broken upstream Helm repo
-# (https://github.com/k8ssandra/k8ssandra-operator/issues/1710).
-# Update vendored .tgz when bumping this version.
 # renovate: datasource=helm registryUrl=https://helm.k8ssandra.io/stable depName=cass-operator
-cass_operator_version: ~0.60.0
+cass_operator_version: ~0.64.1
```
kathrynalpert: You probably checked this, but should we bump `cassandra_server_version`, below?

kathrynalpert (approved):

ansible/group_vars/all/versions.yaml
johnflavin: I actually hadn't thought of bumping that because I figured renovate would have bumped it for us. But it seems like there is an update that could be applied, but renovate didn't do it. I'll bump it manually and try to figure out renovate separately.

johnflavin: Ah, I see. I forgot that renovate will only auto-create PRs for CVEs, but queues everything else on the dashboard. Still haven't worked that into my regular process...

johnflavin: I tried bumping the cassandra version, and I'm glad I did because it exposed an incompatibility with the `cass-operator` version that had been there but I hadn't noticed. It failed to come up because the `server-system-logger` sidecar didn't like our CPU.
```
Fatal glibc error: CPU does not support x86-64-v3
```
I bumped the `cass-operator` version back down somewhat and everything came up.

kathrynalpert: We're having the same CPU issue with Trino FWIW.

PR #391, opened 2026-05-21
[Estimated] Patient age derivation for all versions of HL7 and defensive age calculation

cmoore01:
### Product
Users should see `patient_age` populated now for all versions of HL7. This also moves this field from the base `reports` table to `reports_curated`.

### Technical
We use a particular date field as our best proxy for scan date depending on the version of the report and diff this against the patient's DOB to derive `patient_age`. There are also some new defensive checks:
1. If the DOB is before 1905, we're pretty confident that's not a correct DOB so we null it out (and therefore age will not be populated).
2. If `patient_age > 109`, that's probably not correct, so we null out both DOB and `patient_age`.

## Type of change
- [ ] New feature
- [X] Improvement
- [ ] Bug fix
- [X] Documentation update
- [X] Test update

##### Authorization
N/A

##### Appsec
N/A

### Performance
Should not be particularly expensive.

### Data
Added tests around different possibilities.

### Backward compatibility
This changes the schema to move `patient_age` from the default table to `reports_curated`. It's not expected to be backwards compatible.

## Testing
New automated test verifying the behavior.

## Note for reviewers
1. The two cutoffs on DOB and age are somewhat arbitrary. Should they be tweaked?

2. There are some uses of `patient_age` in such as in `analytics/notebooks/rads/rads_builder.py` or `helm/open-webui-bootstrap/files/payloads/scout-system-prompt.md` . I'm not planning on updating them because we're already in the process of thinking about switching things to use the new views. We should handle it then.

## Checklist
- [X] I have added or updated user documentation, if appropriate


docs/source/dataschema.md
```diff
@@ -235,7 +234,8 @@ in the base report table. A row in the curated table looks exactly the same as a
...
-| `patient_ids`                                             | See note below about `patient_mpi`                                                                                                                                                                                                                                                      |
+| `patient_mpi`                                             | See note below about `patient_mpi`                                                                                                                                                                                                                                                      |
+| `birth_date` & others                                     | Patient age at time of report. See note below about `patient_age`.                                                                                                                                                                                                                      |
```
kathrynalpert: `& others` is confusing to me. Other dates?

extractor/hl7-transformer/src/hl7scout/hl7extractor/curatedtable.py
```diff
@@ -124,12 +126,29 @@ def extract_labeled_patient_id(id_column: str, df: DataFrame) -> Column:
                     ),
                 )
                 .otherwise(F.col("mpi")),
+                "scan_date_proxy": F.when(
+                    F.col("version_id") == "2.7", F.col("requested_dt")
```
kathrynalpert: Out of curiosity, why use version here vs a coalesce operation? If it's 100% consistent, I agree that the version marker is tidier, but is it more defensive to just check for both?

kathrynalpert (approved):

extractor/hl7-transformer/src/hl7scout/hl7extractor/curatedtable.py
cmoore01: Good question. That's probably better here.

docs/source/dataschema.md
cmoore01: I'll list the two explicitly instead

docs/source/dataschema.md
```diff
@@ -252,6 +252,11 @@ for a 2.7 report. This gives us moderate confidence in an identifier that remain
 be present (such as for 2.7 reports without an EMPI id). For an easier to use alternative, see
 {ref}`Longitudinal Patient ID Resolution <longitudinal_pat_id_ref>`.
 
+Scout derives a `patient_age` as the difference between `birth_date` and a chosen scan date proxy based on the HL7 version. For 2.3 or 2.4 reports,
```
johnflavin: Nit: Could use a brief mention of why we need a proxy at all

extractor/hl7-transformer/src/hl7scout/hl7extractor/curatedtable.py
```diff
@@ -79,6 +77,8 @@ def extract_labeled_patient_id(id_column: str, df: DataFrame) -> Column:
             id_column, df, F.concat_ws("_", F.lit(id_column), F.col(id_column))
         )
 
+    unreasonable_age_for_scans = F.col("patient_age") > 109
```
johnflavin: Nit: A magic number like this should be defined in a constant. I'm ok with it being arbitrary, but I at least want it to be easy to identify.

extractor/hl7-transformer/src/hl7scout/hl7extractor/curatedtable.py
```diff
@@ -124,12 +124,29 @@ def extract_labeled_patient_id(id_column: str, df: DataFrame) -> Column:
                     ),
                 )
                 .otherwise(F.col("mpi")),
+                "scan_date_proxy": F.coalesce(
+                    F.col("observation_dt"), F.col("requested_dt")
+                ),
+                "birth_date": F.when(
+                    F.col("birth_date") > "1905-01-01", F.col("birth_date")
+                ).otherwise(F.lit(None)),
```
johnflavin, line 136: Just want to confirm that python's `None` is the right value to use here. I _think_ we have an operation at the end that converts all `None`s to a more db-friendly value, but I don't remember for sure.

extractor/hl7-transformer/src/hl7scout/hl7extractor/curatedtable.py
```diff
@@ -124,12 +124,29 @@ def extract_labeled_patient_id(id_column: str, df: DataFrame) -> Column:
...
             }
         )
         .withColumns(
             {
                 "accession_number": F.col("filler_order_number"),
                 "primary_study_identifier": F.col("filler_order_number"),
+                "patient_age": F.expr(
+                    "CAST(datediff(YEAR, birth_date, scan_date_proxy) AS INT)"
```
johnflavin, line 144: I was initially going to point out that this `datediff` operation seems to only operate on the year, which I'm guessing strips off the dates entirely and elides any complexity about if a birthday has or has not yet occurred in a given calendar year. But I wan't sure if that was correct so I wanted to look up the `datediff` docs to confirm. But I don't see one in the pyspark docs that can take a `YEAR` value in the first parameter (presumably an interval). I only see https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.datediff.html, which admittedly is in the python API which isn't always the same as the `expr` string API. 

I don't doubt that this function does exist because you've been using it. But I can't find its exact details.

johnflavin (approved):

extractor/hl7-transformer/src/hl7scout/hl7extractor/curatedtable.py
cmoore01: Isn't `unreasonable_age_for_scans` basically already that constant that gets reused? I could do something like:

```
unreasonable_age = 109
unreasonable_age_for_scans = F.col("patient_age") > unreasonable_age
```

but is that really better?

extractor/hl7-transformer/src/hl7scout/hl7extractor/curatedtable.py
cmoore01: Yep, `F.lit(None)` gets turned into `NULL` in the delta lake.

extractor/hl7-transformer/src/hl7scout/hl7extractor/curatedtable.py
johnflavin: What I was thinking of would be a global constant outside the function with an all-caps name like
```
UNREASONABLE_AGE_THRESHOLD = 109
```
I do think it's better, yes. It communicates intent and makes the value easier to find. But it's not necessary, hence the "Nit" designation.

extractor/hl7-transformer/src/hl7scout/hl7extractor/curatedtable.py
cmoore01: The databricks doc is probably the clearer one here: https://docs.databricks.com/aws/en/sql/language-manual/functions/datediff3 . The unit is used for the format of the output, so I'm pretty sure it's not throwing out month/date/time information. It certainly would be nice if we had a spark document to point at instead though.

PR #395, opened 2026-05-22
Fix Temporal workflow failure alert/panel NoData when no failures

johnflavin:
### Product
The Temporal workflow failure alert and dashboard panels no longer show "No Data" when there have been no workflow failures; they now correctly show zero.

### Technical
Updated the Prometheus expressions in the Temporal workflow failure alert and the Temporal dashboard panels (failed/timeout/terminate/cancel) to fall back to a zero vector when the `workflow_failed`/`workflow_timeout`/`workflow_terminate`/`workflow_cancel` series have no samples. Per-namespace panels use `or 0 * sum by (namespace) (rate(workflow_success...))` to preserve the namespace label; the aggregate dashboard queries use `or vector(0)`; the alert query uses the namespace-preserving form.

## Type of change
- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
N/A

##### Appsec
N/A

### Performance
Negligible. Adds a second rate query on `workflow_success` only as a label-preserving zero fallback.

### Data
N/A

### Backward compatibility
No data model or API changes. Existing alert/dashboard JSON is overwritten on next Grafana role deploy.

## Testing
Verified the updated PromQL expressions return zero (rather than empty) when no failure series are present, while still preserving the `namespace` label on per-namespace panels.

## Note for reviewers
N/A

## Checklist

### DIFF
--- a/ansible/roles/grafana/templates/alerts/temporal-workflow-failure-alert.json.j2
+++ b/ansible/roles/grafana/templates/alerts/temporal-workflow-failure-alert.json.j2
@@ -21,7 +21,7 @@
               "datasourceUid": "prometheus_datasource_01",
               "model": {
                 "editorMode": "code",
-                "expr": "sum by (namespace) (rate(workflow_failed{operation=\"CompletionStats\", job=\"kubernetes-pods\"}[1h]))",
+                "expr": "sum by (namespace) (rate(workflow_failed{operation=\"CompletionStats\", job=\"kubernetes-pods\"}[1h])) or 0 * sum by (namespace) (rate(workflow_success{operation=\"CompletionStats\", job=\"kubernetes-pods\"}[1h]))",
                 "instant": true,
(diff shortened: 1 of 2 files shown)


kathrynalpert (approved):
Probably worth testing that a true failure still triggers an alert (especially the first one for the site, which I think should be handled by the part of the alert you didn't modify, but worth confirming)

johnflavin:
> Probably worth testing that a true failure still triggers an alert (especially the first one for the site, which I think should be handled by the part of the alert you didn't modify, but worth confirming)

I triggered a workflow failure. I can see it in the dashboard which runs the same query as the alert. So even though the alert isn't firing _yet_—it just checked a few minutes before I ran my workflow and won't check again for an hour—I'm confident it will fire.

PR #402, opened 2026-05-26
Trino AuthZ: ADRs and OPA policies

kathrynalpert:
### Product
N/A - First slice of Trino per-user authorization (policy and design records only).

### Technical
- `policy/trino/main.rego` + `main_test.rego` — the OPA policy (allow rules, attribute-driven row filters, type-aware column masks, the enabled + `scout-user`/`scout-admin` approval gate, hidden-table baseline, DEFINER-view carve-outs) and `opa test` unit tests.
- ADRs 0020–0023 — engine choice, attribute distribution via MinIO bundles, authentication/identity propagation, view security model.

## Type of change
- [x] New feature
- [ ] Improvement
- [ ] Bug fix
- [x] Documentation update
- [x] Test update

##### Authorization
Defines the authorization model but does not enforce it yet. Default-deny; fail-safe (missing/empty attributes deny rows and mask PHI rather than exposing). Details and rationale in ADR 0020.

##### Appsec
Declarative Rego, no interpolation of user input into queries. Reviewed for fail-open paths; the mask and row-filter rules fail safe, with a regression test per case.

### Performance
N/A

### Data
N/A

### Backward compatibility
N/A

## Testing
All these PRs have been tested together, plus there are a few CI pieces. Once they've all been merged onto the `trino-rbac` branch, I'll test as a bolus before merging to `main`.

I will add the OPA tests to CI, but not until the 3rd PR when I add an OPA role + deployment within Scout.

## Note for reviewers
OPA policies we might not want to carry through to prod (like mask PHI) should be configurable via inventory.

## Checklist


docs/internal/adr/0021-opa-user-attribute-distribution.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Alternatives Considered
+
+### OPA pulls attributes from Keycloak via `http.send` with a TTL cache
+
+Rego policy issues an `http.send` to Keycloak's admin API for each unique `(user, attribute-set)` lookup, with `force_cache_duration_seconds` (e.g. 60s) as the staleness bound. Listener does nothing (or pushes per-user invalidation timestamps the policy threads into the URL to bust the cache key).
+
+**Rejected**: keeps decisions coupled to Keycloak — every cache miss is a Keycloak admin-API round-trip, the policy holds a client_credentials token, and a Keycloak outage degrades Trino authorization. Multi-replica fan-out is unsolved: invalidation pushed to one OPA pod doesn't reach the others until their independent TTLs elapse. Cache-busting timestamps (URL-as-cache-key) are non-obvious for future Rego readers — the mechanism doesn't model the problem, it works around OPA's lack of a per-key cache eviction API. Misses goal 4 directly and creates ongoing conceptual debt.
```
johnflavin: I'm not convinced that this alternative isn't sufficient. The arguments for rejecting it don't make sense to me. 

- "keeps decisions coupled to Keycloak" - They will be no matter what. That's ultimately the source of truth for the user attributes.
- "every cache miss is a Keycloak admin-API round-trip" - Maybe, sure. But is that worse than a Minio round trip?
- "a Keycloak outage degrades Trino authorization" - A keycloak outage would degrade all of Scout and prevent login, so I don't think this is an unacceptable consequence.
- "invalidation pushed to one OPA pod doesn't reach the others until their independent TTLs elapse" - Invalidations aren't pushed in this scenario, they're pulled. And the Minio file solution _also_ doesn't reach every pod until their independent TTLs elapse and break their caches, so that's no different.
- "Cache-busting timestamps (URL-as-cache-key) are non-obvious for future Rego readers" - I do not understand what this is talking about. Maybe this is a real reason to not use the "pull from KC" method but I think it relies too much on the details of the caching implementation for me to get it.

docs/internal/adr/0022-trino-auth-and-impersonation.md
```diff
@@ -0,0 +1,221 @@
...
+[ADR 0020](0020-trino-authz-architecture.md) picks OPA as Trino's authorization engine and locks the per-user attribute model. For OPA to evaluate authorization against the actual end user, Trino has to:
+
+1. Authenticate the connection cryptographically (no more hardcoded shared user).
+2. Carry the real end-user identity through to the access-control plugin — not the connection's service-account identity.
+
+Trino was previously unauthenticated; every client connected as the hardcoded user `trino`. This ADR specifies how each Trino-using client authenticates and how end-user identity reaches `input.context.identity.user` so OPA can index into `data.users`.
+
+The shape of the problem differs per client:
+
+- **JupyterHub** has the user's OIDC token available server-side (via `auth_state`) and the kernel runs arbitrary user code — natural fit for per-call user JWT.
```
johnflavin, line 18: Nit: This is technically true but not actually usable yet. JH has the token but it isn't accessible to the notebook user through the JH API until we make a config change. I have that on my XNAT POC branch, so we'll get it in soon.

docs/internal/adr/0022-trino-auth-and-impersonation.md
johnflavin, line 1: My XNAT auth POC also takes KC JWTs from other services, but instead of using them directly it does a KC token exchange. Would there be any benefit to doing a token exchange here? (Or, conversely, is there any actual benefit to doing it in the XNAT auth POC, or should I just take the Jupyter token as-is?)

docs/internal/adr/0022-trino-auth-and-impersonation.md
```diff
@@ -0,0 +1,221 @@
...
+       ↓ scout_voila overrides Voila's GET handler → stashes token in a contextvar
+  Voila spawns kernel → ScoutMappingKernelManager (class config) reads contextvar → kernel env
+       ↓
+  playbook code: scout_trino.connect()
+                   - decodes preferred_username from env token
+                   - mints voila_svc JWT (client_credentials; token: aud=trino; re-minted before exp)
+                   - HTTP: Bearer <voila_svc JWT>, X-Trino-User: <user>
+       ↓
+  Trino: principal=voila_svc, user=alice → OPA evaluates against data.users["alice"]
+                                       impersonation: voila_svc ∈ trino_service_principals
```
johnflavin, line 97: Voila's token-passing makes me concerned about token staleness. Let me think this through...
- On a user request, a fresh token comes in from oauth2-proxy via `X-Auth-Request-Access-Token: <user OIDC token>`
- `scout_voila overrides Voila's GET handler → stashes token in a contextvar` - I don't know the scope of a contextvar, so I don't know how long this would live. But let's keep reading...
- `Voila spawns kernel → ScoutMappingKernelManager (class config) reads contextvar → kernel env` - This makes it seem like we get a token once when the user first loads the voila playbook and we keep it forever, as long as the kernel lives. The kernel is long-lived, and its envs are too. This has a smell of staleness to me.
- At trino request time `mints voila_svc JWT (client_credentials; token: aud=trino; re-minted before exp)` - This is the crux of the matter. This request could be but isn't necessarily on playbook load. Thinking of the cohorting playbook, the user has some amount of time they can fiddle with the query before issuing it. It does say "re-minted before exp" but does that matter? This has to somehow use the source auth token, right? Or maybe it just pulls out all the user info from the source token and uses it to construct its own token? If this new token is tied to the exp of the user auth token, then we have a staleness issue. If all we need from the user auth token is some user info and this token's exp is independent, then we're safe.

docs/internal/adr/0022-trino-auth-and-impersonation.md
```diff
@@ -0,0 +1,221 @@
...
+
+Open WebUI MCP (HTTP Basic + impersonation)
```
johnflavin, line 99: The trino mcp tool is likely being replaced with a _bespoke_ (😀) OWUI tool. Seems like we should analyze what we are able to do there in addition to / instead of the trino mcp.

docs/internal/adr/0023-trino-view-security-model.md
johnflavin, line 1: I don't know enough about this to meaningfully comment. I can see what it's saying, but it could be 100% sound or 100% bullshit and I would not know the difference. As long as the POC works I guess it's correct?

Might need to tap @cmoore01 in to consider this one.

kathrynalpert: What inaccuracy are you worried about here? 

I've verified that this works in practice with `SECURITY DEFINER`, so this ADR is more about recording _why_ it works in case someone is confused in the future.

docs/internal/adr/0022-trino-auth-and-impersonation.md
kathrynalpert: Pass-through depends on adding the `trino-audience` client scope to the `jupyterhub` Keycloak client, so Jupyter access tokens carry `aud=trino` directly, which means we don't have to exchange. But... should we?

docs/internal/adr/0021-opa-user-attribute-distribution.md
kathrynalpert: I will rewrite this to make the case more clear. I'm confident this is not the way to go, most critically because using the OPA bundle leans on functionality already in OPA vs rolling something custom. And while, yes, a KC outage is problematic generally for Scout, it wouldn't impact logged-in users unless we used this approach.

docs/internal/adr/0022-trino-auth-and-impersonation.md
kathrynalpert: Right, I have it in a subsequent PR, too.

docs/internal/adr/0022-trino-auth-and-impersonation.md
kathrynalpert: Your logic is right, but the piece you're missing is that Voila is using impersonation, not the token pass-through we use for Jupyter. As such, all we need from the user's token is the user's identity, which we pass via the Voila service token (which we keep fresh) to Trino. It doesn't matter if the user's token expires.

The reason we can't perform token pass-through (or exchange) in Voila is that the kernel has no way to refresh the user's token. JupyterHub is an OIDC client, with direct connectivity to Keycloak, so it can exchange the user's `refresh_token` server-side for fresh access tokens, which it shares to the kernel via its API. Voila is just a notebook renderer behind `oauth2-proxy`, so it doesn't talk to Keycloak nor have a `refresh_token`.

I'll try to make this more clear in the doc.

docs/internal/adr/0022-trino-auth-and-impersonation.md
kathrynalpert: Yes, the intention would be to write a tool than can send a JWT to Trino like our other services, and remove the PASSWORD auth from Trino. I'm hesitant to get into too much detail here about a service we have yet to write, but I'll at least note the plan.

docs/internal/adr/0022-trino-auth-and-impersonation.md
kathrynalpert: I had Claude research this:

> [RFC 9700 §4.10.2](https://www.rfc-editor.org/rfc/rfc9700.html#section-4.10.2) names scope-value encoding (what our client scope does) as one of two endorsed mechanisms for attaching audience at issuance — the other being [RFC 8707](https://www.rfc-editor.org/rfc/rfc8707.html) resource indicators. §2.3 permits multi-audience as a fallback when single-RS binding "isn't feasible," with the trust-required caveat in RFC 8707 §3. In-cluster, same-realm, same-operator clears that bar.
> 
> We could drop the `trino-audience` scope from the JH client and exchange for a Trino-audience token on demand, which makes every token in the system single-audience. This means smaller per-token blast radius, independent RS-side lifetime, KC audit-log entry per exchange. This affords some improvement in security posture, but it's narrow for our threat model. The realistic leak vectors — logs, exporters, kernel compromise — bypass the wire entirely and surface upstream tokens regardless of what we send downstream. TLS already covers the JH→Trino wire, OPA decision logs already give per-query forensics, and the other audiences in our multi-aud token are equally-trusted in-cluster services with their own per-user authorization — not categorically worse than `trino`.
> 
> RFC 9700 §4.10 lists sender-constrained tokens (mTLS / DPoP, §4.10.1) and audience restriction (§4.10.2) as defenses *in tandem*. If stolen-token resistance is the real concern, sender constraint is the stronger lever — exchange-vs-pass-through is a smaller win than mTLS-vs-bearer. 

I'm inclined to stick with pass-through since this only impacts one service, exchange adds some complexity, and I don't think the security win is all that meaningful. But I'm open to swapping, esp if the XNAT work offers some reusable code for performing the exchange.

docs/internal/adr/0023-trino-view-security-model.md
johnflavin: I'm just trying to flag that I can't give a thorough review to this part, not to the level that I did (or tried to do) with the other ADRs in the PR. Those I could at least follow, maybe ask a clarifying question, at best find a hole in the explanation to patch up. With this one I can't get to even that basic level. My only concern is the weak one that, to the extent that we do want contributions to be reviewed by someone else on the team, this one hasn't been reviewed yet.

johnflavin (approved):

PR #403, opened 2026-05-27
Fix duplicate-key error on hl7_files upsert when retrying after upload failure

cmoore01:
### Product
When an HL7 ingestion activity fails partway through (e.g., transient MinIO upload error, misconfigured hl7OutputPath), Temporal's automatic retry — and any later user-initiated re-run for the same log — now succeeds instead of failing with a Postgres duplicate-key error. Previously, once an attempt failed mid-upload, no subsequent attempt for that log file could succeed without manual DB cleanup.

### Technical
The `hl7_files` upsert in `Hl7File.UPSERT_ON_CONFLICT_SQL` targeted the conflict on the `hl7_file_path` primary key, but `hl7_file_path` is not stable across attempts. When `SplitHl7LogActivityImpl`'s upload catch block runs, it writes *placeholder* rows of the form `<logPath>_<messageNumber>` into `hl7_files`. A successful retry then tries to insert rows with real S3 paths — `ON CONFLICT (hl7_file_path)` does not match the placeholders, the fall-through INSERT fires, and it trips the `UNIQUE (log_file_path, message_number)` constraint from `V1__Create_tables.sql`.

The fix retargets the upsert at the conflict that's actually invariant across attempts: `ON CONFLICT (log_file_path, message_number) DO UPDATE SET hl7_file_path = EXCLUDED.hl7_file_path, date = EXCLUDED.date`. The unique constraint already exists in V1, so no migration is required.

## Type of change
- [ ] New feature
- [ ] Improvement
- [X] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
No changes.

##### Appsec
No changes.

### Performance
No change. Upsert touches the same rows; the conflict target is a different unique constraint on the same table.

### Data
No schema change. Existing `hl7_files` rows are unaffected. On retry, rows previously stuck as placeholders will now be updated in place to point to the real S3 path. Note: any orphaned MinIO objects produced by a partially-successful first attempt are not cleaned up by this change (out of scope; same behavior as before).

### Backward compatibility
Fully backward compatible — no schema migration, no API change, no config change.

## Testing
- Added integration test `TestStatusDatabase#testStatusDbHl7OutputPathFailureRetry` that reproduces the bug end-to-end against the deployed stack: launches an ingestion with a bogus `hl7OutputPath`, then retries with a valid one, and asserts the final `hl7_files` row reflects the real S3 path. Test fails on `main`, passes with this change.
- Test includes a `cleanupHl7FilesForLog` helper that resets DB state for the target log so re-runs on a live cluster still exercise the bug rather than masking it via prior-run state.
- Full CI ingest test suite passing.

## Note for reviewers
The placeholder-row behavior in `SplitHl7LogActivityImpl`'s catch block (lines ~397–401) is what makes this bug possible and what the new ON CONFLICT target accounts for. If you prefer, the alternative shape would be to stop persisting placeholders on upload failure — but that's a larger behavioral change with implications for the `file_statuses` reporting view, so this PR keeps the placeholder behavior and fixes the upsert.

## Checklist


johnflavin (approved):
It seems like this is demoting the `hl7_file_path` from a unique ID and promoting the composite `(log_file_path, message_number)`. I think this is probably fine, but I don't know for sure.

I can see how it would fix the failure scenario that you’re trying to fix but I wonder if it is opening up new, different failure scenarios.

cmoore01:
> It seems like this is demoting the `hl7_file_path` from a unique ID and promoting the composite `(log_file_path, message_number)`. I think this is probably fine, but I don't know for sure.
> 
> I can see how it would fix the failure scenario that you’re trying to fix but I wonder if it is opening up new, different failure scenarios.

Right, but is `hl7_file_path` actually a unique ID right now when the placeholder that gets inserted doesn't match the actual s3 path that gets populated later?

PR #404, opened 2026-05-27
Dependency CVE updates

cmoore01:
### Product
N/A

### Technical
Simple update of checkstyle in all subprojects. Also includes bump of `commons-lang3` from 3.16.0 to 3.20.0 and rather aggressive pinning of `io.opentelemetry` stack from 1.49.0 to 1.62.0.

## Type of change
- [ ] New feature
- [X] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
N/A

##### Appsec
Dependencies are upgraded to remove some transitive vulnerabilities.

### Performance
N/A

### Data
N/A

### Backward compatibility
Should be backwards compatible.

## Testing
Checkstyle worked successfully for me locally in all 3 subprojects. I'm leaning on CI tests to check that in the ingest tests the extractor is still working properly.

## Checklist

### DIFF
--- a/extractor/hl7log-extractor/build.gradle
+++ b/extractor/hl7log-extractor/build.gradle
@@ -20,7 +20,7 @@ repositories {
 
 def temporalVersion = "1.34.0"
 def s3Version = "2.20.0"
-def checkstyleVersion = "10.21.2"
+def checkstyleVersion = "13.4.2"
 
 // TODO: Remove once Spring Boot ships a release that manages postgresql >= 42.7.11.
 // Overrides the Spring Boot BOM's pinned 42.7.10 to address GHSA-98qh-xjc8-98pq
--- a/extractor/hl7log-extractor/build.gradle
+++ b/extractor/hl7log-extractor/build.gradle
@@ -34,6 +34,10 @@ ext['postgresql.version'] = '42.7.11'
 // netty-codec-http2, netty-codec-dns, and netty-transport-native-epoll.
 ext['netty.version'] = '4.1.133.Final'
 
+// TODO: Remove once Spring Boot 3.5.x manages commons-lang3 >= 3.18.0.
+// Overrides the Spring Boot BOM's pinned 3.16.0 to address GHSA-j288-q9x7-2f5v.
+ext['commons-lang3.version'] = '3.18.0'
+
 // TODO: Remove once Temporal ships a release that brings opentelemetry >= 1.62.0.
 // Pins the io.opentelemetry:* family (transitively from temporal-sdk) to 1.62.0
 // to address the opentelemetry-api vulnerability CVE-2026-45292. Imported via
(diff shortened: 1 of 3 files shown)


extractor/hl7log-extractor/build.gradle
```diff
@@ -34,6 +34,21 @@ ext['postgresql.version'] = '42.7.11'
...
+ext['commons-lang3.version'] = '3.20.0'
+
+// TODO: Remove once Temporal ships a release that brings opentelemetry >= 1.62.0.
+// Pins the io.opentelemetry:* family (transitively from temporal-sdk) to 1.62.0
+// to address the opentelemetry-api vulnerability CVE-2026-45292. Imported via
+// the Spring dependency-management plugin so the BOM wins against Maven-style
+// nearest-version resolution (a plain platform() import loses to transitives).
+dependencyManagement {
+	imports {
+		mavenBom 'io.opentelemetry:opentelemetry-bom:1.62.0'
```
johnflavin, line 48: I don't know that this will get checked in CI other than the basic "does it come up?". But I also don't know what temporal does with this library and how we'd be able to tell if it broke.

There are several temporal updates we haven't applied, including a major to 1.0. Did you check if that temporal update would bump this version?

johnflavin (approved):

extractor/hl7log-extractor/build.gradle
cmoore01: > There are several temporal updates we haven't applied, including a major to 1.0. Did you check if that temporal update would bump this version?

Could you explain a bit more what you mean by this? I'm only seeing a 1.35.0 for temporal that does not address the issue. Are you referring to temporal's helm chart? This PR is about the extractor that uses the Java SDK.

It looks like temporal was already specifying this at a 1.25.0 version that was being overwritten by `spring-boot-dependencies:3.5.14` to 1.49.0. The extractor is reasonably tested by CI by the ingest tests, right? Is there something else we could do?

johnflavin: Sorry, I was thinking of the temporal helm chart / application version updates, which as you say is different from the SDK version here.

> The extractor is reasonably tested by CI by the ingest tests, right? Is there something else we could do?

My concern is about whether we're testing the specific code paths that go through this library. But I've done some research and convinced myself that the risk here is essentially zero. This library produces traces. Not only does our code not go through this library, we don't configure temporal such that it will use this library either. As long as it compiles and starts with this version (which the CI is adequately testing), we won't have any issues.

extractor/hl7log-extractor/build.gradle
```diff
@@ -34,6 +34,21 @@ ext['postgresql.version'] = '42.7.11'
 // netty-codec-http2, netty-codec-dns, and netty-transport-native-epoll.
 ext['netty.version'] = '4.1.133.Final'
 
+// TODO: Remove once Spring Boot 3.5.x manages commons-lang3 >= 3.18.0.
```
kathrynalpert, line 37: Should we consider bumping to spring boot 4 here?

kathrynalpert (approved):
Out of scope here, but checkstyle version is now pinned identically in three places, might be worth considering a way to harmonize/DRY (`gradle/libs.versions.toml`?)

extractor/hl7log-extractor/build.gradle
cmoore01: I'll merge this and then look into that for a new PR

PR #406, opened 2026-05-28
Trino AuthZ: Keycloak service accounts, user attributes, and OPA-bundle-publishing event listener 

kathrynalpert:
### Product
N/A - Second slice of the Trino per-user AuthZ feature: adding the **Keycloak side** of the design.

### Technical
- Three service-principal clients used by the impersonation-pattern clients (Superset, Voila, the Open WebUI MCP), 
- Two audience scopes that drive Trino's JWT validation and the MCP's inbound OIDC check, 
- User Profile attributes that hold a user's AuthZ scoping
  - **Row-filter dimensions** — rendered from `trino_attribute_filters` (inventory-driven; default ships only `allowed_facilities` → `sending_facility`). Adding a dimension is a one-line inventory edit: the realm template auto-renders the user-profile attribute and the rego picks up the new row filter via `data.attribute_filters`.
  - **System-level toggles** — hardcoded as binary (true/false) attributes: `mask_phi_fields` (PHI masking opt-out, per ADR 0023) and `bypass_hidden_tables` (admin escape hatch for cross-facility introspection). These aren't row-filter dimensions, but the specific columns/tables to which they apply can be overridden in inventory.
- SPI listener that publishes per-user attributes to OPA via a MinIO bundle. 
- CI security monitoring: brings the listener's gradle deps under the same Dependabot + dependency-graph umbrella as `extractor/hl7log-extractor`. Trivy already scans the bundled JAR via the keycloak image; source-level graph is the more reliable signal for transitives (commons-compress, etc.).

Design rationale and the broader picture are in ADRs 0020–0023, see #402 .

## Type of change
- [x] New feature
- [ ] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [x] Test update

##### Authorization
Adds the three impersonation-pattern service-principal clients and the User Profile attributes the OPA policy uses. Least-privilege defaults: svc clients have `serviceAccountsEnabled=true` + `standardFlowEnabled=false` + `directAccessGrantsEnabled=false`; AuthZ attrs are admin view/edit only.

##### Appsec
- The OPA bundle carries per-user attribute values plus `enabled` and `groups`. No PHI or content data crosses into the bundle.
- Snapshot is copy-on-write under a lock
- Rename/DELETE bookkeeping uses a userId→username map
- Upload is AWS SDK v2, credentials use `StaticCredentialsProvider` when both `KC_OPA_BUNDLE_S3_ACCESS_KEY` and `KC_OPA_BUNDLE_S3_SECRET_KEY` are set, otherwise `DefaultCredentialsProvider` (which handles IRSA via the WebIdentityTokenFile chain).

### Performance
Bundle publishes are debounced (1s window) so a burst of admin events coalesces into one S3 PUT.

### Data
N/A

### Backward compatibility
Realm import is upsert-style — adds new clients/scopes/attributes; existing realm objects untouched. The SPI is `@SpiType internal` (KC-SERVICES0047); the grouped Renovate PR carries a compatibility-check reminder for Keycloak-version bumps.

## Testing
- `./gradlew test` in `keycloak/event-listener/` added to CI
- End-to-end (Keycloak admin event → SPI → S3 bundle → OPA pull → Trino decision) is exercised by the `tests/data-authorization` suite landing in the next PR
- Once everything is merged onto the trino-rbac branch, I'll test as a bolus before merging to main.

## Note for reviewers
STILL IN DRAFT, I made some changes as part of preparing this PR, testing now.

**Reviewable independently but not standalone-deployable.** The realm template references `trino_attribute_filters` and the keycloak role references `s3_opa_bundle_writer` / `opa_bundles_bucket` / `minio_tenant_*`, which will be defined in the upcoming PR's `scout_common/defaults`. Deploying this PR alone would fail with undefined-variable errors.

The Renovate changes add SPI-compatibility reminders to Keycloak-version bump PRs.

## Checklist


ansible/roles/keycloak/templates/scout-realm.json.j2
```diff
@@ -29,8 +36,18 @@
...
     ],
+    {#- adminEventsEnabled drives admin-event delivery to listeners. The
+        opa-user-bundle-publisher listener (ADR 0021) consumes USER admin
+        events (CREATE, UPDATE, DELETE) to keep its in-memory snapshot
+        in sync and republishes the OPA bundle to MinIO on every change.
+        adminEventsDetailsEnabled exposes event.getDetails(), which the
+        listener reads on DELETE to recover the deleted user's username
+        (the user row itself is gone by event-fire time). -#}
+    "adminEventsEnabled": true,
+    "adminEventsDetailsEnabled": true,
```
johnflavin: These were already enabled above on lines 13 and 14. I don't know that it hurts anything, so maybe this is a Nit, but it seems like something we want to avoid.

renovate.json5
```diff
@@ -84,6 +113,28 @@
...
+      "datasourceTemplate": "docker",
+      "depNameTemplate": "quay.io/keycloak/keycloak"
+    },
+    {
+      "customType": "regex",
+      "managerFilePatterns": ["keycloak/Dockerfile"],
+      "matchStrings": [
+        "FROM\\s+(?<depName>quay\\.io/keycloak/keycloak):(?<currentValue>[\\d\\.]+)"
+      ],
+      "datasourceTemplate": "docker"
```
johnflavin, line 137: Should we add a rule to also bump `kcVersion` in `keycloak/event-listener/build.gradle`?

keycloak/event-listener/src/main/java/edu/washu/tag/keycloak/events/OpaUserBundlePublisherProviderFactory.java
```diff
@@ -0,0 +1,339 @@
...
+        // common startup contention without making cold-start painful.
+        scheduler.schedule(() -> seedFromKeycloak(factory), INITIAL_PUBLISH_DELAY_MS, TimeUnit.MILLISECONDS);
+    }
+
+    @Override
+    public void close() {
+        if (scheduler != null) {
+            scheduler.shutdownNow();
+        }
+    }
```
johnflavin, line 157: Nit: this could/should close `uploader`, which in turn should close its `S3Client` instance.

johnflavin (approved):

andylassiter (approved):

PR #407, opened 2026-05-28
Symlink Scout sample notebooks instead of first-spawn copy

andylassiter:
### Product
On the Scout JupyterHub, sample notebooks (Quickstart, Advanced, plus the conda env file) used to be copied into `~/Scout/` on first login and never refreshed. Existing users never saw new samples or updates without manual intervention. This change makes `~/Scout/` a read-only view of the current release that updates automatically on each release. Users open notebooks in place to run them and use **File > Save Notebook As...** to keep an editable copy.

### Technical
- Replaces the legacy `~/Scout/` directory with a symlink to `/opt/scout/samples` (RO). Users can run notebooks in place but need to save copies into their home directory if they want to keep any changes. JupyterLab prompts the user to Save or Discard when closing.
- For existing users whose `~/Scout/` is a real directory from the old first-spawn hook, postStart moves it to `~/Scout.bak-<date>/` before symlinking. Their previous content is preserved, just relocated.

## Type of change
- [ ] New feature
- [x] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
N/A

##### Appsec
N/A

### Performance
Backing up existing `~/Scout` dirs could be slow if users have put other files in the directory.

### Data
User data is preserved:
- Existing `~/Scout/` directories from the old hook are moved to `~/Scout.bak-<date>/` on first spawn under the new code. Nothing is deleted.
- The migration is idempotent: once `~/Scout` is a symlink, the move is skipped on subsequent spawns.

### Backward compatibility
- Existing users keep their old content under `~/Scout.bak-<date>/`.
- The `.scout_quickstart` flag file is no longer created or read; it can sit harmlessly on old user homes.

## Testing
Deployed on dev02 and verified by hand:
- Fresh spawn shows `~/Scout/` with README, Quickstart, Advanced, environment.yml
- Existing user with legacy `~/Scout/` sees it moved to `~/Scout.bak-<date>/` and `~/Scout/` becomes the new symlink
- Notebooks open and run in place from the read-only mount, outputs render, JupyterLab shows the read-only banner
- `mamba env create -f ~/Scout/environment.yml` succeeds
- Bulk-lookup cells in Quickstart write `~/patient_ids.csv` and `~/patient_list_results.csv` successfully

## Note for reviewers
We considered `~/scout-samples/` and `~/Scout/samples/` but landed on `~/Scout/` per review feedback. Avoids the extra dir level and keeps the path users already know.

## Checklist
- [x] My code adheres to the coding and style guidelines of the project (and I have run `pre-commit run --all-files`)


kathrynalpert (approved):
Super excited about this! I could also see a submit/approve/publish flow either via GitHub PRs or swapping to pulling these notebooks from S3 or something.

andylassiter:
> Super excited about this! I could also see a submit/approve/publish flow either via GitHub PRs or swapping to pulling these notebooks from S3 or something.

Yeah I could see pushing this feature further. I was thinking something Git related but I also like the S3 idea.

johnflavin (approved):
I think this is generally a good idea, but I don't like the `~/scout-samples` directory name. Nor do I like the alternative `~/Scout/samples`. I just want to put these directly in `~/Scout`. We'd need to check if any existing user notebook PVs have modified the default files, but if not, I don't see a reason we can't move or remove things from the existing `~/Scout` directory.

andylassiter:
> I think this is generally a good idea, but I don't like the ~/scout-samples directory name. Nor do I like the alternative ~/Scout/samples. I just want to put these directly in ~/Scout. We'd need to check if any existing user notebook PVs have modified the default files, but if not, I don't see a reason we can't move or remove things from the existing ~/Scout directory.

I can revert to `~/Scout`. We could try to handle this in the startup script and just backup the existing dir? 

```
if [ -d ~/Scout ] && [ ! -L ~/Scout ]; then
    mv ~/Scout ~/Scout.bak-$(date +%Y%m%d)
fi
ln -sfn /opt/scout/samples ~/Scout
```

johnflavin (approved):

PR #409, opened 2026-05-29
Trino AuthZ: OPA deployment, Trino HTTPS + OPA enforcement, Superset impersonation

kathrynalpert:
### Product
N/A - Third slice of the Trino per-user AuthZ feature: **end-to-end data authZ via Superset.** Trino now enforces the policy from #402 against the user attributes published by #406, with Superset as the first client.

### Technical
**OPA infra**
- New `scout-opa` Helm chart (`helm/scout-opa/`) owns the Deployment + Service + policy ConfigMap shape — matches the chart-driven pattern every other Scout component uses and keeps the deployment declarative for future GitOps. The `opa` role sources `policy/trino/main.rego`, renders the static data document (`hidden_tables`, `view_owner_principals`, `attribute_filters`, `masked_columns`) and the bundle plugin config from inventory, hashes the bundle-reader credentials, and helm-upgrades. `persist: true` writes the last good bundle to a pod-local emptyDir; `roots: ["users"]` scopes the atomic swap to `data.users` so the static document isn't replaced on every pull. Readiness gate on first bundle load.
- MinIO: `opa-bundles` bucket with versioning enabled (rollback on a bad publish); `opa-bundle-writer` and `opa-bundle-reader` service accounts + scoped IAM policies (writer can only PUT the one object, reader can only GET it).
- `scout_common`: `opa_namespace`, bundle poll-interval defaults; `cluster_issuer_setup` task bootstraps the Scout internal CA via cert-manager (SelfSigned bootstrap → CA Certificate → CA ClusterIssuer).

**Trino**
- HTTPS listener via cert-manager-issued cert from the new internal CA; HTTP listener retained for in-cluster probes (chart limitation, not a misconfiguration).
- OPA access-control plugin wired to `http://opa-trino.scout-analytics:8181`.
- JWT-only authentication: validates `aud=trino` against the Keycloak JWKS, per ADR 0022. User JWTs flow from Jupyter (later slice), service-account JWTs from Superset (this slice) and Voila (later slice). The PASSWORD authenticator that upstream `tuannvm/mcp-trino` needs ships with the MCP slice, not here — PR3 has no MCP consumer to feed.
- Split deployment playbook from superset to acknowledge broader Trino role.

**Superset (first impersonating client)**
- `superset_trino_auth.py`: a `DB_CONNECTION_MUTATOR` that mints a `superset_svc` client-credentials token from Keycloak, caches it process-locally until ~80% of its lifetime has elapsed, then refreshes. The logged-in Superset user is impersonated via `X-Trino-User`. OPA's impersonation rule permits `superset_svc → any user`; the actual data-access decision uses the impersonated user's Keycloak attributes.

**Observability**
- Grafana `opa-decisions` dashboard: allow/deny rate, bundle revision (extracted via a regex query, not a full JSON parse, so it's cheap), per-table decision latency.

**Tests** (smoke-test wiring)
- `tests/data-authorization/seed/*` — a one-shot Job in `scout-data` that reuses the hl7-transformer image (PySpark + Delta + Hive client) to write 6 synthetic rows to `delta.default.test_reports`.
- `tests/data-authorization/authz-tests/*` — a one-shot Job in `scout-analytics` (alpine/curl + jq + bash) that walks 8 scenarios against the live stack: Keycloak admin API → SPI → MinIO bundle → OPA pull → Trino JWT + X-Trino-User → expected row count / masked column / 403.
- CI wiring uses `kubectl create configmap --from-file` (no escape-interpretation pitfalls) and `yq` for the in-place image/warehouse-dir substitution in the seed Job manifest. `wait-for-job.sh` races complete/failed so a Failed Job returns immediately instead of stalling until the per-condition timeout.

Design rationale in ADRs 0020–0023, see #402.

## Type of change
- [x] New feature
- [ ] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [x] Test update

##### Authorization
First PR where OPA is consulted on every Trino query. Default-deny via the `approved_groups` gate (must be a member of `scout-user` or `scout-admin` — gate is hardcoded in `policy/trino/main.rego`); attribute-driven row filters and column masks per the user's Keycloak attributes (PR #406); hidden-table baseline blocks the cross-facility patient-mapping table. Empty/missing attributes deny rows and mask PHI rather than expose. Fail paths exercised in scenario 4 (unset → `1=0`) and scenario 8 (`enabled=false` → all-deny).

##### Appsec
- Trino HTTPS uses the Scout internal CA (cert-manager-issued); the HTTP listener is retained only for probes (in-cluster, not ingress-exposed).
- Superset's service-principal token is cached **process-locally** (one per Superset worker), with refresh at 80% lifetime; the cache is never written to disk and dies with the worker. No cross-tenant token leakage path.
- Test runner verifies Trino TLS via the cert-manager-issued `trino-tls` CA mounted from a Secret (no `curl -k` fallback in the happy path).

### Performance
- OPA bundle poll: 5–15s (configurable via `opa_bundle_min_delay_seconds` / `opa_bundle_max_delay_seconds`), comfortably under the propagation budget in ADR 0021.
- OPA decision latency is per-query, in-process — the policy is small (one `allow` + four data lookups). Dashboard shows p99 in the single-digit-millisecond range.
- `opa_replicas: 1` in CI for test determinism (multi-replica polling lag confuses scenario assertions); production stays at the role default of 2 for availability.

### Data
- The seed Job writes to `delta.default.test_reports`, a synthetic table that lives only in CI. Production deployments are unaffected; `test_reports` is added to `trino_filtered_tables` only in the CI inventory.

### Backward compatibility
- Trino now requires JWT + HTTPS on the client path; the HTTP listener stays up only for internal worker→coordinator comms and the unauthenticated `/v1/info` probe.
- **Existing clients break in transit**: Voila and MCP (still pointed at `http://trino:8080`, no JWT) get 401 "Authentication over HTTP is not enabled" on every query; Jupyter is blocked at the NetworkPolicy layer (egress only opens port 8080, not 8443). Each consumer's own slice repoints to HTTPS + adds JWT auth — Voila/Jupyter via service-principal client_credentials, MCP via the PASSWORD authenticator. Smoke-test still passes because no scenario exercises a Voila/Jupyter/MCP→Trino query; the data-authorization Job goes through Superset, which is updated in this PR.
- New required inventory secrets: `s3_opa_bundle_reader_secret`, `trino_internal_shared_secret`. Realm template already references the matching Keycloak entries from #406.

## Testing
- 8-scenario integration suite (`tests/data-authorization/authz-tests/run.sh`) runs in CI's smoke-test job and exercises the full Keycloak → SPI → MinIO → OPA → Trino path with a real JWT.
- `opa test policy/` (added in PR #402) continues to enforce policy correctness independent of the deployed stack.
- Once everything is merged onto the `trino-rbac` branch, I'll test as a bolus before merging to `main`.

## Note for reviewers
**Voila / Jupyter / MCP are broken-downstream on this PR** — their HTTPS+auth migrations are separate slices.

**Deferred to later slices**: dropping `pyspark-notebook` ships with the Voila PR (where Voila adopts the shared image); the Trino PASSWORD authenticator + bcrypt `password.db` ships with the MCP PR (where the upstream `tuannvm/mcp-trino` consumer arrives).

Design rationale and the broader picture are in ADRs 0020–0023, see #402.

## Checklist


ansible/roles/opa/tasks/deploy.yaml
```diff
@@ -0,0 +1,130 @@
...
+# inventory, hash the bundle-reader credentials, and hand the resulting
+# values to helm upgrade. Per-user attribute data is layered on top by
+# OPA's bundle plugin (ADR 0021): the Keycloak SPI listener publishes a
+# bundle to MinIO; OPA polls and atomically swaps the `data.users` subtree.
+
+- name: Resolve OPA policy path on disk
+  ansible.builtin.set_fact:
+    # policy/trino/ at repo root. Relative from this role:
+    #   ansible/roles/opa/tasks → ../../../../policy/trino
+    opa_policy_path: '{{ role_path }}/../../../policy/trino/main.rego'
```
johnflavin: Nit: Using `{{ scout_repo_path }}` seems easier than traversing from the role dir.

ansible/roles/scout_common/defaults/main.yaml
```diff
@@ -236,8 +251,6 @@ keycloak_subdomain: keycloak
 enable_chat: false
 enable_playbooks: false
 
-# Reserve GPU nodes for workloads that either need GPU hardware (Ollama,
-# Jupyter singleuser) or are cluster-wide storage layers that span every
```
johnflavin, line 240: Nit: I think this comment got removed accidentally.

ansible/roles/scout_common/tasks/cluster_issuer_setup.yaml
```diff
@@ -0,0 +1,78 @@
...
+#
+# Idempotent; can be called from multiple roles. The Certificate has a long
+# duration (10y) and renews 1y before expiry, so re-deploys don't churn it.
+
+- name: Create Scout SelfSigned bootstrap ClusterIssuer
+  delegate_to: localhost
+  kubernetes.core.k8s:
+    state: present
+    kubeconfig: '{{ local_kubeconfig_yaml }}'
+    definition:
```
johnflavin, line 18: Nit: This task (and others added in this file) don't need to be delegated

ansible/roles/trino/defaults/main.yaml
```diff
@@ -7,6 +7,18 @@ trino_helm_repo_name: trino
...
 
+# Trino AuthZ / TLS / OPA configuration (ADR 0020 authorization, ADR 0022 auth).
+# JWT auth requires TLS — Trino terminates HTTPS directly using a cert from
+# the Scout internal CA (cert-manager); the chart's HTTP listener is disabled.
+# The user-facing Service port (`trino_https_port`) is defined in
+# scout_common/defaults/main.yaml so consumer roles can reference it without
+# loading the trino role.
+trino_keystore_password: trinokeystorepass # protects the in-pod keystore; not a network secret
+
+trino_keycloak_jwks_url: 'http://keycloak-service.{{ keycloak_namespace }}:8080/realms/{{ keycloak_realm_name }}/protocol/openid-connect/certs'
```
johnflavin: Nit: This should be built from `{{ keycloak_internal_url }}`

ansible/roles/trino/tasks/deploy.yaml
```diff
@@ -39,6 +39,58 @@
...
+    name: scout_common
+    tasks_from: cluster_issuer_setup
+  when: not aws_deployment
+
+- name: Create Trino AuthZ env Secret (keystore password + internal shared secret)
+  delegate_to: localhost
+  no_log: true
+  kubernetes.core.k8s:
+    state: present
+    kubeconfig: '{{ local_kubeconfig_yaml }}'
```
johnflavin: Nit: I don't _think_ the tasks added to this file need to delegate

ansible/roles/opa/defaults/main.yaml
```diff
@@ -0,0 +1,120 @@
+---
+# OPA role defaults
+#
+# OPA enforces Trino's access-control policy (see policy/trino/main.rego).
+# Deployed as a Deployment + Service in the same namespace as Trino so the
+# query path stays in-cluster (and so a single NetworkPolicy on OPA can
+# allow only the Trino coordinator).
```
johnflavin: It does not appear that there is a network policy defined in the OPA chart. Is that ok, or was that missed?

johnflavin (approved):
Mostly nits and one question.

ansible/roles/opa/defaults/main.yaml
kathrynalpert: Great catch, it was missed. Added it and some tests

PR #410, opened 2026-06-01
POC Scout XNAT auth plugin + python lib

johnflavin:
### Product
No direct product implications yet.

### Technical
- Adds `scout-xnat-auth-plugin`, an XNAT plugin which can do two kinds of JWT token auth in Scout's environment
  - One for direct user browser access that pulls user info from the `X-Auth-Request-Access-Token` header, set by oauth2-proxy
  - Another for internal API access (from user Jupyter notebooks or other services) that takes a user's JWT token for that other service, exchanges it for an `xnat` service token, then auths the user with that
- Adds `scout-xnatpy`, a python library that wraps `xnatpy` and performs the jupyter token auth steps on the user's behalf. The user just connects to xnat in their notebook and `scout-xnatpy` gets the token, auth to XNAT, and keeps a `JSESSIONID` going in the background. The user then interacts with the API via `xnatpy`'s existing functionality.
- Keycloak realm: adds an xnat client
- Jupyter: Configures hub permissions so the notebook user can access their own Jupyter JWT through the hub API (which they use to auth to XNAT)

## Type of change
- [X] Work behind a feature flag (the xnat plugin + python client isn't deployed yet)
- [ ] New feature
- [ ] Improvement
- [X] Change (Some KC realm config, an oauth2-proxy header forwards the user JWT, and a jupyterhub api permission)
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
- The `scout-xnat-auth-plugin` adds a new auth mechanism to XNAT which uses JWT tokens acquired from other places, either directly from keycloak via oauth2-proxy or from user API requests in jupyter or other scout-internal services. 
- We add the `xnat-access` KC role that gets assigned to a user along with the `scout-user` group. We could separate those if we wanted to but I left them coupled for now.
- Add the `xnat` audience to Jupyter's JWT tokens, and configure the `xnat` KC client to exchange Jupyter tokens for XNAT tokens.

I didn't implement any kind of XNAT permission mapping, as in a way to manage XNAT project / data permissions in keycloak, merely an `xnat-access` role for whether they can access XNAT or not. Any user who can get to XNAT can log in and see their projects if they've created any. I considered adding an `xnat-admin` role tied to the `scout-admin` group, and, while that is possible to do, it was unnecessary scope for this POC.

##### Appsec
We now populate the `X-Auth-Request-Access-Token` header in `oauth2-proxy`. This forwards the user's keycloak JWT to all downstream services. We did this so it can be accessed by XNAT directly and jupyterhub -> user notebooks -> XNAT. 

The JWT tokens aren't forgeable without the service secrets. A user's session could in theory be hijacked by an attacker who gets their cookies, but that isn't new.

A user can't set their own `X-Auth-Request-Access-Token` from outside because it gets stripped by traefik, then set by OA2P.

The user could in principle set this header in their notebook with the contents being the JWT, but that gets stopped by the HeaderTrustFilter which verifies that only `oauth2-proxy` tokens are valid. Plus, the notebook user could just use their JWT as a Bearer token and it would work, so why would they try to set it on the header anyway?

### Performance
I haven't really considered performance, but I don't think it will be a big additional load on either Scout as a whole or XNAT specifically. There's nothing automated here, only user actions trigger auth to happen.

### Data
N/A

### Backward compatibility
This only adds a few non-deployed XNAT integrations and some minor changes into KC, OA2P, and JH. Everything should be backwards compatible.

The only incompatibilities I could think of are
- The keycloak XNAT client does need a secret set in the inventory. If your inventory doesn't have that the deploy will fail.
- Right now nothing creates the `xnat` namespace, but jupyter notebooks are set up to allow egress to that namespace. I don't know if that will cause a problem or not. 

## Testing
I deployed all of this along with an XNAT and a POC XNAT-UI service to `dev03`. Everything worked together.

I have also deployed just the changes on this branch to `dev03` and verified that nothing breaks.

## Note for reviewers
I don't know if the jupyter token exchange is even required. Meaning, a notebook user submits their jupyter JWT to XNAT and the plugin exchanges that in KC for an XNAT token. I thought this exchange was necessary, but reviewing #402 which lets trino take other services' tokens without exchange made me question this assumption.

## Checklist
- [X] I have added unit tests, if appropriate


kathrynalpert:
Is the XNAT ansible deployment coming in a subsequent, but soon, PR? And will that PR string up the publish of this plugin as a container image that the XNAT helm chart can use?

johnflavin:
> Is the XNAT ansible deployment coming in a subsequent, but soon, PR? And will that PR string up the publish of this plugin as a container image that the XNAT helm chart can use?

That's right. I haven't wired up the XNAT deploy into our ansible but I do plan to do that soon. My plan was for this plugin jar to be built + published in CI and pulled into XNAT via gradle in an init container; I wasn't planning to package it into a container itself. That should work even air-gapped because we can pull jars through the nexus proxy. Does that sound reasonable?

kathrynalpert:
> My plan was for this plugin jar to be built + published in CI and pulled into XNAT via gradle in an init container; I wasn't planning to package it into a container itself. That should work even air-gapped because we can pull jars through the nexus proxy. Does that sound reasonable?

It does, but it mildly conflicts with the [current state of the XNAT helm chart](https://github.com/NrgXnat/helm-charts/blob/4d6f011f6983dfd282abb3be09f0a29b6bfad3be/helm/xnat/templates/statefulset.yaml#L120). I think it's a nicer approach, though, so I'd like to see you contribute the change back to the XNAT helm repo.

cmoore01:
> > My plan was for this plugin jar to be built + published in CI and pulled into XNAT via gradle in an init container; I wasn't planning to package it into a container itself. That should work even air-gapped because we can pull jars through the nexus proxy. Does that sound reasonable?
> 
> It does, but it mildly conflicts with the [current state of the XNAT helm chart](https://github.com/NrgXnat/helm-charts/blob/4d6f011f6983dfd282abb3be09f0a29b6bfad3be/helm/xnat/templates/statefulset.yaml#L120). I think it's a nicer approach, though, so I'd like to see you contribute the change back to the XNAT helm repo.

Could that replace entirely the need to have plugins published as images? That has always seemed wildly clunky to me and I would love to remove it if we could.

scout-xnat-auth-plugin/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,7 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.14.4-bin.zip
```
kathrynalpert, line 3: Any reason not to use gradle 9?

ansible/roles/oauth2-proxy/templates/values.yaml.j2
```diff
@@ -24,6 +24,11 @@ config:
     banner = "{{ oauth2_proxy_banner }}"
     footer = "{{ server_hostname }}" # Hacky way to pass server hostname to sign_in template
     upstreams = ["static://200"]
+    # Emit the upstream access token from /oauth2/auth so Traefik's ForwardAuth
+    # middleware can forward it. The access token (a signed Keycloak JWT) is the
+    # single source of truth for services that trust oauth2-proxy as the
+    # identity source (eg. scout-xnat-auth-plugin's HeaderTrustFilter).
+    pass_access_token = true
```
kathrynalpert, line 39: Claude is pretty insistent that you need `set_xauthrequest = true` here, as well:
>  - --set-xauthrequest: "…When used with --pass-access-token, X-Auth-Request-Access-Token is added to response headers."
>  - --pass-access-token: "pass OAuth access_token to upstream via X-Forwarded-Access-Token header. When used with --set-xauthrequest this adds the X-Auth-Request-Access-Token header to the response."
>  
> And the v7.7.1 source agrees — the response header is appended only inside if SetXAuthRequest { … if PassAccessToken { … } }. So pass_access_token alone produces only X-Forwarded-Access-Token (an upstream request header), which isn't in Scout's Traefik authResponseHeaders and is meaningless in the static://200 ForwardAuth mode anyway. It does not produce the X-Auth-Request-Access-Token that HeaderTrustFilter reads.

scout-xnat-auth-plugin/src/main/java/edu/wustl/scout/xnat/auth/service/DefaultUserProvisioningService.java
```diff
@@ -0,0 +1,100 @@
...
+        xdatUser.setEmail(StringUtils.defaultString(identity.getEmail(), ""));
+        xdatUser.setFirstname(StringUtils.defaultString(identity.getFirstName(), identity.getPreferredUsername()));
+        xdatUser.setLastname(StringUtils.defaultString(identity.getLastName(), ""));
+        xdatUser.setEnabled(true);
+        xdatUser.setVerified(true);
+
+        try {
+            final UserI admin = Users.getAdminUser();
+            final XdatUserAuth auth = new XdatUserAuth(
+                    identity.getSub(), authMethod, providerId, xdatUser.getLogin(), true, 0);
```
kathrynalpert, line 103: Should this be `buildUsername(identity.getSub())` per the `provision` method, above? Or vice-versa.

scout-xnat-auth-plugin/src/main/java/edu/wustl/scout/xnat/auth/security/BearerTokenFilter.java
```diff
@@ -0,0 +1,175 @@
...
+        try {
+            final UserI user = provisioningService.provision(identity);
+            final ScoutAuthenticationToken token = new ScoutAuthenticationToken(user, "keycloak");
+            SecurityContextHolder.getContext().setAuthentication(token);
+            // Mirror XDAT.loginUser: populate the session-scoped UserHelper so the
+            // Velocity browser path (eg. project pages) sees a usable permission
+            // model instead of the "no access" fallback. Required for browser
+            // round-trips that follow a bearer-authenticated REST call.
+            UserHelper.setUserHelper(request, user);
+        } catch (AuthenticationException e) {
```
kathrynalpert: Should this also handle `AccessDeniedException`?

scout-xnat-auth-plugin/README.md
```diff
@@ -0,0 +1,71 @@
...
+## Build
+
+```bash
+JAVA_HOME=$(/usr/libexec/java_home -v 1.8) PATH=$JAVA_HOME/bin:$PATH \
+  ./gradlew --no-daemon xnatPluginJar
+```
+
+Output: `build/libs/scout-xnat-auth-plugin-0.1.0-SNAPSHOT-xpl.jar`.
+
+Gradle 8.10.2 (vendored as the wrapper) needs Java 8–22.
```
kathrynalpert: Nit: This isn't the version of gradle you're using

scout-xnat-auth-plugin/src/main/java/edu/wustl/scout/xnat/auth/security/HeaderTrustFilter.java
```diff
@@ -0,0 +1,195 @@
...
+    private static String claimAsString(final JWTClaimsSet claims, final String name) {
+        try {
+            return claims.getStringClaim(name);
+        } catch (java.text.ParseException e) {
+            return null;
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    static List<String> extractClientRoles(JWTClaimsSet claims, String clientId) {
```
kathrynalpert: It appears this method is duplicated (see line 55 of scout-xnat-auth-plugin/src/main/java/edu/wustl/scout/xnat/auth/security/JwtValidator.java)

scout-xnat-auth-plugin/src/main/java/edu/wustl/scout/xnat/auth/security/BearerTokenFilter.java
```diff
@@ -0,0 +1,175 @@
...
+    private static List<String> stringListClaim(final JWTClaimsSet claims, final String name) {
+        Object value = claims.getClaim(name);
+        if (value instanceof List) {
+            return (List<String>) value;
+        }
+        return Collections.emptyList();
+    }
+
+    @SuppressWarnings("unchecked")
+    private static List<String> realmRoles(final JWTClaimsSet claims) {
```
kathrynalpert: Is this the same method as `extractRealmRoles` in scout-xnat-auth-plugin/src/main/java/edu/wustl/scout/xnat/auth/security/HeaderTrustFilter.java?

scout-xnat-auth-plugin/src/main/java/edu/wustl/scout/xnat/auth/security/HeaderTrustFilter.java
```diff
@@ -0,0 +1,195 @@
...
+            // Don't echo the provisioning exception detail to the client — it can
+            // leak internal state. The reason is logged above for operators.
+            response.sendError(HttpServletResponse.SC_FORBIDDEN, "Scout auth: user provisioning failed");
+            return;
+        }
+
+        chain.doFilter(request, response);
+    }
+
+    private static String claimAsString(final JWTClaimsSet claims, final String name) {
```
kathrynalpert: Is this duplicated in scout-xnat-auth-plugin/src/main/java/edu/wustl/scout/xnat/auth/security/BearerTokenFilter.java 147?

scout-xnat-auth-plugin/src/main/java/edu/wustl/scout/xnat/auth/security/HeaderTrustFilter.java
```diff
@@ -0,0 +1,195 @@
...
+        final String email = claimAsString(claims, "email");
+        final String firstName = claimAsString(claims, "given_name");
+        final String lastName = claimAsString(claims, "family_name");
+
+        // The xnat-access gate is enforced solely from the token's role claims:
+        // client roles in resource_access.<client>.roles, plus realm roles in
+        // realm_access.roles. No group-name inference fallback.
+        final List<String> roles = new ArrayList<>();
+        roles.addAll(extractClientRoles(claims, properties.getClientId()));
+        roles.addAll(extractRealmRoles(claims));
```
kathrynalpert: I'm surprised we want to check both client and realm roles here, I would assume we wanted one or the other.

scout-xnat-auth-plugin/src/main/java/edu/wustl/scout/xnat/auth/security/BearerTokenFilter.java
```diff
@@ -0,0 +1,175 @@
...
+
+    private ScoutIdentity identityFrom(final JWTClaimsSet claims) {
+        final String sub = claims.getSubject();
+        final String preferredUsername = claimAsString(claims, "preferred_username");
+        final String email = claimAsString(claims, "email");
+        final String firstName = claimAsString(claims, "given_name");
+        final String lastName = claimAsString(claims, "family_name");
+        final List<String> groups = stringListClaim(claims, "groups");
+        final List<String> roles = new ArrayList<>(JwtValidator.extractClientRoles(claims, properties.getClientId()));
+        roles.addAll(realmRoles(claims));
```
kathrynalpert: Same question about client/realm roles - and also could this be DRY so both Bearer and Header filters use the same logic?

scout-xnat-auth-plugin/src/main/java/edu/wustl/scout/xnat/auth/security/BearerTokenFilter.java
```diff
@@ -0,0 +1,175 @@
...
+        chain.doFilter(request, response);
+    }
+
+    private ScoutIdentity identityFrom(final JWTClaimsSet claims) {
+        final String sub = claims.getSubject();
+        final String preferredUsername = claimAsString(claims, "preferred_username");
+        final String email = claimAsString(claims, "email");
+        final String firstName = claimAsString(claims, "given_name");
+        final String lastName = claimAsString(claims, "family_name");
+        final List<String> groups = stringListClaim(claims, "groups");
```
kathrynalpert: >We add the `xnat-access` KC role that gets assigned to a user along with the `scout-user` group. We could separate those if we wanted to but I left them coupled for now.

As far as I can tell, `groups` isn't used anywhere. Can you explain roles vs groups for this functionality?

scout-xnat-auth-plugin/src/main/java/edu/wustl/scout/xnat/auth/security/HeaderTrustFilter.java
```diff
@@ -0,0 +1,195 @@
...
+            return;
+        }
+
+        final ScoutIdentity identity = new ScoutIdentity(
+                sub, preferredUsername, email, firstName, lastName, Collections.emptyList(), roles);
+        log.debug("HeaderTrustFilter authenticating {}", identity);
+
+        try {
+            final UserI user = provisioningService.provision(identity);
+            final ScoutAuthenticationToken token = new ScoutAuthenticationToken(user, "keycloak");
```
kathrynalpert: Nit: magic string `"keycloak"` again

scout-xnat-auth-plugin/src/main/java/edu/wustl/scout/xnat/auth/security/BearerTokenFilter.java
```diff
@@ -0,0 +1,175 @@
...
+            response.sendError(HttpServletResponse.SC_FORBIDDEN, "exchanged token rejected");
+            return;
+        }
+
+        // Step 4: provision and attach.
+        final ScoutIdentity identity = identityFrom(claims);
+        log.debug("BearerTokenFilter authenticating {}", identity);
+        try {
+            final UserI user = provisioningService.provision(identity);
+            final ScoutAuthenticationToken token = new ScoutAuthenticationToken(user, "keycloak");
```
kathrynalpert: Nit: magic string "keycloak" - does this need to match DefaultUserProvisioningService's providerId?

kathrynalpert (commented):
I think the question of token exchange is one to revisit. I'm pretty confident you don't _need_ it (you could use the `xnat` audience that's already on the Jupyter token and remove the `azp=xnat` requirement on the XNAT side), but do you _want_ it? It seems like exchange might be more standard/auditable, at the cost of some maintenance burden/complexity. And I'd really like to understand if we gain anything from a security perspective before we accept that complexity.

I would argue that the way exchange is implemented here might be the worst of both worlds. XNAT exchanges a token only if `xnat` is already in the Jupyter token's audience, at which point that token is already a valid XNAT credential; the exchange is just ceremony. And I'm not sure we can get around that here, since the credential has to come client-side (we're in a notebook). Moreover, this approach means we have to reimplement the exchange logic in multiple places - i.e. if we want to exchange a Jupyter token for Trino token, we can't use the `TokenExchangeService.java‎` in this XNAT plugin, we'd have to reimplement on Trino side.

Interested in your thoughts, maybe I'm missing something.

scout-xnat-auth-plugin/gradle/wrapper/gradle-wrapper.properties
johnflavin: Yes. We're targeting XNAT 1.9.x for now, which is still on Java 8. I think in theory we could use Java 17+ to _build_ the plugin but still target Java 8 bytecode, possibly, but we're also using some old builder plugins that aren't supported by Gradle 9. 

I do want to upgrade this to target XNAT 1.10, JDK at least 17, get rid of some of the old plugins and dependencies and whatever, and get on Gradle 9. But it didn't seem like time well spent.

scout-xnat-auth-plugin/README.md
johnflavin: Sorry, that's my bad. I initially started with essentially a copy of the OIDC plugin's build which is (I think) on this older gradle and then I updated and cleaned some things up, but missed this in the docs.

ansible/roles/oauth2-proxy/templates/values.yaml.j2
johnflavin: I've confirmed that my own Claude agrees. But I'm confused. This _did_ work, I was able to successfully log in through the browser, which tells me that the plugin did get the token off this header in XNAT. So why did it work if the robots think it shouldn't have? I will need to dig a little further on this and see what's going on.

kathrynalpert: It looked like you had this at one point and then removed it, so perhaps the redeploy didn't unset the setting?

johnflavin: Ah, I see. I had this set at one point but accidentally removed it during a refactor. Mystery solved.

scout-xnat-auth-plugin/src/main/java/edu/wustl/scout/xnat/auth/security/BearerTokenFilter.java
johnflavin: I think there are two uses of "group" that are being conflated.
1. You quoted the description I wrote about the `scout-user` group. That's a keycloak concept which is how we grant the `xnat-access` role to users.
2. The "groups" claim in the JWT, which is what this line of code reads. I know less about that; JWTs are a bit of a black box to me. Or... maybe an obscured box? I know a little but not much. It does seem that you’re right, this line is reading something that is never used. I'll remove it.

scout-xnat-auth-plugin/src/main/java/edu/wustl/scout/xnat/auth/security/HeaderTrustFilter.java
johnflavin: I believe you’re right. The `xnat-access` role is only on the `xnat` KC client and gets set on the JWT's client roles, so there's no need to check realm roles here.

scout-xnat-auth-plugin/src/main/java/edu/wustl/scout/xnat/auth/security/BearerTokenFilter.java
kathrynalpert: I'm pretty sure the groups claim holds the scout user groups, e.g, `scout-user`/`scout-admin`

johnflavin: But I'm guessing since we have a group mapper defined for the `scout-user` group -> `xnat-access` role that means the role gets set on the JWT; we can just check `roles` and don't need to check the `groups`. Does that match your understanding?

kathrynalpert: I think that's right, but more generally, [I this we should revisit how/whether token exchange is implemented](https://github.com/washu-tag/scout/pull/410#pullrequestreview-4404726085).

johnflavin: Short story: I intend to rip out the token exchange. Longer comment incoming after I can redeploy and test.

johnflavin:
> I think the question of token exchange is one to revisit. I'm pretty confident you don't _need_ it (you could use the `xnat` audience that's already on the Jupyter token and remove the `azp=xnat` requirement on the XNAT side), but do you _want_ it? It seems like exchange might be more standard/auditable, at the cost of some maintenance burden/complexity. And I'd really like to understand if we gain anything from a security perspective before we accept that complexity.
> 
> I would argue that the way exchange is implemented here might be the worst of both worlds. XNAT exchanges a token only if `xnat` is already in the Jupyter token's audience, at which point that token is already a valid XNAT credential; the exchange is just ceremony. And I'm not sure we can get around that here, since the credential has to come client-side (we're in a notebook). Moreover, this approach means we have to reimplement the exchange logic in multiple places - i.e. if we want to exchange a Jupyter token for Trino token, we can't use the `TokenExchangeService.java‎` in this XNAT plugin, we'd have to reimplement on Trino side.
> 
> Interested in your thoughts, maybe I'm missing something.

I thought about it, and it seems unnecessary. When I was initially planning out this plugin it made sense to get a jupyter token and need to exchange that for an xnat token. But once that was implemented it didn't work. Keycloak won't let you exchange a token unless the initial token (jupyterhub in this example) has the destination client (xnat) in its audience. But when you have that, then there's seemingly no reason to exchange at all; the original token is already valid for the destination service. 

I've done a big refactor to remove the token exchange entirely and streamline the token checks done in the two auth paths. I was also able to remove most of the changes to the scout realm (and drop the required `keycloak_xnat_client_secret` param) by making the `xnat` client `bearerOnly`; nothing actually mints an `xnat` client but it is available as an audience and in group mappings.

kathrynalpert (approved):
Last nit that maybe some `log.info` should be `log.debug`. Otherwise, LGTM!

johnflavin:
> Last nit that maybe some `log.info` should be `log.debug`. Otherwise, LGTM!

I did another round of Claude review. Found at least one actual bug in the python lib's `JSESSIONID` refresh; I'll push a fix for that soon. And it seems to really not like sending the JWT `X-Auth-Request-Access-Token` header to all services by default. I might implement some other middleware which will pass that header, and we can set it explicitly on only those services that need the token. I'm still investigating.

kathrynalpert:
> it seems to really not like sending the JWT `X-Auth-Request-Access-Token` header to all services by default. I might implement some other middleware which will pass that header, and we can set it explicitly on only those services that need the token. I'm still investigating.

Ok, interesting. It keeps telling me we should set `fullScopeAllowed=false`.

johnflavin:
After some refactors and clarifications, I don't feel comfortable with this plugin's auth flow / architecture
- The `xnat` keycloak client is `bearerOnly`, it doen't need a secret, it never gets its own tokens. XNAT is the only KC client in Scout that works this way, all the others get real KC clients which the services use to mint their own real tokens. The way this plugin's auth works goes against ADR-0003, which lays out this auth architecture.
- The only way the OA2P token has the `xnat-access` role anyway is because every token carries every client role. That in itself is a code smell. We should be setting `fullScopeAllowed=false` so each client's tokens are scoped to only that client's roles.
- We are configuring our oauth2-proxy middleware to forward the oauth2-proxy JWT to all services in the `X-Auth-Request-Access-Token` header. Claude really doesn't like this. Services that aren't expecting this header to contain a token might log it or leak it somehow. Token leakage could be bad: anyone who gets this token now gets to act as that user for the lifetime of the token. At the very least we should be forwarding the token only in a custom middleware that we can opt services into rather than the middleware that is default on all services.

The other pieces of this POC—the bearer auth login path and the `xnatpy` wrapper library—are potentially still useful. But we don't need to merge them right now. We still have lots of questions about how XNAT and Scout will integrate, so these plugins/libs may need to change before they actually get used. I can come back to the branch later and get them into shape when the rest of the XNAT integration is in a place where we need to actually get these pieces deployed.

PR #411, opened 2026-06-01
Trino AuthZ: Voila impersonation + scout-notebook consolidation

kathrynalpert:
### Product
N/A - Fourth slice of the Trino per-user AuthZ feature: **Voila joins the JWT-impersonation path.** Voila notebooks now query Trino over HTTPS as the `voila_svc` service principal (Keycloak realm from #406), impersonating the logged-in user via `X-Trino-User`. Same pattern Superset adopted in PR #409.

Importantly, this PR removes the old `pyspark-notebook`: JupyterHub + Voila share a single Trino-only `scout-notebook` image, which hopefully addresses the disk pressure we've been seeing in CI.

### Technical
- **Voila → Trino over JWT, impersonating the end user** — same pattern Superset adopted in #409. `scout_trino.connect()` mints a `voila_svc` token and sets `X-Trino-User`; `voila_runtime` carries the user's identity from oauth2-proxy's `X-Auth-Request-Preferred-Username` header into the kernel (just the username not the full raw token). `connect_rw()` covers reviewer-annotation writes against the unauthenticated trino-rw instance (code we control, not arbitrary writes).
- **The Voila chart now owns its ConfigMap / Secret / NetworkPolicy** instead of the role creating them imperatively — GitOps-ready.
- **Bundled dashboards migrated** (cohort, rads, quality-metrics, followup-detection) from their own unauthenticated connections to `scout_trino`, so each query runs as the logged-in user under OPA. followup's annotation write moves off Spark to trino-rw.
- **One shared `scout-notebook` image** for JupyterHub + Voila (scipy-notebook + trino-python-client + voila; drops `s3fs`/`boto3`), replacing the old `pyspark-notebook`.
- **oauth2-proxy** emits the `X-Auth-Request-*` headers and the forwardAuth middleware forwards the username to upstreams. Jupyter just moves to the shared image variable here; its JWT migration is #412.
- **CI smoke-test split** into `analytics` + `notebooks` halves so the full stack stays under the runner's disk-pressure threshold, plus a new `voila-ingress` NetworkPolicy group and unit tests for the two helpers (details under Testing).

## Type of change
- [x] New feature
- [x] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
Voila adds a second JWT-impersonation client to the AuthZ pipeline (Superset was first in PR #409). OPA's impersonation rule permits `voila_svc → any user`; actual data access is governed by the impersonated user's Keycloak attributes (#406).

##### Appsec
- `voila_svc` token cached process-locally per Voila worker; refresh at ~80% lifetime. Cache never on disk, dies with the worker. Same lifetime properties as Superset's.
- Only the end-user's **username** (`X-Auth-Request-Preferred-Username`) is threaded into the kernel — not the raw access token.
- Voila pods now verify Trino's TLS via the cert-manager-issued `trino-tls` CA mounted from a Secret.
- The shared scout-notebook image drops `s3fs` and `boto3`. Notebook pods can no longer reach MinIO directly even if a user obtains lake creds; all reads go through Trino + OPA enforcement.
- `voila-ingress` NetworkPolicy restricts Voila:8866 to Traefik only. Without it, an in-cluster pod could connect directly and forge `X-Auth-Request-Preferred-Username`. NetworkPolicy is defence-in-depth (label-based, not cryptographic).
- The followup-detection write interpolates a table *identifier* (can't be a bound parameter); it's validated against a strict dotted-identifier regex and row values are bound with `?`. The unavoidable formatted-SQL semgrep warnings carry a justified `# nosemgrep`.

### Performance
- Voila image becomes the shared `scout-notebook` (scipy-notebook + trino-python-client + voila). One build target instead of two. Single image cached on each node — faster cold spawn.
- Smoke-test disk pressure reduced (~5GB saved by dropping pyspark-notebook's build + k3s image import; further reduced by the matrix split spreading playbook deploys across two runners).

### Data
N/A

### Backward compatibility
- The four bundled Voila dashboards are migrated to `scout_trino.connect()` in this PR. **User-authored notebooks that imported `boto3` / `s3fs` for direct lake access will still break on the new image** 
- `jupyter_singleuser_image_*` → `scout_notebook_image_*` inventory rename. Anyone with a pinned override in `inventory.yaml` needs to update the variable name; same image-name value.

## Testing
- `helm template helm/voila` renders cleanly with the new ConfigMap / Secret / NetworkPolicy templates + checksum annotations.
- CI smoke (both sides) confirms pods roll out: Voila + Jupyter on the notebooks side; Superset + the data-authorization assertion suite on the analytics side.
- NetworkPolicy tests (see Technical → CI smoke split) cover the new `voila-ingress` group on the notebooks side and the existing `trino-rw` + `opa-trino` groups on both sides.
- Unit tests (`helm/voila/tests/`, run in the `ansible-and-python-tests` job): `scout_trino` (X-Trino-User from the forwarded username, voila_svc JWT mint + caching, anonymous fail-safe, trino-rw write path) and `voila_runtime` (header → kernel-env threading, per-request identity isolation, fallback warning). 15 tests; third-party imports stubbed so the suite needs only pytest.
- End-to-end Voila → Trino → OPA → row-filter is still not asserted *in-cluster*. The Voila helper logic is now unit-tested (above), and the Trino + OPA + row-filter mechanics are exercised by the analytics-side data-authorization suite (curl + `X-Trino-User`). What's left uncovered is only the live HTTP→kernel wiring through Voila's ingress — see "Note for reviewers".

## Note for reviewers
**Jupyter still on HTTP** — the image + naming in this PR set up the singleuser pod for the JWT migration; the kernel-side spawn-time JWT propagation lands in #412.

**MCP still on HTTP** — same story; MCP migrates in its own PR alongside the bespoke-tool replacement and introduces the PASSWORD authenticator at that point.

**Voila identity path — in-cluster E2E is a follow-up** — `voila_runtime` (header → kernel-env threading) and `scout_trino` (env → `X-Trino-User`) are now unit-tested, but the live path (oauth2-proxy → Traefik → Voila render → kernel) isn't asserted in the smoke-test. A full in-cluster E2E would need to re-seed `test_reports` on the notebooks side (the PySpark hl7-transformer image — fights the disk-pressure reason the smoke test was split) and parse Voila's rendered HTML; deferred as not worth that cost given the unit + data-authorization coverage.

Design rationale and the broader picture are in ADRs 0020–0023, see #402.

## Checklist


.github/scripts/update-versions.sh
johnflavin: This will need to be renamed / removed as well. It would break the script and the Release workflow.

docs/internal/versions-and-releases.md
johnflavin: Nit: Should update this reference to the old name

analytics/notebooks/followup-detection/followup_review_dashboard.py
```diff
@@ -1554,76 +1547,63 @@ def on_export(b):
...
+                for row in ann_data:
+                    cur.execute(
+                        update_sql,
+                        (
+                            row["human_ground_truth"],
+                            row["human_notes"],
+                            row["accession_number"],
+                        ),
+                    )
+                    cur.fetchall()  # Trino DB-API requires a fetch to run DML
```
johnflavin: (This isn't super important since the follow-up detection notebook is mostly for demo purposes, but I think it's worth flagging so we think about any future stuff written in this style.) 

The old spark code wrote to the delta lake as a single atomic merge. This new trino code writes a bunch of small updates in a loop. That means a couple things:
1. Some rows could write successfully but if a later one fails the operation will report failure even though the early rows aren't rolled back.
2. This will write a lot of small updates to the delta log. We would want to consider running periodic cleanup jobs (I don't remember exactly what the delta lake operation is called but I know one exists for this).

.github/scripts/update-versions.sh
kathrynalpert: Good catch, thank you!

analytics/notebooks/followup-detection/followup_review_dashboard.py
kathrynalpert: Updated to be a merge

johnflavin (approved):

PR #412, opened 2026-06-02
Trino AuthZ: Jupyter passthrough

kathrynalpert:
### Product
Fifth slice of the Trino per-user AuthZ feature: **JupyterHub joins the per-user identity path.** Notebook kernels query Trino as the logged-in user over HTTPS, sending the user's own Keycloak access token as the Bearer — no service principal, no impersonation. Trino reads the user from the JWT.

### Technical
- **Jupyter role** — override the built-in `user`/`server` roles with `admin:auth_state!user` so a kernel can read its own auth_state (the Keycloak token) via the Hub API; `refresh_pre_spawn: true`; singleuser env flipped to HTTPS/8443 + `TRINO_CA_CERT`; `TRINO_USER` dropped; NetworkPolicy egress → `trino_https_port`.
- **New `scout` Python SDK (`sdk/python/`)** — read-only, pip-installed into the `scout-notebook` image. Top-level surface: `scout.query()` / `scout.connect()` / `scout.current_user()`. `_identity` auto-detects the mode from env (Voila svc-JWT + `X-Trino-User` impersonation vs. Jupyter user-token pass-through).
- **Token refresh (ADR 0024).** The SDK keeps the short-lived bearer valid with no OAuth in user code: proactive re-fetch within 60s of expiry over a per-request dynamic bearer (one long-lived engine, no `dispose()`), plus a reactive one-shot retry on a Trino 401. On Jupyter the Hub is the custodian, so an oauthenticator `refresh_user_hook` rotates the access token at half-life (stock oauthenticator only re-validates a live token); `auth_refresh_age` and the realm's `accessTokenLifespan` both derive from one `keycloak_access_token_lifespan` var so they can't drift.
- **Dockerfile** installs the SDK after the mamba layer; build context is the repo root, so `docker-build-cache` gained an optional `context` input.
- **Voila chart absorbs the SDK** — `scout_trino.py` is deleted, code updated to use SDK.
- **Notebooks + playbooks** migrate to `import scout; scout.query()/connect()`; followup-detection's writeback uses `from playbook_helpers import connect_writeback`.

Design rationale and the broader picture are in ADRs 0020–0024, see #402.

## Type of change
- [x] New feature
- [x] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
Jupyter joins the JWT path (#409, #411). The kernel sends the user's own access token (issued by the `jupyterhub` client with the `trino-audience` scope, so `aud=trino`); Trino reads the principal from `jwt.principal-field=preferred_username` and OPA enforces that user's attributes.

##### Appsec
- **Passthrough, not impersonation.** A dynamic-auth callable like the SDK's only works with a source of fresh *user* tokens behind it, and Jupyter uniquely has one: the Hub holds `auth_state` (access + refresh) and refreshes server-side, so the kernel re-fetches on demand. The impersonation clients don't: Superset's `DB_CONNECTION_MUTATOR` receives only a username and queries often run outside any user session (Celery workers, the dashboard-import Job); Voila gets a request-time token snapshot with no refresh channel; Trino-MCP can't send JWTs downstream (ADR 0022.) Moreover, **the kernel is an arbitrary-code surface.** A svc credential is an impersonate-*anyone* capability, so it can't share a trust domain with user code — and hosting it server-side would mean adding a trusted query proxy that attributes kernel traffic to users, re-deriving with weaker primitives (and one more impersonation-capable credential) the identity the user's JWT already proves to Trino cryptographically.
- No token exchange: see discussion on #410 for more on this.
- Bearer cached per-kernel, process-local, never on disk; re-fetched near expiry. The long-lived refresh credential never reaches the kernel (Hub-side for Jupyter; client secret in a Secret for Voila).
- Singleuser pods verify Trino's cert via the mounted CA.

### Performance
The SDK reuses one SQLAlchemy engine per kernel; its auth layer fetches the current token on every request, so engines keep working across token expiry instead of needing `engine.dispose()` + reconnect. The reactive 401 retry covers the residual (clock skew, in-flight expiry), so rotation is invisible to queries.

### Data
N/A

### Backward compatibility
- Saved notebooks that hand-roll `create_engine("trino://$TRINO_USER@...")` break (`TRINO_USER` gone, host is HTTPS/8443 + JWT). Migration: `import scout; scout.query(...)`.

## Testing
- Unit (CI): `sdk/python/tests/` (18 — both providers + the 401-retry path) + `helm/voila/tests/` (9).
- Contract (CI): `sdk/python/tests_contract/` (3, against the real trino-python-client) pin the `user=None` → no `X-Trino-User` behavior the pass-through model relies on, so a client upgrade that regressed it fails loudly.
- Token refresh has no CI e2e; verified manually on dev04 by querying across the access-token expiry boundary (token rotates, no 401s).
- Not in CI: end-to-end Hub→Trino→OPA→row-filter on the Jupyter side (the #409 suite only drives Superset).
- Manual testing on `dev04` including running token refresh validation detailed in [this comment](https://github.com/washu-tag/scout/pull/412#discussion_r3356861143)

## Note for reviewers
Distribution name is `scout`; it can switch (e.g. `washu-tag-scout`) if a public PyPI publish ever needs it, keeping the import path. No PyPI publish here — the image is the only consumer.

Conversation in #410 is relevant to this PR, though we decided to hold off on merging that.

## Checklist


sdk/python/src/scout/_query.py
```diff
@@ -0,0 +1,152 @@
...
+    cached engine.
+
+    Whether `user` is set is the entire identity distinction:
+      * Voila    -> provider returns the impersonated username; we set it and
+                    the trino client sends it as X-Trino-User.
+      * Jupyter  -> provider returns None; we omit `user`. The trino client
+                    then emits no X-Trino-User (requests drops the None-valued
+                    header on its Session merge), so Trino derives the session
+                    user from the JWT principal (jwt.principal-field=
+                    preferred_username) instead.
```
johnflavin, line 100: This was flagged by a Claude review as something that isn't tested against a real trino (only a stub) and is important to verify. Are we sure that trino doesn't set a default `X-Trino-User` header even if we pass `user=None`? 

It claims
> The trino client has historically defaulted the session user to `getpass.getuser()` (→ `jovyan` in this image) when not supplied. If that happens with `trino>=0.330`, Trino sees JWT `principal=alice` + `X-Trino-User: jovyan` → impersonation/denied, not pass-through.

But I haven't verified this.

sdk/python/src/scout/_identity.py
```diff
@@ -0,0 +1,173 @@
...
+        return self._token, user
+
+
+class _JupyterHubUserProvider:
+    """Fetches the user's Keycloak access token via the Hub API.
+
+    Cached in-process; re-fetched when near expiry. The Hub itself
+    refreshes the token against Keycloak when refresh_pre_spawn is on
+    plus the OAuthenticator's refresh path, so calling GET /users/<user>
+    returns a fresh access_token.
```
johnflavin: Kind of a Nit: I had more or less the same code in my `xnatpy` wrapper. The claim is that it always keeps the token fresh, but there is a failure scenario: if the hub didn't refresh the user's token, when we in the notebook go to refresh with the hub it could give us that stale / just about to be stale `access_token`.

I don't think this is something we need to fix exactly, we can't control when the hub refreshes tokens. But it would be good to make the documentation / docstrings clear that the token may expire, and would be good to catch that error case and say something along the lines of "you may need to refresh your jupyterhub connection".

helm/voila/files/voila.py
```diff
@@ -2,18 +2,19 @@
 
 Loaded automatically by Voila from /etc/jupyter/voila.py. Voila's
 JupyterApp initialization calls `load_config_file("voila", ...)`, which
-expects the filename pattern `voila.py` (basefilename + ".py") — not
+expects the filename pattern `voila.py` (basefilename + ".py") - not
```
johnflavin: Nit: A lot of weird `—` -> `-` churn in this and `voila_runtime.py`. Doesn't seem necessary.

johnflavin (approved):

sdk/python/src/scout/_identity.py
kathrynalpert: This ended up taking some significant work, see ADR 0024, `sdk/python/src/scout/_query.py`, `sdk/python/src/scout/_identity.py` and `ansible/roles/jupyter/templates/values.yaml.j2` (specifically `eagerTokenRefresh`).

sdk/python/src/scout/_query.py
kathrynalpert: Verified that existing behavior is correct via manual testing + Claude's code investigation: the client doesn't set a default `X-Trino-User` when `user=None`, which means Trino uses the JWT principal as we expect. 

Since that depends on two library behaviors we don't own (and the unit suite stubs them), Claude added a contract test against the real trino/requests: `sdk/python/tests_contract/test_trino_client_contract.py`, so a client upgrade that regresses it fails loudly.

sdk/python/src/scout/_identity.py
kathrynalpert: BTW/FFR, I tested this with
```
import os, time, json, base64, requests
import scout

def hub_token():
  api = os.environ["JUPYTERHUB_API_URL"].rstrip("/")
  r = requests.get(f"{api}/users/{os.environ['JUPYTERHUB_USER']}",
                   headers={"Authorization": f"token {os.environ['JUPYTERHUB_API_TOKEN']}"}, timeout=10)
  r.raise_for_status()
  return (r.json().get("auth_state") or {})["access_token"]

def remaining(tok):
  p = tok.split(".")[1]; p += "=" * (-len(p) % 4)
  return json.loads(base64.urlsafe_b64decode(p))["exp"] - time.time()

seen, t0 = set(), time.time()
while time.time() - t0 < 8 * 60:
  tok = hub_token(); seen.add(tok)
  try:
      scout.query("SELECT 1 AS ok"); q = "OK"
  except Exception as e:
      q = f"FAIL ({type(e).__name__})"
  print(f"t+{time.time()-t0:4.0f}s  expires in {remaining(tok):5.0f}s  "
        f"tokens seen: {len(seen)}  query: {q}", flush=True)
  time.sleep(20)
```

docs/internal/adr/0024-sdk-trino-token-refresh.md
```diff
@@ -0,0 +1,170 @@
...
+  token.
+- The layer-1 Hub hook is a workaround for stock-oauthenticator behavior. It is
+  version-gated and deletable if/when upstream makes refresh proactive; an
+  upstream issue with the dev04 trace should be filed.
+- `connect()` callers on the rare in-flight-expiry path re-run the failing
+  statement themselves; documented in its docstring.
+
+## Implementation Notes
+
+- SDK: `sdk/python/src/scout/_query.py` (`_is_auth_error`, `_with_auth_retry`,
```
johnflavin: Nit: `_is_auth_error` got renamed to `_is_unauthorized` when you split off 401 and 403 handling

johnflavin (approved):
The token refresh stuff looks like it took a lot of work and testing, but I think it makes this SDK and notebook querying more lovable. Nice work!

PR #419, opened 2026-06-04
Automate Cert Renewal

andylassiter:
### Product
No user facing changes.

### Technical
- New `cert_manager` Ansible role (looked at Kate's trino-rbac branch to minimize future merge conflicts).
- Adds a `tls_mode` toggle so clusters can opt into cert-manager-based ACME cert issuance (Let's Encrypt + acme-dns DNS-01) instead of manual file-based certs. Uses cert manager default rotation policy renewing them when 2/3 of their lifetime has passed.
- `tls_mode: file` is the default; no behavior change unless inventory explicitly flips it.
- `scout_common/tasks/cert_manager_acme_issuer_setup.yaml` applies the ACME ClusterIssuer, Certificate, and acme-dns credentials Secret.
- Adds Prometheus scrape config for cert-manager and Grafana dashboard and alerts (CertNotReady, CertExpirySoon, CertRateLimit). 

## Type of change
- [x] New feature

##### Authorization
N/A

##### Appsec
`acmedns_credentials` is vault-encrypted and stored as opaque Secret with `no_log: true`. cert-manager auto-registers a per-cluster RSA-keyed LE account, so compromise blast radius is contained to one cluster.

### Performance
N/A

### Data
N/A.

### Backward compatibility
`tls_mode` defaults to `file`, so existing clusters should keep working without inventory changes. 

## Testing
Deployed end-to-end on dev02. Tested with Let'sEncrypt test environment first, then flipped to LE prod and real browser-trusted cert was issued. Verified Prometheus scrapes cert-manager pods and all 3 alerts deploy. `file` mode still works end-to-end.

## Note for reviewers
I think this should work air-gapped but not 100% certain until deployed on preprod. Can fall back to `file` managed cert if we need to fix.

## Checklist

Screenshot 2026-06-04 at 4 35 15 PM

Screenshot 2026-06-04 at 4 34 03 PM


ansible/roles/traefik/tasks/main.yaml
```diff
@@ -2,39 +2,53 @@
...
-- name: Read TLS cert from remote
-  no_log: true
-  ansible.builtin.slurp:
-    src: '{{ tls_cert_path }}'
-  register: cert_file
+- name: Bootstrap ACME ClusterIssuer + wildcard Certificate
+  ansible.builtin.include_role:
+    name: scout_common
+    tasks_from: cert_manager_acme_issuer_setup
+  when: tls_mode == 'cert_manager'
```
johnflavin, line 20: Before doing either of these conditionals on `tls_mode` it would be good to assert it is either `"file"` or `"cert_manager"`, otherwise a typo will break TLS.

ansible/roles/scout_common/tasks/cert_manager_acme_issuer_setup.yaml
```diff
@@ -0,0 +1,66 @@
...
+      kind: Certificate
+      metadata:
+        name: '{{ cert_manager_acme_certificate_name }}'
+        namespace: '{{ traefik_namespace }}'
+      spec:
+        secretName: '{{ traefik_tls_secret_name }}'
+        issuerRef:
+          name: '{{ cert_manager_acme_issuer_name }}'
+          kind: ClusterIssuer
+        dnsNames: '{{ cert_manager_acme_dns_names }}'
```
johnflavin, line 66: As of now I think this will create the `Certificate` and move on to the rest of the install, while that `Certificate` could take a while to actually get the cert, and that process might fail. It might make sense to add a `wait` after this task on some sort of readiness on this `Certificate`. (Though that might add a lot of waiting time and delay the rest of the install for nothing so I'm not 100% sure it is a good idea.)

ansible/roles/grafana/templates/alerts/cert-manager-cert-expiry-soon-alert.json.j2
```diff
@@ -0,0 +1,89 @@
...
+            {
+              "refId": "A",
+              "relativeTimeRange": {
+                "from": 600,
+                "to": 0
+              },
+              "datasourceUid": "prometheus_datasource_01",
+              "model": {
+                "editorMode": "code",
+                "expr": "avg by (name, namespace) (\n  (certmanager_certificate_expiration_timestamp_seconds - time())\n  /\n  (certmanager_certificate_expiration_timestamp_seconds - certmanager_certificate_renewal_timestamp_seconds)\n)",
```
johnflavin, line 24: I believe the `cass-operator` certs are created by `cert-manager` but are self-signed and don't expire. Might want to filter them out so we don't get spurious alerts on them.

ansible/roles/grafana/templates/alerts/cert-manager-cert-expiry-soon-alert.json.j2
johnflavin, line 1: Would this alert run even in file mode? And would it actually work? On the latter, I don't think it would work properly; the files aren't managed by cert manager so wouldn't have any metrics to scrape. But I do think it would run, it would get `NoData`, and would alert.

Should only create this alert when `tls_mode == 'cert_manager'`.

ansible/roles/traefik/defaults/main.yaml
johnflavin: This should be removed, it is now defined in `scout_common`

johnflavin (commented):
A couple more things that don't have a good line to comment on:
1. The staging playbook is, I think, currently broken in this PR. Staging uses the traefik role, which will currently try to use either the `cert_manager` or `file` TLS modes, but the staging cert configuration is quite different. I don't think the LE configuration will work for staging since its DNS is not public, and the file mode won't work because it doesn't have cert files. Currently we create a self-signed certificate by hand. It might be nice to migrate that to cert-manager, but that isn't currently happening in this PR.
2. Nit: The `cert_manager` role should depend on `scout_common`. This is only a best practices / nice-to-have, since right now the `cert_manager` role doesn't get invoked directly, but only through other roles that _do_ depend on `scout_common`. But if we modify things in the future so we do invoke it standalone this could be a confusing error to debug. 
3. Is there a reason to add a third mode, where we use static cert files to bootstrap the install (bypassing LE cert generation at install time) but use the real renewal process? I'm just thinking that might help with cert churn installing to a dev cluster. We'd just use the filesystem as a cert cache. And in that case it might also be nice to dump the valid certs back to the file system occasionally. But this is all a bit speculative, I don't know that we need any of this.

ansible/roles/scout_common/tasks/cert_manager_acme_issuer_setup.yaml
andylassiter: I think we need a `wait` for auth to work. They kecloak oidc configuration endpoint would not be reachable until the cert is ready. It took 2-3 minutes on my dev stack for the cert to be issued.

ansible/roles/grafana/templates/alerts/cert-manager-cert-expiry-soon-alert.json.j2
andylassiter: The `cass-operator` certs are created by `cert-manager`, are self-signed, but they _do_ expire. They're just auto-renewed by cert-manager. If their renewal silently fails we would want to know but I do think the odds of this are low.

ansible/roles/grafana/templates/alerts/cert-manager-cert-expiry-soon-alert.json.j2
andylassiter: Yes and no on the latter. But it would catch the cass-operator certs, so it will not get `NoData`. If we filter out the cass-operator certs then we would only want to alert when `tls_mode == 'cert_manager'`.

andylassiter:
1. Will look into the staging node.
2. Can do.
3. I was considering something like this too but came to the same conclusion of "But this is all a bit speculative, I don't know that we need any of this."

ansible/roles/grafana/templates/dashboards/cert-manager-dashboard.json.j2
```diff
@@ -0,0 +1,637 @@
...
+        "options": [],
+        "query": "prometheus",
+        "refresh": 1,
+        "regex": "",
+        "type": "datasource"
+      }
+    ]
+  },
+  "time": {
+    "from": "now-24h",
```
johnflavin: I noticed in your demo the time window on the dashboard seemed a bit short. Maybe showing something on the order of the last few days or last week would be a good default.

andylassiter: Good catch, can do.

andylassiter:
Confirmed staging playbook still works.

johnflavin (approved):
You'll need to solve merge conflict, but other than that LGTM

PR #417, opened 2026-06-04
Trino AuthZ: Test hardening and docs

kathrynalpert:
### Product
N/A — no new runtime behavior. Adds user- and operator-facing Data Authorization docs for the shipped AuthZ feature, and hardens test coverage across the AuthZ slices (#402, #406, #409, #411, #412, #414).

### Technical

**Documentation**
- `docs/source/data_authorization.md` — user-facing page: what row filtering and PHI masking mean for a notebook or Superset user; links to the operator guide.
- `docs/source/technical/data_authorization.md` (new) — operator guide: the shipped attributes (`allowed_facilities`, `mask_phi_fields`, `bypass_hidden_tables`), an attribute-setting walkthrough, how propagation works (Keycloak → MinIO bundle → OPA), verifying a user's effective view via `decision_context`, troubleshooting, view-only tables, adding a new restriction dimension, and prefix-based table filtering.
- Notebook docs de-Sparked — examples use the `scout` SDK (Trino-backed) instead of PySpark, matching the Spark-less `scout-notebook` image (#411, #412). Touch-ups to `authentication.md`, `dataschema.md`, `index.md`, `technical/index.md`, and `CLAUDE.md`.

**Test-coverage hardening** — every new test is mutation-tested: break the corresponding production behavior, confirm the matching test (and only it) goes red. Candidates that caught nothing unique were dropped rather than kept as bloat.
- **data-authorization e2e** (`tests/data-authorization/authz-tests/run.sh`, #409 / #414) — the three near-identical `nextUri`-drain loops collapse into one auth-parameterized helper (`_trino_run`); the Bearer-JWT and HTTP-Basic/`PASSWORD` paths become thin wrappers (`trino_query_expect_deny` stays separate — it inverts the error semantics and needs the raw HTTP status). The type-aware PHI mask is now asserted across all three masked columns (`patient_name` / `zip_or_postal_code` → `'[REDACTED]'`, `full_patient_name` → `NULL`), plus a mask-off counter-assertion (`mask_phi_fields=["false"]` returns the real values) that proves masking is *conditional* rather than unconditional.
[...]

PR #422, opened 2026-06-08
User administration console: approvals, data-access attributes, admin lifecycle

kathrynalpert:
### Product

Adds an **in-app user-administration console**, where a `scout-admin` approves pending users, edits their data-access attributes (the `allowed_facilities` / `mask_phi_fields` / … dimensions from the Trino AuthZ feature), promotes/demotes other admins, and offboards users (as promised in https://github.com/washu-tag/scout/pull/417#discussion_r3358105332). An in-console "Open in Keycloak" link allows `scout-admins` to access the old interface, if desired.

Screenshot 2026-06-08 at 12 33 25 PM
Screenshot 2026-06-08 at 12 33 37 PM
Screenshot 2026-06-08 at 12 57 01 PM

### Technical

- **`scout-users` Keycloak SPI** — a `scout-admin`-gated REST resource. The attribute schema is discovered from the User Profile (same `trino_attribute_filters` inventory setting).
- **Launchpad console** at `/admin/users` — a Pending/Active/Admins table with search and a dual-mode editor drawer (approve a pending user, or manage an existing one), gated on `session.user.isAdmin`.
- **Token pass-through** — `/api/users/*` forwards the admin's own JWT to the SPI (same-realm, the ADR 0022 model); the session cookie holds only the refresh token and the proxy mints access tokens server-side, with reactive re-login on a dead-session 401.
- **Admin-event propagation** — every SPI mutation emits a Keycloak `AdminEvent` so OPA (#406) and the approval/offboard emails react; without it an approved user gets zero rows until the bundle re-seeds.
- **Server-side guardrails** (mutation-tested) — no self-offboard, attribute-key allowlist, and SPI access restricted to tokens audienced `scout-users-api` (defense in depth, since this API can grant realm-admin).
- Full upgrade to **Keycloak 26.6.3** #420 only updated the gradle dependency for our custom SPIs
- `fullScopeAllowed=false` on the launchpad client

Full rationale for SPI design, the admin-event fix, the token model, the console-vs-Keycloak boundary is in **ADR 0025**.

## Type of change
- [x] New feature
- [x] Improvement
- [ ] Bug fix
- [x] Documentation update
- [ ] Test update

##### Authorization
- Every endpoint is `scout-admin`-gated by a live group-membership check against the Keycloak user model (instead of a token claim), so a revoked admin loses access on the next call. A fresh realm is console-locked until the first `scout-admin` is granted out-of-band in the Keycloak master console (same process as we use today).
- The SPI accepts only tokens audienced for it (`aud: scout-users-api`, added by the launchpad client's mapper). This blocks cross-resource replay: a `scout-admin`'s token minted for some other resource (e.g. an `aud: trino` notebook token) can't be pointed at user administration. 
- The launchpad client is `fullScopeAllowed=false`, so its tokens carry only in-scope roles (launchpad-admin/launchpad-user), not the user's realm roles. This stops a stolen token from being a turnkey realm-admin credential and forces any escalation through the audited `approve`/promote path, but it is hardening on the **leaked-credential** axis, not a barrier to escalation or data access once an attacker controls an account.

##### Appsec
- The browser no longer holds the Keycloak access token. The refresh token continues to live in a JWE-encrypted `httpOnly`/`secure` cookie; the `/api/users` proxy mints a short-lived access token server-side per request and never returns or logs it. 
    - Storing the access token in the cookie pushed it past the browser/proxy size limit and intermittently dropped the session, so this wasn't primarily a change for improved security posture.
- `GET /users` returns each user's `scoutAuthz` attributes only, never the raw attribute map, so `scout_terms_accepted_at` and the approval-email marker don't leak to the client.

### Performance
Sized for small realms. Search is client-side over the loaded list. The last-admin guard counts members via an indexed `getGroupMembersStream` capped at 2, not a full group scan. A TOCTOU race on two concurrent last-admin removals is accepted (recovery is the Keycloak first-admin bootstrap).

### Data
- The editor and the table summary call out "No access — 0 rows" / "no facilities" to help users understand how to use the attribute selection
Screenshot 2026-06-09 at 6 57 22 AM

- "Pending" users are those who have accepted Terms-of-Use but are not yet in `scout-user`. Users that haven't accepted terms aren't shown.

### Backward compatibility
- next-auth's session cookie schema changed (now holds only the refresh token, previously access + refresh + groups); cookies may need to be cleared

## Testing
- **Keycloak SPI suite** (`./gradlew test` in `keycloak/event-listener/`, JUnit + Mockito) added to CI
- **Manual on dev04**: approve → user lands in `scout-user` and OPA row filters apply (the AdminEvent → bundle path from #406), edit attributes, promote/demote/offboard, search, and approve-auto-advance through the pending queue. 
- **Deferred**: no in-cluster e2e for the console itself. The SPI logic is unit-tested, and the propagation path it depends on is the AdminEvent → OPA-bundle mechanism already exercised by the `data-authorization` suite from #406/#418.

## Note for reviewers
- Curious if we like the "guardrails" about removing one's own access (ditched the "last admin" protection)
- ADR 0022's Voila section is corrected to match reality (username-only header), as it was confusing the 🤖.

### Future work not addressed here
- Move next-auth to a **server-side session store** — opaque session id in the browser, no token material, plus server-side revocation — which hardens the launchpad's pre-existing session model (it predates this work)
- Add a new user-management group that doesn't admin other Scout services 

## Checklist


docs/internal/adr/0025-launchpad-user-approval.md
```diff
@@ -0,0 +1,248 @@
...
+
+A purpose-built **user-administration console in the launchpad**, backed by a
+small **Keycloak REST resource**, with **dynamic, config-driven attributes**.
+
+### 1. scout-users REST resource (Keycloak SPI)
+
+A `RealmResourceProvider` in the existing event-listener SPI exposes, under
+`/realms/<realm>/scout-users/`, a `scout-admin`-gated API (`BearerTokenAuthenticator`
++ a live `scout-admin` group check on every call; no standing admin-API
+credential — the resource acts on the authenticated admin's own authority). As
```
johnflavin: Nit: the formatting on these lines got weird. Splitting the line from 35 to 36 and starting with a `+` makes the renderer think it is a bulleted list, which it is not. So when rendered it looks like the parenthesis on 35 is open with no close, but really it closes on 37. Plus this whole parenthetical is hard to follow through all the details, hyphens, semicolons.

docs/internal/adr/0025-launchpad-user-approval.md
```diff
@@ -0,0 +1,248 @@
...
+`bypass_hidden_tables=false`).
+
+**Guardrails** (package-private predicates, mutation-tested, enforced server-side
+regardless of the UI):
+
+- *Last-admin lockout* — demoting or offboarding the last `scout-admin` is
+  rejected with 409 (members counted via an indexed `getGroupMembersStream`
+  capped at 2, not a full scan). A check-then-act (TOCTOU) race on two concurrent
+  removals is accepted for these small realms; recovery is the Keycloak bootstrap
+  below.
```
johnflavin: I'm not sure we need this. We have a keycloak admin account that we need to use in order to get any users into the system at all. Wouldn't that still be there as a backstop if all `scout-admin` users got removed?

docs/internal/adr/0025-launchpad-user-approval.md
```diff
@@ -0,0 +1,248 @@
...
+  users, rewrite anyone's data-access attributes, promote/demote — no escalation
+  needed.
+- *Self-service is gated only by account control* — if the thief controls an
+  account they can authenticate as (their own, or a pending identity brokered in
+  via the IdP), one `approve` with `allowed_facilities=*` / `mask_phi_fields=false`
+  grants *themselves* full unmasked data access, and a self-promote re-mints a
+  `realm-admin` token from another full-scope client (e.g. the built-in
+  `admin-cli`). The SPI cannot create users or set passwords and self-registration
+  is off, so a thief with **no** realm account is the only one confined to griefing
+  others' access.
```
johnflavin: > a thief with **no** realm account is the only one confined to griefing others' access.

Nit: Odd to see "griefing" as a verb outside of a gaming context. Not necessarily wrong, though I'm not sure I've ever seen it used transitively / with an object, so that may be incorrect usage. And I don't think "is the only one" makes sense here.

keycloak/event-listener/src/main/resources/theme/scout/email/html/email-admin-approval.ftl
```diff
@@ -1,4 +1,9 @@
 <#import "template.ftl" as layout>
 <@layout.emailLayout>
-<p>User <strong>${username}</strong> has requested access to Scout. Please <a href="${scoutUrl}">log in to Scout</a> to enable their account.</p>
+<p>User <strong>${username}</strong> has requested access to Scout<#if scoutSite??> (<strong>${scoutSite}</strong>)</#if>.</p>
+<#if approvalUrl??>
```
johnflavin: I don't think `approvalUrl` can be empty / null, can it? The `UserApprovalEmailEventListenerProvider` defines it as
```java
static String approvalUrl(String scoutBaseUrl, String userId) {
    return scoutBaseUrl.replaceFirst("/+$", "") + "/admin/users?user=" + userId;
}
```
It doesn't look like it can ever be null or empty.

keycloak/event-listener/src/main/resources/theme/scout/email/text/email-admin-approval.ftl
```diff
@@ -1,2 +1,6 @@
 <#ftl output_format="plainText">
-User ${username} has requested access to Scout. Please log in to Scout ( ${scoutUrl} ) to enable their account.
\ No newline at end of file
+User ${username} has requested access to Scout<#if scoutSite??> (${scoutSite})</#if>.
+
+<#if approvalUrl??>Review and approve their request:
```
johnflavin: Same as above: I don't think `approvalUrl` can ever be null/empty.

johnflavin (approved):
LGTM

docs/internal/adr/0025-launchpad-user-approval.md
kathrynalpert: Yep, this was one of my "note for reviewers" questions (Curious if we like the "guardrails" about removing our own access / keeping at least one scout-admin). I've removed it, as I don't think it's useful and it adds complexity and a race.

docs/internal/adr/0025-launchpad-user-approval.md
kathrynalpert: 100%, I missed this.

PR #423, opened 2026-06-08
Add Patients dashboard

kathrynalpert:
### Product
Adds a new **Patients** dashboard to Analytics that reframes the data lake around *patients* instead of *exams*. Every count is at the reconciled-patient level `scout_patient_id`, so a patient with five scans counts once.

Screenshot 2026-06-08 at 11 08 03 AM

### Technical
New `patients` bundle under `helm/scout-dashboards/files/analytics/`:

- `patients_longitudinal` — virtual dataset, one row per `scout_patient_id`, aggregated over `reports_latest_epic_view` + `reports_dx_epic_view`.
- `reports_dx_epic_view` — patient-resolved diagnosis dataset.
- `patient_cohort_events` — per-exam new/returning tagging that drives the trend chart.

## Type of change
- [x] New feature
- [ ] Improvement
- [ ] Bug fix
- [x] Documentation update
- [ ] Test update

##### Authorization
No new roles or AuthZ logic. Superset runs these queries through Trino as the logged-in user (ADR 0022), so the existing OPA row filters (`allowed_facilities`) and PHI column masks apply unchanged, and the datasets read the SECURITY-DEFINER `_epic_view`s (ADR 0023) exactly as the existing core/follow-up dashboards do. The patient datasets expose identifier columns (`scout_patient_id`, `resolved_epic_mrn`, `resolved_mpi`) for dedup/joins, but the dashboard tiles aggregate (counts/distributions) and don't render row-level identifiers. Read-only Trino; no REST/API surface added.

##### Appsec
N/A

### Performance
**TL;DR — fine at small scale; probably too slow for interactive use at 20M. Materialize before leaning on it heavily.**

Claude's analysis:

> Patient-level charts aggregate the whole lake (`GROUP BY scout_patient_id`), not a partition. On prod (~20M reports / ~1.75M patients) one chart measured **~21s** (`EXPLAIN ANALYZE`) — it scans `reports_latest` (14M) and `reports_report_patient_mapping` (21.6M, **twice**, across ~12.4k tiny files) through the epic views. Superset fires ~12 such queries per load with no shared scans, so a **cold dashboard render is on the order of minutes** (heaviest tiles may hit the query timeout); the 2h result cache makes repeat loads fast, but **every filter change re-pays the cost**.
> 
> Mitigations: **(1)** materialize a one-row-per-patient "gold" Delta table in the transformer → sub-second reads (separate PR + ADR); **(2)** `OPTIMIZE` `reports_report_patient_mapping` — the tiny-file double-scan is most of the 21s and *every* epic-view query pays it

### Data
- **`observation_dt` is sparsely populated**, so the patient timeline (`first_exam_dt`/`last_exam_dt`) and the trend's event time use `COALESCE(observation_dt, requested_dt, message_dt)`
- Patient identity uses the reconciled columns; rows with inconsistent identity are excluded upstream by the views.

### Backward compatibility
Compatible

## Testing
Deployed on **dev04** 

## Note for reviewers
Consider performance issues and opine on whether this is worth it. I'm inclined to include it as incentive/reminder to address the concerns and possibly materialize the views.

## Checklist
- [x] My code adheres to the coding and style guidelines of the project (`pre-commit` run on changed files; analytics YAML is prettier-ignored)
- [x] I have commented my code, particularly in hard-to-understand areas (dataset/chart `description` fields document the data caveats)
- [ ] I have added or updated user documentation — N/A for an opt-in/experimental bundle (matches the quality/follow-up precedent); can add an RTD note if preferred
- [x] I have added or updated technical documentation (scout-dashboards README: file layout + rationale)
- [ ] I have added unit tests — N/A (declarative Superset assets)
- [ ] I have added end-to-end tests — N/A (declarative Superset assets)


cmoore01:
Does this dashboard actually require `resolved_epic_mrn`? The visuals I see in the screenshot all look like aggregates where you'd only need the `scout_patient_id`. You might be able to get better performance by doing a simple join from the particular report table to the mapping table and ignoring the view.

kathrynalpert:
> Does this dashboard actually require `resolved_epic_mrn`? The visuals I see in the screenshot all look like aggregates where you'd only need the `scout_patient_id`. You might be able to get better performance by doing a simple join from the particular report table to the mapping table and ignoring the view.

It's a great suggestion, but end users don't have access to the mapping table because there's no way to limit it by facility.

johnflavin (approved):
> Consider performance issues and opine on whether this is worth it. I'm inclined to include it as incentive/reminder to address the concerns and possibly materialize the views.

It's hard to know if it is worth it until we see the additional load it creates on trino once users start interacting with it regularly. But I suppose if it creates a huge burdensome load that would be because it's very successful and people like to use it a lot, so that makes me lean towards trying it out.

PR #430, opened 2026-06-10
Bump open-webui chart to ~14.8.0 and OWUI to 0.9.6 to enable auto-login

andylassiter:
### Product
Users on Chat now auto-login instead of having to click "Sign in with SSO" again. One less click for the user.

### Technical
Bumps the open-webui Helm chart from ~14.6.0 to ~14.8.0 (appVersion 0.9.5 → 0.9.6). The 0.9.6 release adds OAUTH_AUTO_REDIRECT functionality so we no longer have the double login experience.

Tool-call display regression filter (tool_result_attr_filter) stays. Confirmed the bug is still present in 0.9.6 by comparing the SHAs of ToolCallDisplay.svelte and MarkdownTokens.svelte between v0.9.5 and v0.9.6, they're identical. Updated the version anchors in versions.yaml and the role README so the next renovate bump retests at the right version.

## Type of change
- [ ] New feature
- [x] Improvement
- [ ] Bug fix
- [x] Documentation update
- [ ] Test update

##### Authorization
N/A, no changes to auth flow itself.

##### Appsec
N/A

### Performance
N/A

### Data
N/A.

### Backward compatibility
Should be.

## Testing
Deploy to dev02 with `make install-chat`, verify SSO auto-redirect works, smoke-test a tool-calling chat to confirm the workaround filter still renders tool results in the Output panel.

## Note for reviewers
Users briefly see the OWUI Sign In page but are auto redirected from there.

## Checklist

### DIFF
--- a/ansible/group_vars/all/versions.yaml
+++ b/ansible/group_vars/all/versions.yaml
@@ -104,18 +104,19 @@ jupyter_helm_chart_version: ~4.3.5
 # running Pod's image at deploy time (see configure_admin.yaml) so it stays
 # in lockstep with the chart's appVersion without a separate pin.
 #
-# Currently pinned to ~14.6.0 (appVersion 0.9.5): upstream refactor (OWUI
+# Currently pinned to ~14.8.0 (appVersion 0.9.6): upstream refactor (OWUI
 # commit 45e49d33e, Apr 2026) moved tool-call results from a `result="..."`
 # attribute on the <details> tag into the body, and the new body-content
 # rendering path doesn't display results in the chat UI even though they
 # reach the LLM. Worked around by the `tool_result_attr_filter` Function
-# (see ansible/roles/open-webui/defaults/main.yaml). When Renovate proposes
-# a bump that advances appVersion past 0.9.5, verify the tool-call Output
-# panel renders correctly without the filter — if upstream has fixed it,
-# delete the filter source + the role's inventory entry + the chart payload
-# + this note.
+# (see ansible/roles/open-webui/defaults/main.yaml). Confirmed still broken
+# in 0.9.6: ToolCallDisplay.svelte and MarkdownTokens.svelte are byte-
+# identical between v0.9.5 and v0.9.6. When Renovate proposes a bump that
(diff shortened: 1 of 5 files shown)


kathrynalpert (commented):
I think if you update the URL that Launchpad uses, you'll get rid of the "Users briefly see the OWUI Sign In page but are auto redirected from there" (which is the current state we're trying to move past with this change)

andylassiter:
I corrected the Launchpad link, but the behavior is the same. The redirect happens pretty quickly though.

andylassiter:
Trivy scan is blocking the launchpad build. Looking into that now.

andylassiter:
Launchpad is using a node v25 base image which is now EOL. I don't think the base image is receiving updates anymore.

https://nodejs.org/en/about/previous-releases

> Historically (up to Node.js 26), odd-numbered releases (9, 11, etc.) become unsupported after six months, and even-numbered releases (10, 12, etc.) move to Active LTS status and are ready for general use. Starting with Node.js 27, the release cycle will be annual and every major version will move to LTS status after its six-month Current phase (and six additional months of Alpha phase). LTS release status is "long-term support", which typically guarantees that critical bugs will be fixed for a total of 30 months. Production applications should only use Active LTS or Maintenance LTS releases.

kathrynalpert:
> Launchpad is using a node v25 base image which is now EOL. I don't think the base image is receiving updates anymore.

Feel free to do any of the following:
1. remove the launchpad change from this since it doesn't appear to matter :( 
2. bump launchpad base image + deps
3. add trivyignore entry and do (2) as a separate, subsequent PR
4. merge with failing trivy scan (these don't block merge)

kathrynalpert (approved):

andylassiter:
I tried 2. But had issues bumping to 26. `@types/node@26` is not available yet, it is still on v25. Also v26 is considered `Current` not `Acitive LTS` ("Production applications should only use Active LTS or Maintenance LTS releases.") so I think we would need to downgrade to v24. I tried that, too, but the openssl fix was just released yesterday and hasn't made it to the `node:24-alpine` image yet. I'm going with option 4 for now.

PR #431, opened 2026-06-11
Force Superset Database Updates

andylassiter:
### Product
No user facing changes

### Technical
Add `update_database` to `scout-dashboards/import.py` alongside the existing dataset/chart/dashboard updaters. Port or scheme changes in`.Values.trino` never reached Superset's `dbs` row after the initial install.

## Type of change
- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
N/A.

##### Appsec
N/A.

### Performance
N/A. Runs once per deploy inside the existing Bootstrap Job.

### Data
N/A.

### Backward compatibility
Additive: returns early if the Database row does not exist, so clean installs are untouched. On upgrades it overwrites with the chart YAML.

## Testing
Deployed on dev02. Import Job logs `databases updated=1`; Superset's stored `sqlalchemy_uri` matches the rendered YAML (port now 8443/https, was stuck at 8080). I was then able to view data in Superset.

## Note for reviewers
This came up when deploying Trino RBAC to dev02. The Trino port change to 8443 was not being picked up by Superset on update.

## Checklist
- [x] My code adheres to the coding and style guidelines of the project

### DIFF
--- a/helm/scout-dashboards/files/import.py
+++ b/helm/scout-dashboards/files/import.py
@@ -1,19 +1,25 @@
-"""Force-overwrite Superset chart and dashboard params by UUID.
+"""Force-overwrite Superset Database / dataset / chart / dashboard params by UUID.
 
-Workaround for `superset import-dashboards` not reliably overwriting the
-`params`, `query_context`, and `position_json` of existing assets — it
-preserves the existing rows and silently ignores the new YAML for any UUID
-already in the DB.
+`superset import-dashboards` silently preserves existing rows on UUID
+match, so edits to the YAML never reach the DB without this script.
 
-Run AFTER `superset import-dashboards`. The CLI handles creating new assets
-(datasets, charts, dashboards); this script handles updating existing ones.
+Invoked twice by the import Job:
 
-For dashboards we also have to remap chartId references in position_json
-and json_metadata: YAML files capture chartId values from whatever DB the
-export came from, but on import Superset auto-assigns new local IDs by
-auto-increment. UUID is the only stable identifier across instances.
+  1. `python import.py <dir> --databases-only` runs BEFORE the CLI to
+     heal a stale `Database.sqlalchemy_uri`. Required because the CLI's
+     `import_dataset` opens a Trino connection when creating a new
+     dataset, and a stale URI aborts the Job before the post-CLI pass
+     can heal it.
+  2. `python import.py <dir>` (no flag) runs AFTER the CLI. Sweeps
+     databases / datasets / charts / dashboards and force-overwrites
+     params, position_json, json_metadata, etc. on existing rows.
 
-Usage: python import.py [analytics_dir]
+For dashboards we also remap chartId references in position_json and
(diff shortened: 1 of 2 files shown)


helm/scout-dashboards/files/import.py
```diff
@@ -61,6 +62,50 @@ def _coerce_bool(val) -> bool:
 )
 
 
+# Fields on `Database` we want the chart's YAML to be authoritative for. The
+# YAML lives in the chart's `_helpers.tpl` (database_name, sqlalchemy_uri,
+# allow_*, expose_in_sqllab, impersonate_user, extra). `password` is NOT here —
+# Trino auth is JWT-based and the password column should stay NULL. `uuid` is
+# the lookup key and must not be overwritten.
+_DATABASE_FIELDS = (
```
kathrynalpert: Hard-coding this means we have to stay in sync with superset.

Claude recommended some checks to detect drift:
```python
_unknown = set(_DATABASE_FIELDS) - {c.key for c in Database.__table__.columns}
assert not _unknown, f"_DATABASE_FIELDS not on Database model: {_unknown}"

# in update_database():
ignored = set(data) - set(_DATABASE_FIELDS) - {"uuid", "version", "extra", "password"}
if ignored:
  print(f"warning: database YAML keys not synced: {sorted(ignored)}")
```

It also noted [Superset's importer helper](https://github.com/apache/superset/blob/master/superset/commands/database/importers/v1/utils.py#L40): `superset.commands.database.importers.v1.utils.import_database(config, overwrite=True, ignore_permissions=True)` --> would it be better to use this directly?

kathrynalpert (approved):
Currently, `import.py` runs strictly after `superset import-dashboards` -- I believe the latter may use Trino to validate changes (e.g., if you add a new "dataset", Superset will query Trino to get its schema during the import). So it might be most correct to have this new logic around database updates run prior to the import. (Practically, would this ever be an issue? Hard to know. Probably not?)

johnflavin (approved):

helm/scout-dashboards/files/import.py
andylassiter: Not a fan of the hard-coded fields either. I'll try these out.

andylassiter:
> Currently, import.py runs strictly after superset import-dashboards -- I believe the latter may use Trino to validate changes (e.g., if you add a new "dataset", Superset will query Trino to get its schema during the import). So it might be most correct to have this new logic around database updates run prior to the import. (Practically, would this ever be an issue? Hard to know. Probably not?)

I had been asking Claude about the ordering and timing of things yesterday and also had it do another deep dive this morning. It still feels confident this isn't an issue (but Claude always feels confident). This seems like an easy refactor though so I'll try it.

andylassiter:
Pushed the refactor. I did test again on dev02. But you may want to do a quick re-review before I merge.

kathrynalpert (approved):
LGTM, the proof will be in the preprod upgrade!

PR #434, opened 2026-06-11
(Minor) Bump up helm timeout for scout-dashboards

johnflavin:
I had a timeout failure deploying `scout-dashboards` after the patient dashboards from #423

### DIFF
--- a/ansible/roles/superset/tasks/deploy.yaml
+++ b/ansible/roles/superset/tasks/deploy.yaml
@@ -137,6 +137,7 @@
     helm_chart_name: scout-dashboards
     helm_chart_ref: '{{ scout_dashboards_helm_chart_ref }}'
     helm_chart_namespace: '{{ superset_namespace }}'
+    helm_chart_timeout: '{{ superset_helm_timeout }}'
     helm_chart_values:
       image:
         repository: '{{ superset_image_repository }}'


ansible/roles/superset/tasks/deploy.yaml
```diff
@@ -137,6 +137,7 @@
     helm_chart_name: scout-dashboards
     helm_chart_ref: '{{ scout_dashboards_helm_chart_ref }}'
     helm_chart_namespace: '{{ superset_namespace }}'
+    helm_chart_timeout: '{{ superset_helm_timeout }}'
```
kathrynalpert, line 140: Wait we already had this var defined and just weren't using it?

kathrynalpert (approved):

ansible/roles/superset/tasks/deploy.yaml
johnflavin: I believe it was used elsewhere and I'm repurposing it here

johnflavin: Yes, already used on line 99 of this file: https://github.com/washu-tag/scout/blob/273a6d1092994f19d16530d0b9e7ff3a17fa6ae7/ansible/roles/superset/tasks/deploy.yaml#L99

kathrynalpert: Yeah, I forgot we now have 2 charts for superset to be gitops-y. One for superset itself, one for our custom dashboards!

PR #437, opened 2026-06-12
Optional Data generator deployment

cmoore01:
### Product
Users are not expected to interact with this.

### Technical
Adds optional deployment of data-generator project so developers can more easily create datasets for testing and/or demos. Deployment is for a temporal worker with configurable number of replicas.

## Type of change
- [X] New feature
- [ ] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
N/A

##### Appsec
N/A

### Performance
Technically, you can use it with ollama as the target for LLM report generation, but the performance is quite, quite slow.

### Data
N/A

### Backward compatibility
Compatible.

## Testing
Working on dev01

## Checklist


ansible/playbooks/data_generator.yaml
```diff
@@ -0,0 +1,11 @@
+---
+- name: Deploy data-generator
+  hosts: server
+  gather_facts: false
+  pre_tasks:
+    - name: Skip if data-generator not enabled
+      ansible.builtin.meta: end_play
+      when: not (enable_data_generator | default(true) | bool)
```
andylassiter: Should this be `default(false)`?

helm/data-generator/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "helm.name" -}}
```
andylassiter: I think `helm.*` should be renamed to `data-generator.*` here and in your other templates. It looks like this chart was created with `helm create helm` then the folder was renamed to `data-generator`? Take a look at our other helm charts like [launchpad](https://github.com/washu-tag/scout/blob/main/helm/launchpad/templates/_helpers.tpl#L4) or [scout-dashboards](https://github.com/washu-tag/scout/blob/main/helm/scout-dashboards/templates/_helpers.tpl#L4).

helm/data-generator/templates/ingress.yaml
```diff
@@ -0,0 +1,43 @@
...
+          {{- range .paths }}
+          - path: {{ .path }}
+            {{- with .pathType }}
+            pathType: {{ . }}
+            {{- end }}
+            backend:
+              service:
+                name: {{ include "helm.fullname" $ }}
+                port:
+                  number: {{ $.Values.service.port }}
```
andylassiter: Claude is saying this ingress will fail because we don't create a `service` in the chart chart. But do we even need an `ingress` for the data generator? I think this file can be dropped (and the `ingress:` in `values.yaml`).

helm/data-generator/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: data-generator
+description: Helm chart to deploy the data-generator for creating synthetic data
+type: application
+version: 0.1.0
+appVersion: "1.0.0"
```
andylassiter: I think this should be:
```
version: 0.0.0-dev
appVersion: "latest"
```
Does the data-generator have a version?

andylassiter (approved):

PR #439, opened 2026-06-15
Fix execute_sql connection-target leak across roles

johnflavin:
### Product
N/A

### Technical
`scout_common`'s `execute_sql` never consumed the `sql_dbname` parameter and instead self-referentially `set_fact`'d its connection variables (`db_connect_name`, etc.). Because `set_fact` persists as a host fact for the rest of the play, one role's connection target leaked into every later `execute_sql` call. In a full deploy, the hive role sets the target to the `hive` DB (in `create_readonly_user`); open-webui's later `ALTER SCHEMA public OWNER TO openwebui` then ran against the leaked `hive` DB instead of `openwebui`, reassigning hive's `public` schema to `openwebui`.

We observed this on a dev cluster: a routine hive image bump triggered a metastore rollout, and the new pod CrashLoopBackOff'd with `permission denied for schema public` (the `hive` role no longer owned its own schema). The version bump only surfaced the pre-existing corruption — the long-running old pod never re-validated its schema.

Fix: `execute_sql` now derives connection settings into fresh `psql_*` facts with inline defaults each invocation (no cross-role leakage), and consumes `sql_dbname` as the canonical target DB (`db_connect_name` still honored for back-compat). The hive caller switched to `sql_dbname`.

## Type of change
- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
N/A

##### Appsec
N/A — fix prevents DDL from running against an unintended database, which had reassigned schema ownership cross-database.

### Performance
N/A

### Data
N/A

### Backward compatibility
`sql_dbname` is now the canonical parameter; legacy `db_connect_name` is still honored, so existing callers continue to work.

## Testing
YAML validated for the changed task files. Manually verified the fix on the affected dev cluster: restored `hive` schema ownership, restarted the metastore, and confirmed the schema re-initialized (74 tables, `VERSION` 3.1.0) and the pod went Ready.

## Note for reviewers
N/A

## Checklist

### DIFF
--- a/ansible/roles/scout_common/tasks/execute_sql.yaml
+++ b/ansible/roles/scout_common/tasks/execute_sql.yaml
@@ -5,11 +5,9 @@
 # later execute_sql call. (That leak previously let open-webui's ALTER SCHEMA
 # run against the hive database.) Instead we derive everything below into
 # distinct psql_* facts, applying defaults inline so each call starts clean.
-# `sql_dbname` is the canonical parameter; `db_connect_name` is still honored
-# for backward compatibility.
 - name: Configure DB derived variables
   ansible.builtin.set_fact:
-    psql_dbname: "{{ sql_dbname | default(db_connect_name | default('postgres', true), true) }}"
+    psql_dbname: "{{ sql_dbname | default('postgres', true) }}"
     psql_user: "{{ db_connect_user | default('postgres', true) }}"
     psql_pass: '{{ db_connect_pass | default(postgres_superuser_password, true) }}'
     psql_sslmode: "{{ db_connect_sslmode | default('require', true) }}"
--- a/ansible/roles/scout_common/tasks/execute_sql.yaml
+++ b/ansible/roles/scout_common/tasks/execute_sql.yaml
@@ -1,15 +1,24 @@
-- name: Configure DB variables
-  ansible.builtin.set_fact:
-    db_connect_name: "{{ db_connect_name | default('postgres', true) }}"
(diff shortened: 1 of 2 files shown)


ansible/roles/scout_common/tasks/execute_sql.yaml
```diff
@@ -1,15 +1,24 @@
...
+# distinct psql_* facts, applying defaults inline so each call starts clean.
+# `sql_dbname` is the canonical parameter; `db_connect_name` is still honored
+# for backward compatibility.
+- name: Configure DB derived variables
   ansible.builtin.set_fact:
-    db_connect_name: "{{ db_connect_name | default('postgres', true) }}"
-    db_connect_user: "{{ db_connect_user | default('postgres', true) }}"
-    db_connect_pass: '{{ db_connect_pass | default(postgres_superuser_password, true) }}'
-    db_connect_sslmode: "{{ db_connect_sslmode | default('require', true) }}"
+    psql_dbname: "{{ sql_dbname | default(db_connect_name | default('postgres', true), true) }}"
```
kathrynalpert: I don't really think the complexity here is warranted, I don't think we need backward compatibility. I would vote for `"{{ sql_dbname | default('postgres', true) }}"`

kathrynalpert (approved):

ansible/roles/scout_common/tasks/execute_sql.yaml
johnflavin: Fixed.

PR #453, opened 2026-06-24
feat(hl7-transformer): upgrade to Spark 4.1.1 and Delta Lake 4.3.0

kathrynalpert:
### Product
N/A — no user-facing change; the silver Delta tables and how they're queried are unchanged.

### Technical
Upgrades the hl7-transformer runtime to **Apache Spark 4.1.1** (`scala2.13-java21` base) and **Delta Lake 4.3.0**, with the supporting build/CI changes:

- **Image**: base → `spark:4.1.1-scala2.13-java21`; jars → `delta-spark_4.1_2.13` 4.3.0 + `delta-storage` 4.3.0, `hadoop-aws` 3.4.2, **AWS SDK v2** bundle 2.46.17 (replacing the v1 bundle); drop the hand-pinned Hadoop 3.2.2 / antlr jars (the base bundles them); remove the duplicate pyspark-wheel jar copy; upgrade the venv's pip/setuptools.
- **pyproject**: `pyspark==4.1.1`, `delta-spark==4.3.0`, `temporalio ~= 1.29.0`, `requires-python >= 3.10`. Stays on Spark 4.1.1 (not 4.1.2) because delta-spark pins `pyspark<=4.1.1`.
- **spark-defaults**: `spark.sql.ansi.enabled false` (Spark 4 defaults ANSI on; the HL7 extraction relies on out-of-range array access and unparseable timestamps yielding NULL), and the AWS-lane S3A credentials provider moves to the SDK v2 `WebIdentityTokenFileCredentialsProvider`.
- **tests/ingest**: `_2.13` Spark 4.1.1 / Delta 4.3.0 artifacts, exclude `query-spec`'s transitive Spark 3.5 jars, add Spark 4's two new JVM module options, and a 1g gradle-daemon heap (the Spark 4 dependency graph OOMs the 512MiB default).
- **Trivy**: make the hl7-transformer scan enforcing (image is clean) with a reasoned `.trivyignore.yaml` for the base-image CVEs we can't patch.

## Type of change
- [x] Improvement

##### Authorization
N/A.

##### Appsec
Clears the aging Spark 3.5.4 base's CVE backlog; the hl7-transformer scan is now a blocking gate. Remaining findings are upstream Spark-base jars, covered by a reasoned ignorefile.

### Performance
Engine-output parity timed over the staging corpus (within noise of 3.5.4); no behavioral change.

### Data
Behavior-preserving. Extraction output is byte-identical between Spark 3.5.4 and 4.1.1 with ANSI off (0/910 rows differ over the staging corpus, timestamps included). Delta 4.3.0 introduces no default table-feature/protocol change that affects existing readers; Trino reads the freshly-written tables in CI.

### Backward compatibility
Data/query contract unchanged. The change is an internal runtime upgrade (Spark 3.5→4.1, Scala 2.12→2.13, Java 17→21, Delta 3.x→4.3.0, AWS SDK v1→v2); the AWS-lane S3A credentials-provider config moves to the SDK v2 class to match.

## Testing
- New hl7-transformer unit tests pass under pyspark 4.1.1 (and under ANSI on — the reader is ANSI-safe).
- Engine parity 3.5.4 vs 4.1.1 over the 910-message staging corpus: 0 differences.
- `tests/ingest` compiles against the Spark 4.1.1 / Delta 4.3.0 `_2.13` artifacts.
- CI deploy-and-test exercises the full ingest → Delta-4.3.0-write → Trino-read path.

## Note for reviewers
Top of the stack: based on `pr/smolder-native-reader` (#452), which is based on `pr/cve-refresh` (#451). GitHub auto-retargets the base as each lands, so review/merge bottom-up (#451 → #452 → this).

## Checklist
- [x] My code adheres to the coding and style guidelines of the project (ran `pre-commit` on the changed files)
- [x] I have updated documentation where appropriate (CLAUDE.md prerequisites)
- [ ] Technical documentation / ADR — N/A
- [ ] Unit tests — N/A here (the new reader's tests land in #452; existing suites pass on the bump)
- [ ] End-to-end tests — N/A (CI deploy-and-test covers the path)


tests/ingest/build.gradle
```diff
@@ -75,12 +77,17 @@ test {
 }
 
 dependencies {
-    testImplementation("edu.washu.tag:query-spec:${vQuerySpec}")
+    // query-spec declares spark-sql_2.12 (Spark 3.5) at runtime scope; Spark 4
+    // is Scala 2.13 only, so excluding it here keeps the 2.12 jars off the
+    // classpath and runs query-spec against the 2.13 Spark below instead.
+    testImplementation("edu.washu.tag:query-spec:${vQuerySpec}") {
+        exclude group: 'org.apache.spark'
+    }
```
johnflavin, line 86: @cmoore01 Would it be possible to build an updated `query-spec` against scala 2.13 + spark 4 so we can avoid this workaround here?

cmoore01: I would think that should be doable, yep.

kathrynalpert: If we made it a `compileOnly` dep, would that be better so we don't have to keep the two in sync (provided APIs don't conflict)

johnflavin: > If we made it a `compileOnly` dep, would that be better so we don't have to keep the two in sync (provided APIs don't conflict)

I don't understand. Do you mean the `query-spec` dependency here or the spark dependency within `query-spec`?

kathrynalpert:
The latter

tests/ingest/build.gradle
johnflavin: I've added a note to the comment that when washu-tag/data-generator#14 is done we can remove this `exclude`. But I am fine merging as-is.

johnflavin (approved):

Issue #457, opened 2026-06-29
Split hl7-transformer temporal activities


johnflavin:
The data ingest from HL7 and the derivative table generation should be in different idempotent temporal activities so a failure in a later part doesn't cause the whole ingest to rerun.

cmoore01:
How is the `SparkSession` going to get passed along? Or is the emphemeral spark being torn down and rebuild between these activities. I agree this is better design, but I'm a bit concerned it might be better to wait until we have a persistent spark service.

johnflavin:
That is a legit concern. It's possible we may need different spark sessions per activity, or a single spark session that gets instantiated outside the activities and passed in (which would then mean activities need to be throttled for access to this shared resource), or maybe we have the activities in different pods entirely. I am not sure what the implementation should be.

PR #474, opened 2026-06-30
fix: harmonize default hl7-transformer timeout

### LINKED ISSUE #473: hl7_transformer_timeout default from jinja template doesn't match main.yaml
In #370 , I updated `deltaIngestTimeout: {{ hl7_transformer_timeout | default(480) }}` in `hl7log-extractor.values.yaml.j2`. However, that default never actually gets used because the `main.yaml` default is already defined: `hl7_transformer_timeout: 60`.

cmoore01:
### Product
N/A

### Technical
See #473 

## Type of change
- [ ] New feature
- [ ] Improvement
- [X] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
N/A

##### Appsec
N/A

### Performance
N/A

### Data
N/A

### Backward compatibility
N/A

## Testing
N/A

## Note for reviewers
I'm waffling between just saying we should drop this `default` entirely since we already define it. Maybe it's defense-in-depth? I'm leaving the structure as it is for now.

## Checklist

### DIFF
--- a/ansible/roles/extractor/defaults/main.yaml
+++ b/ansible/roles/extractor/defaults/main.yaml
@@ -15,7 +15,7 @@ hl7_transformer_image_tag: latest
 hl7log_extractor_timeout: 480
 hl7log_extractor_heartbeat_timeout: 30
 hl7log_extractor_concurrency: 150
-hl7_transformer_timeout: 60
+hl7_transformer_timeout: 480
 # HL7 Transformer resources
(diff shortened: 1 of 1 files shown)


kathrynalpert (approved):

kathrynalpert:
Arguably, it'd be nice to remove the `| default(480)` that isn't used, but I think we should revisit the whole thing as part of #457

PR #272, opened 2026-01-09
HL7 Listener ADR + POC

andylassiter:
### Product
N/A

### Technical
Add Architecture Decision Record (ADR) for HL7 listener implementation using Apache Camel K and Kafka. Documents POC findings, proposed architecture, and production considerations for real-time HL7 message ingestion.

## Type of change
- [ ] Work behind a feature flag
- [ ] New feature
- [ ] Improvement
- [ ] Bug fix
- [ ] Refactor (code improvement with no functional changes)
- [x] Documentation update
- [ ] Test update

##### Authorization
N/A

##### Appsec
N/A

### Performance
N/A

### Data
N/A

### Backward compatibility
N/A

## Testing
N/A

## Note for reviewers
The POC role can be deployed independently via `make install-hl7-listener` without affecting existing Scout deployments.

## Checklist
- [x] My code adheres to the coding and style guidelines of the project (and I've run `pre-commit run --all-files`)
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added or updated user documentation, if appropriate
- [x] I have added or updated technical documentation, including an architectural decision record, if appropriate
- [ ] I have added unit tests, if appropriate
- [ ] I have added end-to-end tests, if appropriate

PR #475, opened 2026-07-01
feat(hl7-listener): opt-in, event-driven HL7 ingest path for collection and observation

kathrynalpert:
### Product
Stands up a real-time HL7 ingestion path so Scout can eventually receive reports as they are produced (over MLLP) instead of via nightly log files, removing the ingestion latency, WashU specificity, and opening possibility for change tracking. In this PR the listener runs in **shadow mode**: it captures the live feed to object storage but nothing downstream consumes it yet, so there is no user-facing change. The goal is to observe real traffic and plan the cutover from log-file to event-driven ingest.

### Technical
Adds a real-time pipeline alongside the existing log-file pipeline: `MLLP :2575 → listener → Kafka (Strimzi, KRaft) → batcher (zip) → object-storage bronze`. The listener and batcher are a single **Camel-on-Spring-Boot** app (Camel 4.18.2 LTS): the routes stay declarative YAML, wrapped in a Gradle/Spring Boot project + Dockerfile, **built in CI and published to GHCR** like every other Scout image, and deployed from a Helm chart (`helm/hl7-listener`) of plain Deployments (Flux-ready for the GitOps transition). The batcher consumes Kafka via the **batching consumer** and commits offsets only after the durable object-storage write, giving **at-least-once** delivery with Kafka as a replayable buffer. This supersedes the #272 Camel K POC — **no in-cluster operator build, no per-Integration image, and no prod→staging registry push**; prod only pulls (Harbor mirror when air-gapped). The `hl7-raw` bucket, and the batcher's S3 auth, are provisioned by the `hl7-listener`/`minio` roles; the chart carries only the workloads. Monitoring ships as Grafana dashboards (HL7 Listener and Kafka/Strimzi), a Prometheus scrape of the Spring Boot Actuator `/actuator/prometheus` endpoint, and four Grafana alerts: no-HL7-received, pipeline exchange failures, batcher-stalled (Kafka lag building), and Kafka-broker-down. ADR **0028** documents the architecture (including why not Camel K).

## Type of change
- [x] Work behind a "feature flag" (here: "shadow mode")
- [ ] New feature
- [ ] Improvement
- [ ] Bug fix
- [x] Documentation update
- [ ] Test update

##### Authorization
N/A — no changes to user roles, Trino/OPA policy, or the data users can see. The listener has no user-facing surface.

##### Appsec
- The MLLP listener accepts unauthenticated TCP on port 2575, which is standard for HL7 v2 interface engines; access is controlled at the network layer and should be restricted to the upstream sender.

- Contrary to a first pass using Camel K (in-cluster build with prod→staging registry push and Harbor push credential in the production cluster.), the application image is **built in CI and only pulled** by the cluster (through the Harbor `ghcr-proxy` mirror when air-gapped), exactly like every other Scout image. This closes the supply-chain concern raised on #272. 

- The batcher writes object storage as `lake-writer`: a static-key secret on MinIO, and an **IRSA ServiceAccount (no static keys)** on AWS.

### Performance
Sized for a guess at the production feed (~200K messages/day, bursty), not the ADT test trickle. The batcher consumes Kafka with the **batching consumer** — `maxPollRecords: 1000` caps burst batches and `pollTimeoutMs: 30000` flushes partial batches after 30 s of inactivity — and zips each batch in a single O(n) pass. At prod volume it writes ~1000-message zips (~200/day, a few MB each) rather than the ~20K tiny zips a size of 10 would produce (see @johnflavin's comment on #272). Offsets are committed only after the durable upload (at-least-once), and the highest offset **per partition** is committed so every partition advances.

Kafka is a single-broker KRaft node (10Gi PV) used only as a transport buffer, since object-storage bronze is the durable record; retention is capped at 2 days and `retention.bytes` at 2Gi/partition (~6Gi across 3 partitions) so a high-volume feed can't fill the PV and crash the broker. The listener and batcher pods are lightweight (100m–500m CPU, 256–512Mi).

### Data
Raw HL7 lands in a dedicated `hl7-raw` bucket (`s3://hl7-raw/hl7-batches/YYYY/MM/DD/*.zip`), physically separate from the log-file bronze (`s3://lake/hl7`) and the Delta silver (`s3://lake/delta`). The bucket and `lake-writer`'s access to it are declared in the `minio` role (on-prem), alongside the other Scout buckets. Nothing downstream reads `hl7-raw` yet, so there is no change to the `reports` table or any existing data — it's purely additive capture. Batch files/paths are timestamped by batch time rather than receipt time, which is acceptable for shadow observation and noted for revisit at cutover.

### Backward compatibility
Fully backward compatible. The existing log-file → Temporal → transformer path is unchanged and remains the active pipeline. The listener stack is an **independent, opt-in deploy (`make install-hl7-listener`)**; not deploying it changes nothing.

## Testing
Validated on dev04 against the live hospital test feed with the CI-shaped image (built via `docker buildx` for linux/amd64 and pushed to GHCR as `hl7-listener:latest-test`). The live Cloverleaf feed (ADT) flows end-to-end, and a synthetic `ORU^R01` was pushed through the new build: MLLP receive + `AA` ACK → HAPI parse → Kafka → zip → `hl7-raw` upload → batch manifest back to Kafka.

The dev04 run also exercised the at-least-once path and surfaced a multi-partition commit bug: only the last record's partition was being committed, so other partitions lagged and re-delivered on restart. Fixed by committing the highest offset per partition — after the fix all partitions reach and hold `lag=0`, and an uncommitted message **replays** (rather than drops) across a batcher restart, confirming at-least-once.

No automated unit/e2e tests were added.

## Note for reviewers
- **Shadow mode / bronze-only.** The listener captures to the separate `hl7-raw` bucket; the log-file pipeline is untouched. The intent is to learn about the real feed before wiring the transformer cutover, which is out of scope here.
- **Supersedes #272** ("HL7 Listener ADR + POC"), with @andylassiter credited as co-author; the ADR is carried forward as **ADR 0028**. This implements Andy's own follow-up — wrap the routes in a Gradle project + Dockerfile + a plain Deployment rather than have the Camel K operator build them.
- **Why not Camel K.** The operator builds a per-Integration image *in-cluster* at deploy time, which needs Maven access and registry **push** credentials — in air-gapped mode a prod→staging Harbor push. Building in CI and pulling (like every other service) removes that push, drops the fragile in-cluster air-gapped build path, and yields a plain Helm chart Flux can own. See ADR 0028.
- **Batching / at-least-once.** The batcher uses Camel's Kafka batching consumer (one zip per poll, O(n)) rather than `ZipAggregationStrategy` (O(n²) rewrites) or a hand-rolled `AggregationStrategy`, and defers the offset commit until after the S3 upload. See ADR 0028's "Why batching" / "At-least-once delivery".
- **Cross-backend storage.** The `hl7-raw` bucket + `lake-writer` policy live in the `minio` role on-prem; on **AWS** they must be provisioned in Terraform (the bucket and the `lake-writer` IAM role policy), and the batcher authenticates via IRSA. That AWS side is not covered here, and is unvalidated until it runs on EKS (the on-prem/MinIO path is what dev04 exercises).
- **Deploy ordering.** `make install-lake` provisions the `hl7-raw` bucket (via the `minio` role), so it must run before `make install-hl7-listener`. 
- Deployable independently via `make install-hl7-listener` without affecting existing Scout deployments.

## Checklist


kathrynalpert:
Converting back to draft while I investigate how to avoid pushing an image from prod -> staging. This is a requirement for using the Camel K operator, so I think we need to move away from that pattern, as we don't want any network traffic or creds from prod to staging.

andylassiter:
> Converting back to draft while I investigate how to avoid pushing an image from prod -> staging. This is a requirement for using the Camel K operator, so I think we need to move away from that pattern, as we don't want any network traffic or creds from prod to staging.

Agree. We would need to build the image ourselves instead of having the camel-k operator build it. I think the routes could stay as yaml, then wrap it in a gradle project + dockerfile and deploy a `Deployment`.

ansible/roles/hl7-listener/defaults/main.yaml
```diff
@@ -0,0 +1,92 @@
...
+kafka_hl7_batches_topic: hl7-batches
+kafka_hl7_batches_topic_partitions: 1
+kafka_hl7_batches_topic_replicas: 1
+
+# S3/MinIO configuration for batch uploads
+hl7_batcher_s3_bucket: '{{ hl7_raw_bucket | default("hl7-raw") }}'
+hl7_batcher_s3_endpoint: 'http://minio.{{ minio_tenant_namespace | default("minio-tenant") }}.svc.cluster.local:80'
+hl7_batcher_s3_region: us-east-1
+hl7_batcher_s3_access_key: '{{ s3_lake_writer | default("lake-writer") }}'
+hl7_batcher_s3_secret_key: '{{ s3_lake_writer_secret | default("") }}'
```
johnflavin, line 81: Should add an assertion to the role that this is not empty / `''`, otherwise if someone forgets to set this in their inventory the deploy will succeed but S3 auth will fail.

ansible/roles/hl7-listener/tasks/deploy.yaml
```diff
@@ -0,0 +1,225 @@
...
+    api_version: v1
+    kind: Namespace
+    state: present
+
+- name: Create operators namespace
+  kubernetes.core.k8s:
+    name: '{{ strimzi_operator_namespace }}'
+    api_version: v1
+    kind: Namespace
+    state: present
```
johnflavin: We have a reusable task in `scout_common` to create namespaces. It's small, but why not?

hl7-listener/src/main/resources/camel/batcher.camel.yaml
```diff
@@ -0,0 +1,54 @@
...
+            name: CamelFileName
+            expression:
+              simple: '${header.kafka.KEY}-${date:now:yyyyMMddHHmmssSSS}.hl7'
+        - log:
+            message: 'Adding message to batch: ${header.CamelFileName}'
+            loggingLevel: DEBUG
+        - aggregate:
+            completionSize: '{{hl7.batcher.completion.size}}'
+            completionTimeout: '{{hl7.batcher.completion.timeout}}'
+            aggregationStrategy: '#zipAggregator'
```
johnflavin: This was flagged by Claude as a bad fit for our needs, and an `O(n^2)` cost in disk IO and CPU (though notably not memory). Every time a message comes in, this appends a new message to a zip on disk. Though "append" is not the right word because the zip format can't append, once written it has a fixed index, it can only rewrite the whole thing with all the existing messages + a new message. That means copying the old zip (if there are any previous messages already written) and adding a single new message, then writing it all out again. That happens for every message, meaning if there are n messages the first message gets written n times, hence the `O(n^2)` IO.

I asked it some follow-up questions about why it is this way and what it's good for.

> The O(n²) is a real consequence of two deliberate design choices that are fine at its intended scale:
>
> - **The zip format can't be appended to.** Once a zip's central directory is written, you can't tack on another entry; you must rewrite the archive. That's a `java.util.zip` / format limitation, not a Camel bug.
> - **Camel chose bounded memory over speed.** `ZipAggregationStrategy` streams the existing archive from *disk* (rename-to-temp → copy all entries → append the new one) rather than holding every message in the heap. So each aggregation is O(current size) and it never buffers the whole batch in memory — but n aggregations cost 1+2+…+n = O(n²) of disk/CPU.
>
> **Intended use:** aggregating a *small* number of files/messages into one archive — the canonical "poll a directory / split a file, then re-zip the pieces" pattern where n is tens to low hundreds. At that scale the rewrite cost is negligible and you get a valid zip materialized after every step with flat memory. Nobody designed it to stream 1000s of live messages, which is what `completionSize: 1000` does here.
>
> **What to use instead** (ranked by effort/fit):
>
> 1. **Custom aggregation strategy that collects bytes and zips once at completion (recommended).** Accumulate `(filename, byte[])` in a list during aggregation, then build the zip in a single O(n) pass in the completion route. Messages are a few KB each, so 1000 of them is a few MB of heap — trivial against the 512Mi limit (and cheaper than the current temp-file churn). This also composes naturally with the byte-size batching the ADR said would need a custom `completionPredicate` anyway.
> 2. **Custom strategy holding one open `ZipOutputStream`** (to a temp file, or streamed to S3), calling `putNextEntry` once per message and `finish()` on completion. O(n) time, bounded memory. More code than 1, but keeps memory flat if you're worried about very large batches.
> 3. **Switch the container format to tar(.gz).** `TarArchiveOutputStream` (commons-compress) appends linearly with no central-directory rewrite → naturally O(n). Individual messages stay extractable. The catch is the downstream reader: the file-based path (`hl7log-extractor` → `hl7-transformer`) produces/consumes zip, so this changes the eventual cutover contract — *but* nothing consumes `hl7-raw` yet (observer mode), so this is the cheapest moment to pick the better format if you're open to it.
> 4. **Stopgap only:** drop `completionSize` to ~50–100 and lean on `completionTimeout`. Reduces the constant but not the scaling, and inflates object count — a mitigation, not a fix.
>
> My recommendation is **1**: smallest change, O(n), well within the memory budget, and it sets you up for size-based batching later. If you expect occasional very large bursts and want to keep memory strictly flat, **2**. Consider **3** only if you're willing to revisit the archive format before the transformer cutover.

hl7-listener/src/main/resources/camel/batcher.camel.yaml
```diff
@@ -0,0 +1,54 @@
...
+    from:
+      uri: 'kafka:{{kafka.hl7.topic}}?brokers={{kafka.brokers}}&groupId={{hl7.batcher.group.id}}&autoOffsetReset=earliest'
+      steps:
+        - log:
+            message: 'Consumed from Kafka: partition=${header.kafka.PARTITION}, offset=${header.kafka.OFFSET}, key=${header.kafka.KEY}'
+            loggingLevel: DEBUG
+        - setHeader:
+            name: CamelFileName
+            expression:
+              simple: '${header.kafka.KEY}-${date:now:yyyyMMddHHmmssSSS}.hl7'
```
johnflavin: The `${date:now:...}` naming makes me a bit nervous about unlikely but possible collisions because it's only down to the millisecond. 

It would be safer to use `${exchangeId}` instead. That's a globally unique UUID that camel puts on "exchanges", which is its generic term for any messages + replies. See https://camel.apache.org/components/4.18.x/languages/simple-language.html#_core_camel_functions

It's still wildly unlikely that two messages with the same `${header.kafka.KEY}`—which I think is the MSH-10 message control ID—would be processed at the same millisecond, which would be the actual collision risk. But if we can eliminate it entirely by using the `${exchangeId}` instead of a timestamp, why not?

hl7-listener/src/main/resources/camel/batcher.camel.yaml
johnflavin, line 1: Claude flagged a risk of losing messages here because they get accepted when written _to the zip_ but not _to s3_. If we're mid zip batch and the batcher restarts, kafka won't replay those messages because they've already been accepted.

> Batcher Kafka consumer uses default auto-commit, so offsets advance when a message enters the in-memory zip aggregator — before the batch is uploaded to S3 — creating a silent data-loss window that defeats the ADR's Kafka-replay guarantee.

ansible/roles/grafana/templates/dashboards/hl7-listener-dashboard.json.j2
```diff
@@ -0,0 +1,1082 @@
+{% raw %}
```
andylassiter: `{% raw %}` and `{% endraw %}` in Grafana dashboards were replaced in #259.

ansible/roles/grafana/templates/dashboards/kafka-strimzi-dashboard.json.j2
```diff
@@ -0,0 +1,624 @@
+{% raw %}
```
andylassiter: Same #259

ansible/roles/hl7-listener/defaults/main.yaml
kathrynalpert: Good call, done

hl7-listener/src/main/resources/camel/batcher.camel.yaml
kathrynalpert: Good catch, robot!

hl7-listener/src/main/resources/camel/batcher.camel.yaml
kathrynalpert: Went with Kafka batching consumer, see ADR updates.

andylassiter (approved):

hl7-listener/src/main/java/edu/washu/tag/hl7listener/Hl7BatchZipper.java
```diff
@@ -0,0 +1,80 @@
...
+        try (ZipOutputStream zip = new ZipOutputStream(buffer)) {
+            for (Exchange child : batch) {
+                Message message = child.getMessage();
+                String key = message.getHeader("kafka.KEY", "unknown", String.class);
+                byte[] body = message.getBody(byte[].class);
+                if (body == null) {
+                    body = new byte[0];
+                }
+                zip.putNextEntry(new ZipEntry(key + "-" + child.getExchangeId() + ".hl7"));
+                zip.write(body);
```
johnflavin: If the body is `null`, should we even write a zip entry for it? Seems to me like we should skip it. Would still need to acknowledge the message with kafka, though.

johnflavin (approved):

PR #486, opened 2026-07-06
fix(dcm4chee): Update dcm4chee dev PACS deployment

johnflavin:
### Product
N/A - this dcm4chee PACS isn't deployed in production, only dev

### Technical
Updates for the `dcm4chee` playbook to make the PACS work properly in today's Scout

- Moves UI ingress to a subdomain. We don't do subpath routing anymore, and this never worked on a subpath and had to be served at the domain root anyway.
- Adds a LoadBalancer to support DIMSE
- Coordinates the arc/postgres/slapd version bump to family 35 with per-image Renovate `allowedVersions` pins

## Type of change
- [x] New feature
- [x] Improvement
- [x] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
The dcm4chee ingress now sits behind Scout's oauth2-proxy edge gate (ADR 0003) — the same hardcoded ForwardAuth stack every other Scout web service uses — so every HTTP path on the subdomain (the Archive UI and the WildFly management console) requires a logged-in, Scout-approved user; the console additionally challenges for its own ManagementRealm credentials on top. dcm4chee is not integrated with Keycloak SSO — this is a pure ingress-layer gate and dcm4chee stays auth-unaware behind it. DIMSE (a raw-TCP LoadBalancer, not the HTTP ingress) is not gated. `dcm4chee_admin_password` is now required (deploy asserts it) and must be vault-encrypted.

##### Appsec
The internal management hop uses `insecureSkipVerify` because that port serves the image's baked-in, unverifiable demo cert; it is a pod-internal hop only, with real TLS terminated at the ingress. The admin secret is created with `no_log: true`.

### Performance
N/A — dev-only PACS deployment configuration.

### Data
The postgres major bump (17 → 18) is not an in-place upgrade and requires wiping existing dcm4chee data; acceptable because dcm4chee is dev-only and reinstalled from scratch.

### Backward compatibility
Adds a new required variable (`dcm4chee_admin_password` when the console is enabled). The postgres major bump breaks data compatibility (fresh install required). Renovate `allowedVersions` regexes lock arc/postgres/slapd to family 35 so `loose` versioning can't bump one image out of step.

## Testing
Console reachability and login were verified end-to-end through the ingress. No automated tests added (dev-only tooling).

Other verifications:
- Ingested a bunch of sample data from `/beegfs/shared`
- Connected XNAT using DQR

## Note for reviewers
dcm4chee is a dev/test PACS and is not part of the main playbook or any prod system, which is why the required-password and public-console tradeoffs are acceptable here.

## Checklist


ansible/roles/dcm4chee/defaults/main.yaml
```diff
@@ -5,6 +5,44 @@ dcm4chee_timezone: '{{ timezone }}'
...
+# https://dcm4chee.<server_hostname>/console instead of WildFly's default non-standard
+# MANAGEMENT_HTTPS_PORT (:9993/console). HAL's native /console + /management paths are routed to
+# the management HTTPS port (9990 always 302s to :9993 absolutely; 9993 serves relative
+# redirects) via a dedicated arc-management service whose Traefik ServersTransport speaks HTTPS
+# to the pod. TLS to the browser terminates at the ingress.
+# NOTE: the dcm4chee image seeds no WildFly management user, so the console will prompt for
+# ManagementRealm credentials that don't exist until one is created (e.g. add-user.sh, or the
+# image's WILDFLY_ADMIN_USER/WILDFLY_ADMIN_PASSWORD env). This flag only fixes reachability.
+# SECURITY: the dcm4chee ingress is not behind oauth2-proxy, so this exposes the WildFly
+# management console on the public subdomain. Acceptable for dev PACS; gate or disable for
```
kathrynalpert: Isn't the whole UI also on the public subdomain? Should that have auth?

ansible/group_vars/all/versions.yaml
```diff
@@ -145,14 +145,24 @@ nexus_helm_chart_version: 5.22.0
...
-dcm4chee_ldap_version: 2.6.10-35.0
+
+# dcm4chee archive stack: arc, postgres, and slapd ship as one "family" (the number after the
+# dash on db/ldap; the minor on arc). They MUST stay on the same family — arc waits on and
+# expects a matching db/ldap — so each pin carries an allowedVersions that locks Renovate's
+# `loose` versioning to the current family (and postgres major). This prevents the kind of
+# out-of-step jump PR #400 made: 2.6.10-35.0 was an unpublished base/family combo that wedged
+# the deploy. Current family: 35 (arc 5.35.x, postgres 18.x-35, slapd 2.6.x-35.x). To move to
+# a new family, bump all three values AND update all three regexes together in one change.
+# renovate: datasource=docker depName=dcm4che/dcm4chee-arc-psql allowedVersions=/^5\.35\.\d+$/
```
kathrynalpert: Claude is flagging this: 

> I don't think `allowedVersions` works here the way you'd expect. It's a `packageRules`-only option, not something a custom regex manager can read from an inline comment. And its presence actually breaks our manager's `matchStrings` (the regex expects \n right after `versioning=`/`extractVersion=`), so arc/postgres/slapd stop matching entirely, meaning they silently drop out of Renovate for both version and CVE PRs.
> 
> Quick check: run the `matchStrings` regex from renovate.json5 against the new comments — the three dcm4chee lines no longer match. Fix is to move these into `packageRules` (like the existing k3s-io/k3s pin) with your same regexes, and strip `allowedVersions=` back out of the comments

.

ansible/roles/dcm4chee/defaults/main.yaml
johnflavin: Yes, and yes. I've added the oauth2-proxy annotations to the ingress so now you have to be logged-in to see any of this.

ansible/group_vars/all/versions.yaml
johnflavin: Good catch. I've fixed.

kathrynalpert (approved):

PR #488, opened 2026-07-07
feat(launchpad): gate MinIO Lake card behind enableMinio flag

kathrynalpert:
### Product
Deployments that use cloud object storage instead of an in-cluster MinIO no longer show a "Lake" card in the launchpad's admin tools that links to a host that doesn't resolve. Deployments running MinIO see no change — the card remains by default.

### Technical
Adds an `enableMinio` chart value (default `true`) that emits `ENABLE_MINIO` into the launchpad container. The Next.js server component (`page.tsx`) reads `ENABLE_MINIO !== 'false'` into a boolean prop, and `HomeClient` conditionally includes the Lake entry in the admin-tools grid. Default `true` preserves existing behavior for any deployment that doesn't set the flag.

## Type of change
- [ ] New feature
- [x] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
No change. This gates the visibility of a navigation card only; it does not grant or restrict access to the MinIO console itself, which remains protected by the existing auth layers.

##### Appsec
No new user input. The flag is a server-side env var read at request time and passed as a boolean prop; it is not reflected into output.

### Performance
None — one additional boolean prop.

### Data
Not applicable.

### Backward compatibility
Fully backward compatible: the chart value defaults to `true`, so existing deployments (including the Ansible role, which doesn't set it) render exactly as before. No API or data-model changes.

## Testing
- `helm lint` passes; `helm template` renders `ENABLE_MINIO` exactly once — `"true"` with default values, `"false"` with `--set enableMinio=false`
- `npm run build` (Next.js production build) passes
- `pre-commit` hooks (prettier, eslint, yaml checks) pass on all changed files

## Note for reviewers
This is the last remaining change from the `demo-gitops` branch that never landed on main (the extractor S3-native work landed separately as #353). Re-applied onto the current `HomeClient.tsx` structure (subdomain URLs as props, SCOUT_ENV/docsUrl plumbing). Once merged, the `demo-gitops` branch can be deleted and downstream deployments can consume `launchpad:latest` instead of the stale `latest-demo-gitops` tag.

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/helm/launchpad/templates/deployment.yaml
+++ b/helm/launchpad/templates/deployment.yaml
@@ -73,6 +73,8 @@ spec:
               value: {{ .Values.enableChat | quote }}
             - name: ENABLE_PLAYBOOKS
               value: {{ .Values.enablePlaybooks | quote }}
+            - name: ENABLE_MINIO
+              value: {{ .Values.enableMinio | quote }}
             {{- if .Values.scoutEnv }}
             - name: SCOUT_ENV
               value: {{ .Values.scoutEnv | quote }}
--- a/helm/launchpad/values.yaml
+++ b/helm/launchpad/values.yaml
@@ -7,6 +7,10 @@ replicaCount: 1
 # Feature flags
 enableChat: false
 enablePlaybooks: false
+# Show the MinIO "Lake" card in the launchpad. Default true so deployments
+# still running an in-cluster MinIO continue to surface the console; sites
+# that have cut over to cloud object storage set this false.
+enableMinio: true
 
 # Environment label shown next to the Scout brand (e.g. "synthetic", "production").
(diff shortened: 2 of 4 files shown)


helm/launchpad/templates/deployment.yaml
```diff
@@ -73,6 +73,7 @@ spec:
               value: {{ .Values.enableChat | quote }}
             - name: ENABLE_PLAYBOOKS
               value: {{ .Values.enablePlaybooks | quote }}
+<<<<<<< HEAD
```
johnflavin: Missed some merge markers here

johnflavin (approved):

Issue #482, opened 2026-07-02
reports table merge is inefficient on re-ingest


cmoore01:
What happens is when we process a batch of `x` reports which have already been ingested before verbatim, we merge into `reports` by overwriting all of the previous rows, where nothing changes other than the update timestamp. This is undesirable behavior for a few reasons:
1. That's a very large merge and can take a while to do.
2. It bloats delta history without a benefit.
3. Probably most significantly, each of those merged reports creates an update that must be processed by all of the derivative tables, which is very expensive.

We could do one of:
1. Much more aggressively: for a row, if `source_file` (the unique ID) already exists in `reports`, skip it entirely. (This would block scenarios where we update what data gets extracted and want to re-ingest)
2. Much more conservatively: if a row matches exactly (except for the updated timestamp) one that's already in `reports`, just skip it. I don't know how expensive this would be to run, but I think it would cut out a LOT of pointless updates and downstream churning.

johnflavin:
Empirical confirmation + concrete implementation options from dev03 testing while validating the base/derive activity split. (I'd opened #491 before spotting this ticket — closed as a duplicate in favor of this one.)

**How this reconciles the two options in the description:** a stored **content hash** keyed on the immutable `source_file` gives option 1's efficiency (skip already-ingested files) *without* its downside (it wouldn't block re-ingesting after you change extraction logic — a changed hash still triggers an update), and it's cheaper/more robust than option 2's full-row compare (one hash column instead of an all-column `IS DISTINCT FROM`, and it handles the dynamically-created per-authority ID columns).

## Evidence (dev03, `10k_varied`, re-ingesting the identical manifest into a throwaway table)

Delta history (`$history.operation_metrics`, i.e. each commit's `commitInfo`):

| version | operation | rowsUpdated | rowsInserted | filesAdded | filesRemoved | changeFilesAdded |
|--:|--|--:|--:|--:|--:|--:|
| 0 | CREATE TABLE | – | – | – | – | – |
| 1 | MERGE (first ingest)  | 0 | 9988 | 1 | 0 | 0 |
| 2 | MERGE (re-ingest) | **9988** | 0 | 1 | **1** | **1** |
| 3 | MERGE (re-ingest) | **9988** | 0 | 1 | **1** | **1** |
| 4 | MERGE (re-ingest) | **9988** | 0 | 1 | **1** | **1** |

The single data file is fully replaced each merge (disjoint filenames, zero overlap):

```
[...]

PR #495, opened 2026-07-08
feat(hl7-transformer): split derivative table creation into new activity (#457)

### LINKED ISSUE #457: Split hl7-transformer temporal activities

### LINKED ISSUE #458: hl7-transformer workflow cancelation can cause pod restart
In the past with the hl7-transformer we experienced odd disconnections between pyspark running in the main process and the forked spark process. We worked around that by catching spark errors and signaling that the pod was unhealthy so k8s would restart it. We noticed that there was a workflow cancelation sent from temporal which arrived during a spark calculation; this caused the python temporal listener to stop the activity which stopped the spark process, causing an error to be received by pyspark, marking the pod unhealthy and triggering a pod restart. We should be able to handle temporal cancelation signals without restarting the pod.

johnflavin:
## Description
Fixes #457 

### Product
Basically no user-facing changes. HL7 ingestion is more resilient.

### Technical
Splits the monolithic ingest-and-derive Delta step into two independently-retryable Temporal activities — `ingest_hl7_files_to_delta_lake` (parse + atomic MERGE into the base table + per-file status writes) and `derive_delta_tables` (derives curated/latest/dx/mapping from the base table's change data feed) — orchestrated base → `signalRefresh` → derive (#457). A base failure marks files failed and skips derive; a derive failure fails the workflow without touching the base table or marking files failed. Adds a per-invocation `deriveDeltaTablesTimeout` workflow arg (defaults to 480 min), extracts a shared `spark_activity_session` context manager that owns Spark session lifecycle and error/cancellation cleanup, and fixes the base activity's `spark.stop()` error handler.

Also changed how exceptions are handled so if the activity has received a cancel signal from temporal, we don't rethrow / mark the pod unhealthy / restart the pod. This fixes #458.

## Type of change
- [ ] New feature
- [x] Improvement
- [x] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
N/A — no changes to user roles, data visibility, or API surface.

##### Appsec
N/A — internal Temporal activity/worker changes only; no new external input surface.

### Performance
The base table becomes queryable before the longer derivative derivation runs, and derivative failures retry independently and resume incrementally via the CDF rather than re-ingesting. Per-activity timeouts are independently configurable (base and derive each default to 480 min). Verified on dev against a ~10k-report synthetic dataset.

### Data
The base MERGE is a single atomic Delta transaction, so a derive failure leaves the base table untouched and no partial base rows are written. The incremental derive dedupes CDF windows spanning multiple commits to avoid the `DELTA_MULTIPLE_SOURCE_ROW_MATCHING_TARGET_ROW` wedge; unparseable/empty/junk messages are handled per existing per-file error behavior.

### Backward compatibility
No Delta table/data-model changes. `IngestHl7LogWorkflowInput` gains optional `deriveDeltaTablesTimeout` and `createMapping` (both default sensibly). The internal activity contract between the Java workflow (hl7log-extractor) and Python worker (hl7-transformer) changed — the two images must be deployed together.

## Testing
Added Python unit tests (`test_derive.py`): derivative happy path, derivative-failure-does-not-touch-base, cancel-during-derivative fault isolation, incrementality, the CDF dedup/wedge regression, and `spark_activity_session` error/cancel handling; plus Java `IngestHl7ToDeltaLakeWorkflowImplTest` asserting base→signal→derive ordering, base-failure marks files failed + skips derive, and derive-failure does not mark files failed. CI now runs the hl7-transformer pytest suite as its own separate test activity in parallel with `unit-test` to remove these new tests from the single python environment with all the dependencies running all the other unit tests.

Manually verified on dev03: ran the log-extraction workflow on synthetic 10k data, then exercised the child transformer workflow — cancelling mid-derive left the base table's version and row count unchanged and a rerun resumed to completion; a tight derive timeout self-healed via incremental resume.

Additional testing on dev03 for #458 specifically. Verified red/green: before my fix, a cancel signal caused the pod to restart (at least sometimes, it seems like a race condition) and after the fix it did not.

## Note for reviewers
~~This does not fix 458 (cancelation -> spark failure -> pod restart); the error handling logic is moved into a context manager but otherwise left alone.~~ Does fix #458 as well.

This ~~also~~ does not fix 482 (unconditional MERGE of unchanged rows).

## Checklist


extractor/hl7-transformer/tests/test_derive.py
```diff
@@ -0,0 +1,355 @@
...
+    seed_reports(table, [report_row("s3://bucket/a.hl7", filler="ACC1")])
+    _derive(spark, table, create_mapping=False, health_file=health)
+    assert spark.table(f"default.{table}_curated").count() == 1
+
+    seed_reports(table, [report_row("s3://bucket/b.hl7", filler="ACC2")])
+    _derive(spark, table, create_mapping=False, health_file=health)
+    assert spark.table(f"default.{table}_curated").count() == 2
+
+
+@pytest.mark.flaky(reruns=2)
```
kathrynalpert: Nit: Don't love a flaky test, is there anything we can do to make it more reliable?

extractor/hl7log-extractor/src/main/java/edu/washu/tag/extractor/hl7log/workflow/IngestHl7ToDeltaLakeWorkflowImpl.java
```diff
@@ -118,10 +132,25 @@ public IngestHl7FilesToDeltaLakeOutput ingestHl7FileToDeltaLake(IngestHl7FilesTo
...
-        // Signal the view refresh entity workflow
+        // Signal the view refresh entity workflow. Fires right after the base ingest and
+        // its per-file status writes (reverted to the pre-derivative-tables placement), not
+        // behind the long derivative step.
         signalRefreshActivity.signalRefresh(workflowInfo.getWorkflowId());
         logger.info("WorkflowId {} - Signaled view refresh workflow", workflowInfo.getWorkflowId());
 
+        // Derive downstream tables from the base table's change data feed. A failure here
+        // fails this workflow (surfaced by the existing workflow-failure alert) but does
+        // NOT mark files failed — they are already committed to the base table (#457).
```
kathrynalpert, line 143: I think this makes sense - even if reports don't make it to the derivative table, they're still "green" in the ingest status dashboard because they're in the `reports` table. They should come into the derivative tables with the next run of the "derive" activity. 

This means we'll need to attend workflow failures, not just the ingest table, to determine if something's going awry at this step. We have alerting for this, so seems reasonable.

kathrynalpert (approved):
Should we do a full ingest on preprod before merging this, or wait to get #482 and test the whole shebang?

johnflavin:
> Should we do a full ingest on preprod before merging this, or wait to get https://github.com/washu-tag/scout/issues/482 and test the whole shebang?

My preference for the sequence is
1. Merge this
2. Test on preprod / full reingest
3. Turn to #482, possibly with another reingest

We could flip 2. and 3. if you want, I'm fine either way, but I don't want to wait to merge this until after fixing #482.

extractor/hl7-transformer/tests/test_derive.py
johnflavin: Fixed. The test was flaky because it relied on a `sleep` for timing events happening on other threads, which wasn't reliable. Now we use some threading primitives to verify the derivative activity is running, cancel it, and check that we don't rerun the base ingest activity.

Issue #483, opened 2026-07-02
QA: report-viewer service


andylassiter:
## Scope
QA the new report-viewer service (embedded SPA + backend + OWUI tools) that replaces mcp-trino as the Scout Chat data path.

## Acceptance criteria

**User-facing - Chat**
- [x] Asking to "find" reports opens an embedded viewer inside the chat with a table of matching reports
- [x] Asking for aggregates / counts / distinct values (e.g. "how many CTs in 2024") returns an inline answer, no viewer
- [x] Asking for a specific report by accession / lake path returns the full report content
- [x] Asking to "plot / chart / graph / show the trend / show the distribution" produces an actual chart, not a table
- [x] "Discuss in Chat" from the viewer brings the selected report context into the conversation

**User-facing - Viewer**
- [x] Table renders with sensible default columns; column picker lets you add/remove; columns are resizable
- [x] Filters modal narrows the result set as expected
- [x] Rows expand/collapse to show full report text; hovered rows highlight
- [x] Table pagination works across filters and sorting
- [x] Explain Search shows a plain-language explanation of the search and the SQL generated by the LLM
- [x] CSV export downloads the full result set from the search's SQL (see #535 instead, I think this could be better)
- [x] Expand/contract button in the viewer header toggles the iframe between compact and tall inside the OWUI chat

**Technical - Auth / infra**
- [x] OWUI new-signup webhook sets iframeSandbox UI flags automatically; existing users backfilled
- [x] Per-user Trino authorization honored via service-principal impersonation (ADR 0022) - row filters and PHI masks apply based on the caller's attributes
- [x] Owner-only access enforced server-side on users searches (user a cannot access user b's search)
- [x] Report-viewer metrics flow end-to-end: Prometheus scrapes the service and every panel on the Scout Report Viewer Grafana dashboard has data populating

## Environment
- Preprod cluster with Chat enabled
- Two Scout users with differing `allowed_facilities` / `mask_phi_fields` to exercise per-user Trino AuthZ in Chat

## Additional context
- Parent feature: #463
- Related issue: #471
- Related ADRs: 0022 (Trino identity propagation), 0020/0021 (OPA AuthZ), 0003/0012 (oauth2-proxy + security headers)
- Related QA issue for the AuthZ layer this depends on: #464

cmoore01:
What are the flags mentioned in "OWUI new-signup webhook sets iframeSandbox UI flags automatically"?

andylassiter:
[...]

Issue #471, opened 2026-06-29
Scout Chat does not readily generate the Vega-Lite charts


andylassiter:
Gemma4 produces json but doesn't create the chart. When nudged:

> The user wants me to "Display the charts".
Looking at the previous turn, I provided a JSON object that looks like a chart configuration, but it wasn't in the required vega code block format. The prompt instructions explicitly state: "If the user asks for a chart, return Vega-Lite JSON in a ```vega code block."
I previously returned a generic JSON structure that was not Vega-Lite. I need to convert the data I already retrieved (the distribution of patients by age for whole body MRIs) into a valid Vega-Lite specification and wrap it in a ```vega code block.

GPT-OSS tries to use an external renderer:

> Below is a ready‑to‑use chart (generated via QuickChart) that visualises the same data.
![Patients‑by‑Age Bar Chart]((external link removed for security)(54%2C%20162%2C%20235%2C%200.6)%27%2CborderColor%3A%27rgba(54%2C%20162%2C%20235%2C%201)%27%2CborderWidth%3A1%7D%5D%7D%2Coptions%3A%7Btitle%3A%7Bdisplay%3Atrue%2Ctext%3A%27Distribution%20of%20Patients%20by%20Age%20%28Whole%20Body%20MRI%2C%20Last%205%20Years%29%27%7D%2Cscales%3A%7BxAxes%3A%5B%7BscaleLabel%3A%7Bdisplay%3Atrue%2ClabelString%3A%27Age%27%7D%7D%5D%2CyAxes%3A%5B%7BscaleLabel%3A%7Bdisplay%3Atrue%2ClabelString%3A%27Number%20of%20Patients%27%7D%2Cticks%3A%7BbeginAtZero%3Atrue%7D%7D%5D%7D%7D%7D)
If the image does not render, you can open the URL directly in a browser.

When nudged:
> The user wants to display the chart. We need to provide a Vega-Lite chart using the data. We should not use external URLs. Use vega-lite JSON in a ```vega code block. Provide the chart spec.

andylassiter:
I tried to address this a part of #501 with prompt updates that nudge the model toward Vega-Lite but it still slips up. The real fix is probably a proper plotting feature.

andylassiter:
[...]

PR #501, opened 2026-07-09
feat(report-viewer): add embedded viewer service to chat

### LINKED ISSUE #463: Add an embedded report viewer tool to Chat
Add a report viewing tool to Chat that renders search results from Trino in an embedded iframe instead of returning their contents to the LLM as tool output. Today the LLM has to pull report rows into its context to show them to the user, which bloats context window very quickly. An iframe lets the user see the actual search results, while the report stay out of the context window.

andylassiter:
### Product
Adds a new report-viewer service that replaces mcp-trino as Scout Chat's data path. When a user asks Scout Chat to find reports, the LLM calls into report-viewer and renders results in an embedded iframe inside the chat, instead of pulling report rows into the LLM's context window. Aggregate questions stay inline. Single-report reads return the full report content in the chat.

The viewer is a small React SPA hosted alongside a Python FastAPI backend, framed by OWUI. Users can filter, sort, resize columns, expand rows to see the full report text, download the search as CSV, and pull selected reports into the chat context.

### Technical
- New service under `report-viewer/` (FastAPI + React SPA), Helm chart at `helm/report-viewer/`, Ansible role at `ansible/roles/report_viewer/`
- New OWUI tool `scout_report_viewer_tool.py`
- Trino identity propagation via service-principal impersonation per ADR 0022 (`report_viewer_svc` + `X-Trino-User`)
- JWT audience validation, `aud` pinned to `report-viewer`
- Ingress gated by oauth2-proxy + security-headers, external surface restricted to `/api` and `/spa`
- yoyo migrations for the report-viewer Postgres schema
- mcp-trino removed; Trino now JWT-only (`http-server.authentication.type=JWT`)
- Grafana dashboard + Prometheus scrape
- ADR 0029 documents the design; ADR 0022 updated to capture the new pattern
- OWUI system prompt rewritten around the new three-tool interface
- OWUI signup webhook auto-sets the iframeSandbox UI flags per user

## Type of change
- [x] New feature

##### Authorization
Per-user Trino AuthZ via `X-Trino-User` impersonation (ADR 0022), so OPA row filters and PHI masks still apply. Search access is owner-scoped in Postgres. The OWUI tool only forwards the caller's access token, no refresh token, no client-set impersonation.

##### Appsec
JWT validated (aud + iss). Trino params bound, identifiers allowlisted. External surface locked to `/api` and `/spa` behind oauth2-proxy. NetworkPolicy deny-default; pod runs non-root, RO rootfs, caps dropped. OWUI settings backfill role is column-scoped GRANT only.

### Performance
Not blazing fast. Searches run one sample + one count query, reads paginate the saved SQL as a subquery (no materialization). Queries through `reports_latest_epic_view` are on the slower side, and single-report reads pull through `reports_latest_epic_view` to get resolved patient IDs.

### Data
Flow: LLM authors SQL, report-viewer runs it through Trino (impersonating the caller) to validate + count + sample, then persists the search to its own Postgres. Report rows themselves live in Delta Lake and are re-queried through Trino on every read; nothing about the rows is cached or copied out. Report-viewer stores per search in its `searches` table: the SQL, the LLM's plain-language explanation, cached row count, the user, and optional match_terms/match_diagnoses. The OWUI signup webhook additionally writes iframe-sandbox UI defaults into OWUI's Postgres (would like to remove this when upgrading OWUI to 0.10.x if possible).

### Backward compatibility
mcp-trino is removed and Trino no longer accepts password auth (JWT-only).

## Testing
Backend unit tests under `report-viewer/tests/`, including JWT audience/iss regressions and Trino client cache single-flight. Frontend TypeScript typechecks via `tsc --noEmit`. New Playwright test at `tests/auth/tests/owui-iframe-seed.spec.ts` covers the OWUI iframe-sandbox seeding path. Deployed and exercised on the dev cluster. QA issue #483 tracks the full acceptance criteria checklist.

## Note for reviewers
Sorry for the big PR!

#471 (Vega-Lite charts) not solved. Prompt tweaks nudge the model toward Vega-Lite but it still slips up. The real fix is probably a proper plotting feature but keeping Vega-Lite for now until we have a better solution.

Closes #463
Related to #483 (QA)
Related to #471

## Checklist


report-viewer/src/scout_report_viewer/routes/owui_webhook.py
```diff
@@ -0,0 +1,100 @@
...
+    raw_user = body.get("user") or "{}"
+    try:
+        user_obj = _json.loads(raw_user) if isinstance(raw_user, str) else raw_user
+    except Exception:
+        user_obj = {}
+    new_user_id = user_obj.get("id") or body.get("user_id")
+    new_user_email = user_obj.get("email") or body.get("email")
+
+    if not new_user_id:
+        log.info("webhook fired with no user id", extra={"action": action})
```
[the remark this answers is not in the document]
andylassiter: False positive. `action` is scrubbed via scrub_for_log where it is unpacked from the webhook body (strips CR/LF), and it lands in extra={} which the JsonFormatter JSON-escapes. Field is needed: log fires only when a webhook arrives without a user_id, and `action` tells us which OWUI event dropped the field.

docs/internal/adr/0019-trino-rw-instance-for-views.md
```diff
@@ -12,7 +12,7 @@ Trino is the shared query layer for several Scout services — not just one user
 
 - **Superset**: SQL Lab text editor, virtual datasets, charts and dashboards
 - **Voilà / Playbooks**: notebooks served as web apps, querying Trino directly
-- **Open WebUI / MCP Trino tool**: natural-language SQL via chat (when the optional chat feature is enabled)
+- **Open WebUI / report-viewer tool**: natural-language SQL via chat, routed through the report-viewer microservice (when the optional chat feature is enabled)
```
johnflavin: Shouldn't amend ADRs. They're intended to be immutable once `Accepted`. You can add a new header at the top that mentions of "MCP Trino" are superseded by ADR-0029.

docs/internal/adr/0020-trino-authz-architecture.md
```diff
@@ -38,7 +38,7 @@ This ADR picks the authorization engine and the policy data model. The distribut
...
+| Approval gate | The `user_enabled` rule requires both (a) `enabled: true` on the user's bundle entry (from Keycloak's `UserModel.isEnabled()`) AND (b) membership in `scout-user` or `scout-admin`. The approved-group set is hardcoded in `policy/trino/main.rego` because the same group names are hardcoded in the Keycloak realm template (`ansible/roles/keycloak/templates/scout-realm.json.j2`) — same team, same change. The group check exists because Keycloak federates from upstream IdPs: a user can land in the realm with `enabled=true` purely from a successful upstream OIDC login, before going through Scout's approval flow that adds them to `scout-user`. Service principals (`superset_svc`, `voila_svc`, `report_viewer_svc`) bypass the gate via the `is_system_identity` carve-out since they never appear in `data.users`. |
```
johnflavin: Same as 0019, just add a brief note in the headers.

docs/internal/adr/0022-trino-auth-and-impersonation.md
johnflavin, line 1: This ADR has some pretty substantial changes. I do think the approach should still be to make a note in the header that parts are superseded by another ADR elsewhere, but crafting that message might be tricky given how extensive the changes are.

helm/report-viewer/values.yaml
```diff
@@ -0,0 +1,103 @@
...
+  tag: latest
+  pullPolicy: IfNotPresent
+
+imagePullSecrets: []
+nameOverride: ''
+fullnameOverride: ''
+
+serviceAccount:
+  create: true
+  automount: true
```
johnflavin: Nit: I don't think we need to auto mount the service account creds. (I think this is probably the default with a new helm chart and we likely do it elsewhere so maybe ignore this.)

report-viewer/src/scout_report_viewer/routes/owui_webhook.py
```diff
@@ -0,0 +1,101 @@
...
+    except Exception:
+        user_obj = {}
+    new_user_id = scrub_for_log(user_obj.get("id") or body.get("user_id"))
+    new_user_email = scrub_for_log(user_obj.get("email") or body.get("email"))
+
+    if not new_user_id:
+        log.info("webhook fired with no user id", extra={"action": action})
+        metrics.OWUI_WEBHOOK_EVENTS.labels(action="other", result="skipped").inc()
+        return None
+    if action and "signup" not in action.lower() and "new" not in action.lower():
```
johnflavin: This branch should also trip if `action` is not set. Currently this only filters out requests where `action` _is_ set but doesn't have one of the allowed values.

report-viewer/src/scout_report_viewer/routes/searches.py
```diff
@@ -0,0 +1,837 @@
...
+    where_parts: list[str] = []
+    filter_params: list = []
+    for fcol, fvals in filters:
+        clause, p = _filter_clause(fcol, fvals, alias="s")
+        where_parts.append(clause)
+        filter_params.extend(p)
+    where_sql = (" WHERE " + " AND ".join(where_parts)) if where_parts else ""
+    if sort_spec:
+        scol, sdir = sort_spec
+        order_sql = f" ORDER BY s.{_quote_ident(scol)} {sdir} NULLS LAST"
```
johnflavin: Claude flags that this branch doesn't have a stable default like the other branch does.

> Unstable pagination under user sort. With an explicit sort, the query is `ORDER BY s.<col> <dir> NULLS LAST` with no unique tiebreaker — `OFFSET`/`LIMIT` paging over a non-unique column (`sex`, `modality`, `message_dt`, …) can skip or duplicate rows across pages. Default order (`primary_report_identifier`) is fine. Fix: append `, s.primary_report_identifier`.

ansible/roles/report_viewer/templates/values.yaml.j2
```diff
@@ -0,0 +1,69 @@
...
+    - podSelector:
+        matchLabels:
+          app.kubernetes.io/name: open-webui
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: '{{ prometheus_namespace }}'
+      podSelector:
+        matchLabels:
+          app.kubernetes.io/name: prometheus
+          app.kubernetes.io/component: server
```
johnflavin, line 59: This was flagged by Claude as breaking some of the user impersonation safety claims. I'm not sure that it is a big deal, and its response even says maybe we should just accept and document this, but I do think it is correct that this opens a small hole where internal requests don't have the same safety as external requests.

> The header-trust path is wider than "Traefik only" (PHI-mask bypass)
> `auth.py` path 2 trusts the `X-Auth-Request-Preferred-Username` header _verbatim_ and forwards it as `X-Trino-User`. The safety claim (in `auth.py`:8-10, `networkpolicy.yaml`:2-5, and ADR 0029 §47) is "NetworkPolicy restricts this to Traefik so it can't be forged." As deployed that's inaccurate: the rendered `NetworkPolicy` (`ansible/roles/report_viewer/templates/values.yaml.j2`:47-57) admits the OWUI pod (unscoped `open-webui` selector) and the Prometheus server on `:8000` — the same port serving `/api`. Any code executing in those pods can `POST` to `/api/...` with a forged username header, no bearer token, and read that user's OPA-scoped reports.
> - External forgery is closed — I confirmed the `oauth2-proxy-auth forwardAuth` overwrites the client header (`authResponseHeaders`). The residual risk is purely the in-cluster non-Traefik peers, which never traverse that middleware.
> - Prerequisite is code-exec in an OWUI/Prometheus pod — but OWUI runs model-/tool-driven Python, and a metrics scraper asserting user identity to a PHI service is the weakest link.
> - Decision to make: gate the header path behind a Traefik-injected shared-secret header, split `/metrics` onto a separate port, or explicitly accept+document that the OWUI/monitoring namespaces are inside the trust boundary. `report-viewer` can't distinguish a Traefik-origin request from an OWUI-origin one at L7 today.

report-viewer/src/scout_report_viewer/routes/searches.py
```diff
@@ -0,0 +1,837 @@
...
+        "epic_mrn",
+        "mpi",
+        "message_dt",
+        "modality",
+        "service_name",
+        "sending_facility",
+        "patient_age",
+        "sex",
+    }
+)
```
johnflavin: I think this list should match what's in `models.py`.

> Sort/filter whitelist exceeds guaranteed projection. `_SORTABLE_COLUMNS`/filters span 9 columns, but only `primary_report_identifier` + `accession_number` are enforced as projected (`models.py`:39-42). A minimal saved search 502s (`COLUMN_NOT_FOUND`) when the user sorts/filters on any other column. Works today only because the LLM voluntarily projects the standard set — fragile contract.

johnflavin (approved):

report-viewer/src/scout_report_viewer/auth.py
```diff
@@ -0,0 +1,139 @@
+"""Auth - two paths, first match wins.
+
+1. **Bearer JWT** validated against Keycloak JWKS. Used by the OWUI tool
+   path (forwards `__oauth_token__`) and by anything else that wants to
+   present a real end-user token. Validates signature + exp + iss + aud
+   (`aud=report-viewer`, stamped by the `report-viewer-audience` client
+   scope).
+2. **oauth2-proxy header** (`X-Auth-Request-Preferred-Username`) - the
+   ingress path. The NetworkPolicy restricts ingress to Traefik so the
+   header can't be forged from inside the cluster.
```
kathrynalpert: It could from within OWUI itself, right? Not saying we need to close that hole, but we should be clear about the risk (esp if we get to a point of wanting ACE in OWUI).

helm/report-viewer/values.yaml
```diff
@@ -0,0 +1,103 @@
+# Default values for report-viewer.
+
+replicaCount: 1
+
+image:
+  repository: ghcr.io/washu-tag/report-viewer
+  tag: latest
+  pullPolicy: IfNotPresent
```
kathrynalpert, line 8: Nit (bc I'm sure we do this all over the place): unspecified `pullPolicy` + `latest` will always pull, but I think by specifying `pullPolicy: IfNotPresent`, we lose that "benefit" of `latest` versions. I don't know that we care, maybe we even prefer `pullPolicy: IfNotPresent`. Just wanted to mention how it works because it's a bit of a confusing combination to specify.

docs/internal/adr/0029-report-viewer.md
```diff
@@ -0,0 +1,149 @@
...
+
+Splitting the surface this way keeps the intent explicit for the LLM: cohort building persists, lookup and analytics do not. Collapsing all three into one endpoint would either force persistence on every operation or make the LLM signal intent through arguments the researcher never sees.
+
+### Just-in-time cohort evaluation
+
+A search is a saved SQL query plus minimal metadata. Nothing about which rows match is stored. Every read wraps the saved `sql` as a subquery and applies pagination, sort, and filter at the Trino layer. The `searches` table holds `id`, `owner_sub`, `id_column`, `sql`, `sql_explanation`, `match_terms`, `match_diagnoses`, `row_count` (cached at create time via one `SELECT COUNT(*)`), `owui_chat_id`, and `created_at`. Schema migrations are yoyo. Query refinement is not tracked with a `parent_id` column. When a user narrows a cohort, the LLM is instructed to re-emit full SQL for the new search and the old search is unchanged.
```
kathrynalpert: I'm not familiar with `yoyo`, I thought it was an acronym. Can you make clear that it's a tool (link or something)?

docs/internal/adr/0029-report-viewer.md
```diff
@@ -0,0 +1,149 @@
...
+|---|---|---|
+| `scout_find_reports` | `POST /api/searches` | Yes, as a saved search |
+| `scout_get_reports` | `POST /api/reports/read` | No, transient |
+| `scout_query_sql` | `POST /api/reports/query` | No, transient |
+
+`scout_find_reports` accepts SQL plus metadata (`sql_explanation`, `match_terms`, `match_diagnoses`) emitted by the LLM, saves it as a named search, and returns a sample of rows plus evidence context. The resulting search is browseable in an iframe above the LLM reply and supports CSV export. The other two tools execute Trino reads and return rows directly, without persisting a search row or rendering an iframe.
+
+Splitting the surface this way keeps the intent explicit for the LLM: cohort building persists, lookup and analytics do not. Collapsing all three into one endpoint would either force persistence on every operation or make the LLM signal intent through arguments the researcher never sees.
+
+### Just-in-time cohort evaluation
```
kathrynalpert, line 32: Maybe add some commentary about why and the problem we're punting here.

docs/internal/adr/0029-report-viewer.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Authentication
+
+There are two inbound callers and one outbound Trino pattern.
+
+**Inbound.** The SPA and iframe reach the backend through oauth2-proxy at the ingress (ADR 0003 approval gate). oauth2-proxy forwards identity-only via `X-Auth-Request-Preferred-Username`. No access token is forwarded. The OWUI tool runtime posts in-cluster with `Authorization: Bearer <__oauth_token__>`, the user's Keycloak access token. FastAPI validates the JWT against Keycloak JWKS: RS256 only (pinned allowlist), signature, `exp`, `iss`, and `aud=report-viewer` (stamped by a client scope on the OWUI client). NetworkPolicy restricts Bearer-bearing in-cluster traffic to OWUI pods so the header cannot be forged from elsewhere in the cluster. Both paths resolve to the same `User` model.
```
kathrynalpert: >NetworkPolicy restricts Bearer-bearing in-cluster traffic to OWUI pods so the header cannot be forged from elsewhere in the cluster. 

I don't follow this: Bearer traffic doesn't need network-level protection, it's validated cryptographically (signature, exp, iss, aud). The concern is the unauthenticated `X-Auth-Request-Preferred-Username` header traffic, which NetworkPolicy limits to OWUI (and Prometheus). 

Discussing this gap (OWUI+Prom could forge `X-Auth-Request-Preferred-Username` traffic) explicitly is probably worth doing, as I think it does have implications for OWUI especially (one example: allowing users to add tools would allow them to run code server-side, where they could pass arbitrary headers). Claude is suggesting a "shared secret" fix, wherein traefik and report-viewer share a secret that OWUI doesn't know, so legit traffic through traefik gets the secret injected, where pod-to-pod traffic from OWUI to report-viewer wouldn't have it (bearer-tokened traffic wouldn't need the secret, but trusted-header traffic would).

report-viewer/src/scout_report_viewer/routes/searches.py
```diff
@@ -0,0 +1,837 @@
...
+            # safe: LLM-authored SQL wrapped as subquery, OPA is the AuthZ boundary
+            # nosemgrep: python.sqlalchemy.security.sqlalchemy-execute-raw-query.sqlalchemy-execute-raw-query
+            _cols, count_rows = await trino_client.execute(count_sql, user=user.sub)
+        row_count = int(count_rows[0]["n"]) if count_rows else 0
+    except Exception as exc:
+        log.exception("trino count query failed")
+        # Non-fatal: search still gets created, just without cached
+        # count. Reads will report 0 until the next refresh, which is
+        # better than failing the whole create.
+        row_count = 0
```
kathrynalpert: What leads to a "refresh"? What are the implications of `row_count=0` on the UI?

report-viewer/src/scout_report_viewer/routes/searches.py
```diff
@@ -0,0 +1,837 @@
...
+    search_id: str,
+    user: User = Depends(get_current_user),
+) -> StreamingResponse:
+    ds = await store.get_search(search_id, owner_sub=user.sub)
+    if ds is None:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND)
+    sql = ds["sql"]
+    uploaded_ids = ds.get("uploaded_ids")
+
+    async def gen():
```
kathrynalpert, line 788: Claude flagged performance here:
>CSV export is O(n²) on Trino: each 500-row chunk re-runs the full query with ORDER BY … OFFSET. A 100k-row export is 200 full sort-scans. The trino dbapi cursor streams — one query + fetchmany loop is both simpler and much faster. Also: _csv_quote doesn't neutralize Excel formula injection (=, +, @ leading cells from report text).

report-viewer/src/scout_report_viewer/db.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+_pool: AsyncConnectionPool | None = None
+
+
+async def open_pool() -> AsyncConnectionPool:
+    """Lazy-init a process-wide AsyncConnectionPool."""
+    global _pool
+    if _pool is None:
+        _pool = AsyncConnectionPool(
```
kathrynalpert: > db.py pool init race: open_pool() assigns the global before await _pool.open(...) completes, so a concurrent request can grab a not-yet-open pool and get PoolClosed; and if the first open() raises, _pool stays assigned — worth verifying the service actually recovers when Postgres is down at first request, and resetting _pool = None on failed open.

docs/internal/adr/0029-report-viewer.md
```diff
@@ -0,0 +1,149 @@
+# ADR 0029: Report Viewer Service for Chat
+
+**Date**: 2026-07-07
+**Status**: Accepted
+**Decision Owner**: TAG Team
+**Supersedes**: Open WebUI's Trino-access parts of ADR 0019, ADR 0020, and ADR 0022 (the MCP Trino tool, openwebui_mcp_svc, and the interim PASSWORD authenticator)
```
kathrynalpert, line 6: Instructions for teardown of existing MCP Trino tool are probably appropriate

CLAUDE.md
kathrynalpert, line 1: Missing ADR 29 (also 28, which was my bad and outside scope of this PR)

docs/internal/adr/0029-report-viewer.md
```diff
@@ -0,0 +1,149 @@
...
+
+OWUI's `message.embeds` iframe sandbox must include `allow-same-origin` and `allow-forms`. `allow-same-origin` is required so the iframe keeps its own origin identity rather than loading as a unique opaque origin. Without it, the SPA cannot use `document.cookie` or storage APIs, and `fetch('/api/...')` on its own backend cannot send the oauth2-proxy session cookie that authenticates the request. `allow-forms` covers SPA form submissions.
+
+In OWUI 0.9.6 these flags are per-user settings with no admin-global override, and both default to `false`, so every new researcher would hit a broken iframe until they toggled the flags manually. To avoid that, report-viewer receives an OWUI signup webhook and seeds the flags on the new user's `settings` row before their first hydrate (see below).
+
+### OWUI new-user webhook
+
+`POST /webhooks/owui-new-user` receives OWUI's admin signup webhook (configured via OWUI's `WEBHOOK_URL` field on install-chat). On a signup event the receiver runs one idempotent `UPDATE "user" SET settings = jsonb_set(...)` against OWUI's Postgres to set `iframeSandboxAllowSameOrigin` and `iframeSandboxAllowForms` to `true`.
```
kathrynalpert: Claude:
>ADR documents jsonb_set while the code deliberately rejected it for SELECT-then-UPDATE.

docs/internal/adr/0029-report-viewer.md
```diff
@@ -0,0 +1,149 @@
...
+
+**Trino auth mode.** Open WebUI no longer calls Trino directly, so the interim `PASSWORD` authenticator from ADR 0022 (`openwebui_mcp_svc`, `password.db.j2`, `authenticationType: PASSWORD,JWT`) is removed. Trino runs JWT-only. Every client, including report-viewer, authenticates with JWT.
+
+**Flows.** Both inbound paths converge on the same `report_viewer_svc` JWT + `X-Trino-User` impersonation call to Trino.
```
kathrynalpert, line 59: A little more explanation here would be helpful - why we have two flows and which actions use which flow. 

I think: the second flow details how the user's chat prompt -> LLM tool call -> report viewer, while the first is how the result from that call can be rendered in the browser. (And two flows are needed bc there's no JWT in browser.)

ansible/roles/open-webui/templates/values.yaml.j2
```diff
@@ -178,6 +178,12 @@ extraEnvVars:
   - name: RAG_EMBEDDING_ENGINE
     value: 'ollama'
 
+  # OWUI's SSRF guard blocks RFC1918 hosts, which would reject the
+  # in-cluster webhook to report-viewer. Opt out via the
+  # upstream-endorsed flag, open-webui/open-webui#24587.
+  - name: ENABLE_RAG_LOCAL_WEB_FETCH
```
kathrynalpert, line 184: I think this is a security hole that would allow users to query other Scout services. Per Claude: 
>`ENABLE_RAG_LOCAL_WEB_FETCH=true` turns off OWUI's SSRF guard everywhere, so any logged-in chat user can point OWUI's built-in URL fetch (`/api/v1/retrieval/process/web`) at internal endpoints like Prometheus or Temporal and pull the response back into chat.

kathrynalpert (commented):
~~Reminder to make corresponding inventory PR!~~ (Sorry, you already did this, and linked it, and I just missed it)

Couple other small Claude findings, defer to you on whether they're worth fixing:
  - Readiness probes /healthz, which never checks the DB — pod reports Ready while every request 500s; the app.py comment even claims the probe distinguishes this.
  - python-jose has no >=3.4 floor (CVE-2024-33663/64 fixed there); consider PyJWT long-term since the code already hand-patches jose's aud quirk. And psycopg2-binary exists only for yoyo — postgresql+psycopg:// in the migration backend URL drops it.
  - Frontend minors: dead scroll-into-view (e.currentTarget read inside rAF is null), "open chat ↗" navigates to a 404 when chat_origin is unset, friendlyError throws on empty string.
  - Nits: ids.py mentions a "TTL" that doesn't exist; csv_upload.py docstring describes an inlined IN (...) that's actually bound; the new top-level import bcrypt in bootstrap.py defeats its own deliberate lazy import; Dockerfile user searchs typo;  resolved_id_column = scrub_for_log(...) mutates a value that then flows into the DB (no-op today, but scrub at the log site instead).

report-viewer/src/scout_report_viewer/db.py
johnflavin: I think the solution here would be to first assign to a temp `pool` and only assign `_pool = pool` when everything is done. But then `_pool` is still unassigned while you’re constructing the pool, so someone else could enter this block. Which means... 🥁 double checked locking to the rescue again! (Or maybe better concurrency primitives instead of a `global _pool`.)

ansible/roles/open-webui/templates/values.yaml.j2
andylassiter: This was required for the webhook to work. But we are dropping the webhook with the OWUI 0.10.x upgrade and I removed this already in that branch. Going to leave as is here.

ansible/roles/report_viewer/templates/values.yaml.j2
andylassiter: Kate and you both flagged this. I went with the Traefik-injected shared-secret header approach ([a73db26](https://github.com/washu-tag/scout/pull/501/commits/a73db26a2b5ff1a2316e18407f32114943966d76)).

CLAUDE.md
andylassiter: I added ADR 29.

docs/internal/adr/0029-report-viewer.md
andylassiter: Flavin also flagged this. I addressed in [a73db26](https://github.com/washu-tag/scout/pull/501/commits/a73db26a2b5ff1a2316e18407f32114943966d76) with a Traefik-injected shared-secret header.

report-viewer/src/scout_report_viewer/auth.py
andylassiter: Updated alongside [a73db26](https://github.com/washu-tag/scout/pull/501/commits/a73db26a2b5ff1a2316e18407f32114943966d76).

.pre-commit-config.yaml
```diff
@@ -27,6 +27,8 @@ repos:
           - eslint-config-next@15.0.0-rc.1
           - next@15.0.0-rc.1
           - typescript-eslint@8.9.0
+          - typescript@5.6.3
+          - ts-api-utils@1.4.3
```
johnflavin, line 31: We need this fix on `main`. Other branches are failing the lint job. If you need to delay merging this branch for whatever reason, I'll pull it in on a new PR, but I'd prefer to just merge this one.

helm/report-viewer/values.yaml
andylassiter: > bc I'm sure we do this all over the place

Yeah we are, I'm going to leave as is for consistency.

report-viewer/src/scout_report_viewer/db.py
andylassiter: > Or maybe better concurrency primitives instead of a global _pool

Claude started inventing its own pattern here, I went with your second suggestion instead and dropped the global in 692ec99.

.pre-commit-config.yaml
andylassiter: I've addressed everyone's comments but hadn't finished replying. I'll merge then finish with the comments.

report-viewer/src/scout_report_viewer/routes/searches.py
andylassiter: "refresh" was stale Claude wording, but the UX around 0 results wasn't great, the viewer would still be shown. Fixed in 09c96fa. The OWUI tool won't display the report viewer table if 0 rows are returned.

report-viewer/src/scout_report_viewer/routes/searches.py
andylassiter: Addressed in e393f3a. Sort/filter columns moved into models.py.

PR #507, opened 2026-07-13
Deploy data-generator from published OCI chart

cmoore01:
Swap the local helm/data-generator chart for the published OCI artifact at ghcr.io/washu-tag/charts/data-generator. Dev override via data_generator_chart_path mirrors the xnat pattern.

### Product
N/A

### Technical
Hopefully makes maintenance easier by removing custom helm chart and pointing at published OCI. Publishing added in https://github.com/washu-tag/data-generator/commit/a92faf7ea4f81ff246700fee31958a934692e6f7 

## Type of change
- [ ] New feature
- [ ] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
N/A

##### Appsec
N/A

### Performance
N/A

### Data
N/A

### Backward compatibility
Should be backwards compatible

## Testing
Successfully deployed to dev01

## Checklist


ansible/roles/data-generator/tasks/deploy.yaml
```diff
@@ -12,10 +12,13 @@
     tasks_from: deploy_helm_chart
   vars:
     helm_chart_name: data-generator
-    helm_chart_ref: '{{ data_generator_helm_chart_ref }}'
+    # Pull the published OCI chart from GHCR (helm runs on the jump node via the
```
kathrynalpert, line 15: Do we need to run this on the jump node now? Ideally, this would be pulled through harbor.

kathrynalpert (approved):

ansible/roles/data-generator/tasks/deploy.yaml
cmoore01: Wouldn't that also be true on the existing XNAT chart pull? I don't think we should be running this playbook in prod, but we would for XNAT, which mentions:
```
**Chart**: the upstream chart is published as an OCI artifact
   (`oci://ghcr.io/nrgxnat/charts/xnat`, `xnat_chart_oci_ref`),
   pinned to `xnat_chart_version` (`group_vars/all/versions.yaml`). Helm pulls it
   directly from GHCR on the jump node — the `deploy_helm_chart` wrapper delegates
   to localhost for `oci://` refs.
```

If needed, we could bring in whatever solution is done for xnat first, but that's where it would start, right?

johnflavin: I thought we were able to pull OCI images through harbor already, but Charlie (and the comments) are correct that we currently cannot. We'd need to add that in the helm deploy task. To sketch it out: we would probably add an argument to that task for oci chart urls, and when given one we would have to manually modify it to add the proxy url when air gapped (because helm can't transparently proxy oci charts, see helm/helm#13615, though there is a proposal for it in helm/community#391).

Issue #517, opened 2026-07-14
Complete access revocation — purge a user's saved data and refresh their cached results


kathrynalpert:
**Problem**
Lowering a user's access only affects future queries. Data produced or saved while they had
broader access can linger: results they saved in notebooks or chat history, and Superset's
cached query results. So right after an access reduction, a user can still come across data
they should no longer see. (Cross-user leakage and general staleness are handled separately
in #502; this is the completeness pass for the person whose access was reduced.)

**Proposed solution**
Give admins a deliberate way to close the loop when they reduce someone's access — an
admin-triggered action, not automatic on every change:
- Purge that user's saved artifacts: notebook outputs and chat history.
- Clear or invalidate that user's Superset cached results.

For the Superset cache there are two ways to go, to decide during design:
- Actively clear the user's cached entries as part of the purge, or
- Key the cache on the user's access (their facilities + masking) so an access change makes
  old entries unreachable on its own — which also lets users with identical access share
  cache. (This is the longer-term cache refinement noted on #502.)

**Additional context**
Only relevant when reducing access, which is rare compared to granting it. Pairs with the
admin warning that access changes aren't retroactive (#515). Part of #464.

PR #518, opened 2026-07-14
fix(superset): per-user result cache and SQL Lab export ownership

### LINKED ISSUE #502: Superset dashboards can show one user's data to another
**Expected behavior**
On every dashboard and chart, a user sees only the rows and columns their access allows.

**Actual behavior**
Superset caches chart and dashboard results without tying the cached copy to the viewer.
Whoever loads a chart first has their results cached and handed to everyone who opens it
afterward, until the cache expires (up to 2 hours today). So a user with narrower access can
be shown another user's rows and unmasked columns, and someone whose access just changed
keeps seeing stale results until they force a refresh. A shared dashboard built on a
user-defined ("virtual") dataset leaks the same way.

**Steps to reproduce**
1. As a user with broad access, open a dashboard so its results are cached.
2. As a user with narrower access, open the same dashboard.
3. The second user sees the first user's data until a manual refresh or cache expiry.

**Suggested fix**
Enable Superset's per-user result caching (the `CACHE_QUERY_BY_USER` feature flag), so cached
results are never shared between users, and shorten the cache lifetime on the sensitive
datasets so changes take effect more quickly.

A longer-term refinement (keying the cache on the viewer's actual access so an access change
automatically invalidates their old entries) is folded into the access-revocation follow-up
(#517).

Reported from #464.

### LINKED ISSUE #509: SQL Lab CSV export can be downloaded by other users
**Expected behavior**
Downloading a SQL Lab query result returns only your own results.

**Actual behavior**
The SQL Lab CSV export endpoint (`/api/v1/sqllab/export/<id>`) checks that the caller can use
the database, but not that they own the query. Since everyone shares the same database
connection, any signed-in user can download another user's results by supplying that query's
short, guessable ID. Because the stored results are returned, the download also reflects the
masking that applied to the original runner, not the caller.

**Steps to reproduce**
1. Run and export a query in SQL Lab as one user; note the export URL.
2. As a different user, request the same export URL.
3. The second user receives the first user's results.

**Suggested fix**
Add an ownership check: extend Superset's query access check so the export only returns
results for queries the requesting user ran (admins with full query access excepted).

(Also report the missing check upstream to Superset as a bug.)

**Note**
A sibling endpoint (`/api/v1/sqllab/results/`) that fetches results by an internal key has no
access check at all, but that key is a random UUID rather than a guessable ID, so the
practical risk is probably low; we'll document it and report it upstream rather than adding a separate guard now.

Reported from #464.

### LINKED ISSUE #519: scout-dashboards chart changes don't deploy via install-superset (helm module skips local-file edits)
## Expected behavior
Editing the scout-dashboards chart — a dashboard, a dataset YAML, or `files/import.py` — and running `make install-superset` deploys the change: the release upgrades and the import job re-runs.

## Actual behavior
`make install-superset` reports the scout-dashboards release as unchanged and doesn't upgrade it, so file-only edits never reach the cluster. On dev04 the release sat at revision 7 for ~35 days across multiple deploys; the `dashboard-config` ConfigMap and `import.py` were 41 days old. As a result, per-dataset `cache_timeout` edits (and any other dashboard/dataset changes made in that window) silently never applied.

## Steps to reproduce
1. Edit a file under `helm/scout-dashboards/` (e.g. a dataset YAML, or `files/import.py`).
2. Run `make install-superset`.
3. The "Install scout-dashboards" task reports `changed: false` and the release revision doesn't advance; the change isn't live.

## Environment
Any deploy of the superset role. `kubernetes.core.helm` installing the local scout-dashboards chart, whose version is frozen at `0.0.0-dev`.

## Additional context
Root cause: `kubernetes.core.helm` decides whether to upgrade a local chart from the chart version + the passed values, not the rendered templates. With the version pinned at `0.0.0-dev` and the values static, file-only edits are invisible to the module, so it skips the upgrade — and the chart's own `checksum/config` hook (which would re-run the import) never fires because Helm never re-renders.

Fix: pass a checksum of the chart's files as a Helm value so the module sees a change and runs the upgrade on any file edit.

Found while deploying the Trino AuthZ QA fixes (#464); fixed in #518.

kathrynalpert:
### Product
Closes two ways Superset could show a user data they shouldn't see, plus a deploy gap found along the way:
- Dashboards and charts were served from a cache that wasn't tied to the user identity, so a user with narrower access could be shown a broader-access user's rows and unmasked columns until the entry expired (#502).
- The SQL Lab CSV export endpoint didn't check who owned the query, so a signed-in user could download another user's results using its short, guessable id (#509).
- The dashboards chart wasn't actually re-deploying when its files changed, so dashboard/dataset edits silently never reached the cluster (#519).

### Technical
- **#502:** Trino impersonation happens at DB-connection time, downstream of the results-cache lookup — so a cache hit never re-ran as a different user and never reached the OPA policy. Enable the `CACHE_QUERY_BY_USER` feature flag so the data cache is keyed per user. 
- **#517:** If a given user's access changes, though, they would need a hard-refresh to bypass their own cache. Making this automatic or offering a manual "purge" option to the admin on access changes is beyond the scope of this fix. However, I did modify caching a bit here, so that cache lifetime is set once at the **database level** (`Scout Data Lake` `cache_timeout = 1800`), which is visible and editable in the UI (Edit database → Performance → Chart cache timeout); datasets are shipped with no per-dataset timeout (was previously `7200s` for each table).
- **#509:** The export path (`/api/v1/sqllab/export/<client_id>`) only checked database access, which every user passes on the single shared connection. `KeycloakSecurityManager.raise_for_access(query=...)` now requires the query belong to the current user unless the caller has `all_query_access`.
- **#519:** `kubernetes.core.helm` skips upgrading a local chart on file-only edits (version pinned at `0.0.0-dev`, values static), so `scout-dashboards` changes (including this PR's dataset edits) never deployed. The deploy now hashes the chart files and passes the digest as a value, so the module upgrades — and the import hook re-runs — whenever a file changes. `update_dataset` also now syncs `cache_timeout`, so existing datasets get their stale per-dataset value cleared and inherit the DB default.

## Type of change
- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
Tightens authorization: cached data is per-user, and SQL Lab exports are scoped to the query's owner. Admins (`all_query_access`) keep query history. Not addressed here: `/api/v1/sqllab/results/` also lacks an ownership check, but it's keyed by a server-side UUID rather than a guessable id, so it's documented as a low-risk follow-up.

##### Appsec
N/A

### Performance
Per-user cache keys mean a lower hit rate than a shared cache; the 1800s database default restores cache hits so dashboard loads don't all run live against Trino.

### Data
N/A

### Backward compatibility
Compatible

## Testing
Deployed to dev04. Note the #519 fix was required before the cache-timeout change actually applied. After the fix, redeploying shows the DB "Chart cache timeout" = 1800 and datasets inheriting it. Acceptance checks (two accounts of differing access): (1) a narrower-access user opening a dashboard first populated by a broader-access user sees only their own rows/columns without a force-refresh; (2) a user cannot download another user's SQL Lab export (403) and can still download their own.
Screenshot 2026-07-14 at 9 36 28 AM

Screenshot 2026-07-14 at 8 26 59 AM

## Note for reviewers
- LMK if you think we should attempt to add a check to `/api/v1/sqllab/results/`, too. It's not hard, but it's a little hacky because there isn't a great injection point, so we have to monkey patch a `validate` method to add the check. 

## Checklist

Closes #502
Closes #509
Closes #519


andylassiter (approved):
> LMK if you think we should attempt to add a check to /api/v1/sqllab/results/, too. It's not hard, but it's a little hacky because there isn't a great injection point, so we have to monkey patch a validate method to add the check.

If the risk is low and the fix is hacky I'm okay with skipping for now.

PR #522, opened 2026-07-14
fix(opa): Trino AuthZ policy hardening + masking opt-in + attribute rename

### LINKED ISSUE #510: Patient-mapping table becomes visible if the report table is renamed
**Expected behavior**
The internal table linking patients to reports is always hidden from users and can't be
queried directly without explicit permission on linking tables.

**Actual behavior**
That table is hidden by matching its exact name. The report table name is a configurable
setting, and the mapping table's name is derived from it, but the hidden-list still uses the
default name. So if an operator changes the report table name, the mapping table is created
under the new name, the hidden-list no longer matches it, and it becomes visible in table
listings and directly queryable by any user.

**Suggested fix**
Derive the hidden-list entry from the same setting the mapping table's name is built from
(`report_delta_table_name`), so the two can't drift apart.

Reported from #464.

### LINKED ISSUE #511: Make field masking opt-in (off by default)
**Description**
Today, identifier masking is on by default for new users. It only redacts a couple of
columns (patient name and ZIP); date of birth, medical record number, and the free-text
report body are never masked. So the default gives a misleading impression of broad coverage
and has caused admin confusion, while adding little real protection.

Change the setting to off by default: today the access policy masks unless a user's setting
is explicitly "off", so flip it to mask only when the setting is explicitly "on", and change
the new-user default in the admin console to match. Admins can still turn it on per user
where they want those specific columns redacted.

**Note — reduces default protection.** This is an intentional loosening: masking becomes
something an admin opts into rather than the default. It's worth doing because the current
default oversells what it covers, but it should land as its own reviewed change with the
documentation updated to match. Pairs with the rename/documentation change (#512).

Related to #464.

### LINKED ISSUE #512: Rename the "mask PHI fields" setting and document what the controls cover
**Description**
The admin setting labeled "mask PHI fields" oversells what it does: it redacts a few
identifier columns, not all PHI (date of birth, medical record number, report text, etc. pass
through). Rename the display label to something honest, e.g. "redact select identifier
columns" (consider updating the underlying variable name since this is the first deploy).

While here, add a short written note describing what the data-access controls do and don't
cover: they filter live queries per user; they don't retroactively scrub data a user already
saved elsewhere, and masking is limited to the listed columns. This gives future reviews a
clear reference point.

Pairs with the masking-default change (#511).

Reported from #464.

### LINKED ISSUE #513: Confirm and document behavior for reports with no sending facility
**Description**
Users are restricted to specific sending facilities. Rows that have no sending facility are
currently hidden from facility-restricted users (and visible to full-access users), and a
user with no facilities assigned sees nothing — both of which are the safe, intended
behavior. Add tests to lock this in (including the edge cases of an empty string or a null in
a user's facility list) and document the behavior in the data-access docs, so it can't
regress silently.

Reported from #464.

### LINKED ISSUE #514: Document that a custom report-table name skips facility filtering
**Description**
Facility row-filtering applies to the standard report tables. If an ingest is run against a
differently-named table, that table isn't covered by the filter. This can only happen when
an operator deliberately overrides the table name (the `reportTableName` ingest parameter,
which otherwise defaults to the standard report table), and the ability to do so is useful
for testing, so we don't want to block it, but it should be documented as a convention: in
an environment with real data, don't tweak the `reportTableName` outside of inventory.yaml.

Related to #464.

kathrynalpert:
### Product
Several Trino data-authorization fixes, plus a rename for clarity:
- The internal patient-mapping table stays hidden even if the report table is renamed (#510).
- Identifier masking is now **opt-in (off by default)** instead of on-by-default (#511).
- The "mask PHI fields" control is relabeled to reflect what it actually does, with a written note on what the access controls do and don't cover (#512).
- Confirmed + tested behavior for rows with no sending facility (#513).
- Documented that a custom report-table name (other than whatever is set in inventory) bypasses row filtering (#514).
- Renamed the underlying attribute `mask_phi_fields → redact_select_identifiers` to match the new label. First deploy, so there's no existing user-data to migrate.

### Technical
- **#510:** the OPA hidden-tables baseline is derived from `report_delta_table_name` rather than hardcoding `reports_report_patient_mapping`, so overriding the report-table name no longer un-hides the facility-less PHI mapping table. (Other table names were already derived from `report_delta_table_name`.)
- **#511:** `should_mask` fires only when the attribute is explicitly `"true"` (unset/empty → no mask); the realm attribute's default flips to `"false"`.
- **#512:** display label → "redact select identifier columns"; a scope / non-goals note added to ADR 0020 and the data-authorization docs (masking covers a few structured columns, not DOB / MRN / free-text).
- **#513:** rego tests pin the fail-closed behavior for empty-string, null-in-array, and null-facility-row cases; wildcard-vs-scoped documented.
- **#514:** ingest docs note that a custom `reportTableName` isn't facility-filtered (non-prod/testing only).
- **Rename:** `mask_phi_fields → redact_select_identifiers` across the OPA policy, the Keycloak realm, the event-listener bundle publisher (+ tests), the Superset auth comment, the data-authorization integration tests, and docs/CLAUDE.md. No behavior change: the bundle publisher forwards attributes by pass-through, so no Keycloak image rebuild is needed for the new name to flow through.

## Type of change
- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [x] Refactor (code improvement with no functional changes)
- [ ] Documentation update
- [ ] Test update

##### Authorization
- #510 closes a PHI exposure that triggers whenever `report_delta_table_name` is overridden (the mapping table becomes directly SELECT-able / `SHOW TABLES`-visible).
- **#511 deliberately reduces default protection** — masking becomes opt-in. Justified: the control only redacts `patient_name` / `full_patient_name` / `zip_or_postal_code`, while DOB, MRN, and the free-text report body are never masked, so on-by-default oversold coverage. This is documented in the new scope note, and per-user isolation of live queries is unaffected.
- #513 confirms/pins fail-closed behavior for missing sending facility (scoped users see no null-facility rows; wildcard users do).

##### Appsec
Policy, realm, and attribute-name changes only; no new input surface.

### Performance
N/A.

### Data
No change to ingested data or schema. The rename does not migrate existing user attributes (first deploy).

### Backward compatibility
The attribute rename would strand existing `mask_phi_fields` values, but this is the first deploy of the attribute so there are none. No API changes.

## Testing
- `opa test policy/` → 95/95 (includes the new null/empty-facility and renamed-mapping-table tests). Keycloak event-listener test suite passes with the renamed key. Deploy paths: OPA policy → `make install-trino`; realm + event-listener → `make install-auth`.
- Manually verified redaction still works on dev04.

## Note for reviewers
Let me know if you think this should default to ON. My feeling was that it's not all that useful, so we probably want to turn it on sparingly. (It's a bit more of a proof that we can do this vs a useful feature at this point.)

## Checklist

Closes #510
Closes #511
Closes #512
Closes #513
Closes #514


docs/internal/adr/0025-launchpad-user-approval.md
```diff
@@ -62,7 +62,7 @@ audience mapper), so a `scout-admin`'s token minted for another resource — e.g
 Because the schema is discovered, **adding an authorization dimension stays the
 one-line inventory edit ADR 0020 promised**: the new attribute flows into the
 form and its server-side validation with no code change here. Safe defaults are
-config-driven too (`scoutDefault`: `mask_phi_fields=true`,
+config-driven too (`scoutDefault`: `redact_select_identifiers=true`,
```
andylassiter: Should this be `redact_select_identifiers=false`? `scoutDefault` is now `false` in the scout-realm.

andylassiter (approved):

PR #523, opened 2026-07-14
docs(gitops): ADRs and initial plans for gitops migration

kathrynalpert:
Initial ADRs with plans for versioning, chart publishing, and gitops deployment.


johnflavin:
I don't understand how dev builds work. The scheme seems to rely on CI building artifacts that are merged to `main`. But if I'm developing a new thing on a branch, how do I deploy it? Where are my artifacts built and deployed, and how does my cluster pull them?

kathrynalpert:
> I don't understand how dev builds work. The scheme seems to rely on CI building artifacts that are merged to `main`. But if I'm developing a new thing on a branch, how do I deploy it? Where are my artifacts built and deployed, and how does my cluster pull them?

Great question. I just pushed up a commit that describes how we'd probably do it (see Appendix C of ADR 0031). I wouldn't say it's a dramatically better DX than what we have today, but I think the "full path" option (once we get there) is probably an improvement.

andrey-embarklabs (approved):

andylassiter (approved):

PR #538, opened 2026-07-16
fix(ingest): Deterministic OBX order

### LINKED ISSUE #537: OBX line order in report_text / report sections is nondeterministic across ingests
## Expected behavior
I was testing a fix for #482 where I hashed the row contents before merging and check against an existing hash column, not re-writing if the hash matches. On a pure reingest I expected 0 rows to be written; the row content should be identical, so the hashes should also be identical.

## Actual behavior
A small number (64 out of 9988, ~0.06%) of the rows were not identical.

What follows is Claude's report from when it found this:

### What was observed

Re-ingesting the identical 9,988-file corpus (`s3://scratch/i482/full-10k_hl7-manifest.txt`) into a fresh table (`reports_i482_b`) twice, on the content-hash-gated transformer:

- Run 1: `MERGE ins=9988` (v1), all `content_hash` populated.
- Run 2 (byte-identical input): expected a complete no-op, got `MERGE upd=64` (v2) — 64 rows re-hashed differently on the second parse of identical bytes.

Diffing the 64 rows' CDF pre/post images (Trino):

```sql
WITH ch AS (
  SELECT * FROM TABLE(system.table_changes(
    schema_name => 'default', table_name => 'reports_i482_b', since_version => 1))
), pre AS (SELECT * FROM ch WHERE _change_type = 'update_preimage'),
   post AS (SELECT * FROM ch WHERE _change_type = 'update_postimage')
SELECT count(*) changed,
  count_if(pre.report_text IS DISTINCT FROM post.report_text) rpt_text,
  count_if(pre.report_section_findings IS DISTINCT FROM post.report_section_findings) sec_gdt,
  -- every other column checked: 0 diffs (status, other sections, diagnoses,
  -- patient_ids, names, epic_mrn)
  count_if(array_sort(split(pre.report_text, chr(10)))
         = array_sort(split(post.report_text, chr(10)))) same_lines_reordered
FROM pre JOIN post ON pre.source_file = post.source_file;
--> changed=64, rpt_text=64, sec_gdt=42, same_lines_reordered=64
```

All 64 diffs were `report_text` (42 also `report_section_findings`), and **every single
one was the same set of lines in a different order** — no content differences at all.

### Root cause

In `import_hl7_files_to_deltalake` (`extractor/hl7-transformer/src/hl7scout/hl7extractor/deltalake.py`),
the report DataFrame is built by:

1. `F.posexplode(obx_lines)` — one row per OBX segment, **with** its index (`pos`);
2. `groupby("source_file").agg(F.concat_ws("\n", F.collect_list("obx-5")))`.

`pos` is carried along but never used for ordering. After the groupBy shuffle,
`collect_list` order is whatever order rows arrive in — Spark makes no intra-group
ordering guarantee — so multi-OBX messages can have their lines concatenated in a
different order from run to run depending on shuffle/AQE scheduling. In this corpus it
bit ~0.6% of reports per run.

Same pattern in two more places:

- `extract_observation_id_suffix_content` (the `report_section_*` builders) —
  `collect_list` again, same instability;
- `F.first("obx-11")` for `report_status`, and `pivot(...).agg(F.first("id_number"))`
  for the per-authority patient-ID columns — `first()` after a shuffle is equally
  nondeterministic (neither happened to flip in this run, but both can).

### Impact

- **Data quality (independent of #482):** `report_text` and the parsed report sections can silently change line order every time the same message is re-ingested. Before the content-hash gate this was invisible — every re-ingest rewrote every row anyway — but the stored text was already unstable. Multi-line findings sections read differently depending on which ingest run last touched the row.
- **With the content-hash gate (#482):** affected rows re-hash and re-merge on every re-ingest (bounded churn — 64/9,988 here — instead of the promised zero), and each re-merge cascades a CDF batch into the derivative tables.

## Steps to reproduce
Hard to force locally (it's shuffle-scheduling dependent); reliable at scale: ingest a multi-OBX-heavy manifest twice into a throwaway table on a pre-`69b9ad04` transformer and diff CDF pre/post images as above — or simply compare `report_text` across the two versions with `FOR VERSION AS OF`.

## Environment
dev03, using synthetic 10k dataset

johnflavin:
## Description
Fixes #537 

### Product
Ensure report text columns (findings, impression, technician note, addendum) are always assembled in the same order as the fields appeared in the original HL7 message.

### Technical
The transformer explodes OBX segments (keeping their `pos` index) and then does a `groupBy(source_file)` aggregation; `collect_list`/`F.first` after that shuffle carry no ordering guarantee. This bug was latent before and possibly never showed up in reality; it only occurs when there is "large-scale shuffle/spill pressure". I did reproduce it with only a 10k test dataset, however, so it is possible (...likely) that the same bug did occur in a production environment. I'm currently running queries to find this bug in a live production environment by querying the delta changelog. The queries I'm using are taking a long time; will update the description later if/when I get definitive results. (UPDATE: Added a reviewer note with real findings.)

Each OBX line is now bundled into a `pos`-led struct, collected, and `array_sort`ed so file order is restored after the shuffle; `report_text`, the parsed sections, and `report_status` are derived from that ordered array. The patient-ID pivot's `F.first("id_number")` is likewise replaced with `F.min("id_number")` so a stable value is chosen every run. This stabilizes the order of all the fields derived from OBX.

## Type of change
- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
N/A

##### Appsec
N/A

### Performance
Operates at production ingest scale. The extra work is a per-group `array_sort` over a report's OBX rows (small); the aggregation shape is otherwise unchanged.

### Data
Handles multi-OBX reports whose line order was previously nondeterministic; `concat_ws` continues to skip NULL OBX-5 values, matching the prior `collect_list` behavior. TX-type `~`→newline handling is preserved.

### Backward compatibility
No schema, API, or dependency changes. Output for a given input is now stable. If any data were previously disordered due to this bug, it will be corrected on a reingest. (Which will likely be necessary anyway after #482.)

## Testing
- There will be some unit tests coming soon which cover this, but they were tied up in the implementation for #482 and didn't cherry-pick cleanly.
- Added `TestScoutQueries.testObxLineOrderIsDeterministic`, which runs during the CI `deploy-and-test` phase. It ingests a new synthetic `obx_ordering` log (150 multi-OBX messages with zero-padded ascending markers in file order), asserts `report_text` and each parsed section come back in ascending marker order, then re-ingests and requires byte-identical `report_text` per `source_file`. I had hoped that I could reproduce the bug in CI by running this test without the fix — see https://github.com/washu-tag/scout/actions/runs/29515858529/job/87681635195 — but I couldn't reproduce it. Claude says, "reliably forcing the reorder needs large-scale shuffle/spill pressure, so a single-node CI run is unlikely to reproduce the original failure — the test is a faithful regression guard that is deterministically green with the fix."

## Note for reviewers
The 4,649-line log fixture is the synthetic test corpus (150 identical multi-OBX messages); the substance is in `deltalake.py` and `TestScoutQueries.java`.

**UPDATE**: I have identified some real errors in a production environment. Mostly they aren't meaningful semantic changes; I often saw a full `EXAMINATION` line move from the beginning of the combined report in one version to the end of the report after a reingestion. But I did see one instance of a multi-line examination break apart. In one version it had `EXAMINATION: <foo> 2 OR MORE VIEWS, <bar> 2 OR` followed by `MORE VIEWS` on its own line; after a reingestion that report started with `MORE VIEWS` on its own line and ended with the `EXAMINATION:` line.

But in what I reviewed, I never saw any lines within the middle of the report (where the more semantically meaningful content lives) switching places. Always the first line moves to the end, or vice versa.

## Checklist


tests/ingest/staging_test_data/obx_ordering/20260601.log
kathrynalpert, line 1: I don't necessarily object to this, but is there some reason the existing test data was insufficient for these tests? It seems like existing data might be slightly richer, and then you could just add a re-ingest check to existing test code. 

(I'm also a little "meh" on adding a test (and CI time) for this at all, given that we can't reproduce the original issue with it. I'm not really sure what this protects.)

kathrynalpert (approved):

tests/ingest/staging_test_data/obx_ordering/20260601.log
cmoore01: That last bit is what I'd really like to understand. 0.6% isn't _that_ small of an occurrence rate. Why have we never seen it (to my knowledge) in the tests?

johnflavin: > is there some reason the existing test data was insufficient for these tests? ...you could just add a re-ingest check to existing test code.

This is probably the right approach. I think the reason for adding the new data was that it was easy to prove it was in the right order since every line was just a number. But I think the existing test data could also be fine without that property. There already is a reingest test, so I could just piggyback on that and maybe add some more assertions.

> 0.6% isn't that small of an occurrence rate. Why have we never seen it (to my knowledge) in the tests?

The way I understand it is that it only shows up when there is enough scale in the data that in the operations we do for reports one spills its exploded OBX lines into multiple spark partitions, which itself is non-deterministic, and even when then does happen it's non-deterministic whether spark gathers those partitions back in the original order or not. I think the test data just doesn't have the scale to trigger this to happen. But I checked on preprod and it definitely did happen there in a few thousand reports (limited to a window of 50 delta versions, which I had to limit because otherwise the queries took too much memory).

PR #539, opened 2026-07-17
feat(ingest): Calculate row hash, skip merging unchanged reports to lake

### LINKED ISSUE #482: reports table merge is inefficient on re-ingest

johnflavin:
## Description

Fixes #482

### Product

No user-facing changes. Re-ingesting HL7 files that are already in the lake with unchanged content is now a no-op.

### Technical

We add a `content_hash` column to the delta lake, which is computed as a hash of all the values in that row serialized to a JSON string (except the `updated` timestamp and `content_hash` itself). When we ingest a batch, we calculate this hash for all the rows and anti-join the `(source_file, year, content_hash)` against the existing results in the delta lake. If rows are the same, they don't show up in the result, and we don't write anything to the delta lake for that row; if rows have changed then we do include them in the merge. The derivative table calculation only triggers off what changed, so they come along more or less for free.

We still records "success" for all files in the batch even if they were skipped. We can't write nothing because the log splitting workflow would have just written "pending" or "staged" or whatever it writes, and we don't want to leave files in that state if there is nothing wrong. I considered writing a new status for these files, something like "no change", but I would rather not proliferate any more statuses or the downstream management of them in grafana. So we just say "success" even though nothing happened.

There is a conditional branch in the code detecting if there is no `content_hash` column and running an `ALTER TABLE` to add it. I don't think this is strictly necessary, I think the delta lake is set to alter the schema if you write columns that it doesn't have, but it seemed harmless to keep this. It should only fire once on the first run after this update; we could remove it in the future once all existing tables are migrated.

I initially wanted to not even process the files through the transformer if we already ingested this file, determined by hashing the hl7 file in the log splitting workflow and passing that to the transformer / delta lake ingest workflow to compare with the stored hash. However, I didn't go with that implementation for a few reasons related to technical complexities.
1. While we almost always launch the log splitting workflow which itself launches the ingest workflow as a child, you _can_ launch the ingest workflow on its own if you want. And I didn't want to have to also figure out how to get the standalone workflow to pass the file hash from that other starting point.
2. More importantly, a hash of the file doesn't capture any changes to the ingest workflow itself. If the file is unchanged but we changed how we ingest, we would want that to show up as "different" and be merged. That implies we would need to store the input file hash as well as some version string for the ingest workflow, which we'd need to remember to increment any time we changed things, and so on. It seemed prone to error. Or maybe we add a "reingest anyway" arg to the workflow that we can pass if we need to bust the cache. I didn't really like either of those.

## Type of change
- [ ] New feature
- [x] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization

N/A

##### Appsec

N/A

### Performance

This does add one anti-join of the new ingest with the existing table pre-merge, which I don't think is very expensive but I don't have exact metrics. If the entire batch is new, this time is wasted compared to the current code. But if parts of the batch are identical, we don't have to remerge them and don't have to calculate all the derivative tables on them again (saves time and disk space).

### Data

New files ingest just the same as before.

The hash should trip and cause a merge for a file that has already been ingested (same `source_path`) but has changed content, or the file is the same but the ingest workflow has changed and the to-be-merged row ends up different. The only time a merge should be skipped is unchanged file + unchanged ingest = entirely unchanged row.

### Backward compatibility

This adds a new column to the `reports` schema, but it will auto-migrate the first time ingest is run after the update.

## Testing

- Added unit tests for this behavior. Ran them red/green: tests run before the fix showed full merges of unchanged content every time the ingest was rerun, but no merged content after the fix.
- Also ran ingests of synthetic data on my dev cluster to verify the same red/green behavior

## Note for reviewers

I currently have the commits for #538 on this branch as well. That was found and fixed as part of this work; a test reingest was expected to have 0 changed rows but did have a small number, which turned out to be because of the reordering bug. Ideally those commits go away and the PR re-points to `main` when that gets merged. If not I'll do some branch shenanigans to fix this PR.

I didn't do an ADR for this. It arguably could have a small one. Let me know if you think we need that.

## Checklist


extractor/hl7-transformer/src/hl7scout/hl7extractor/sparkutils.py
```diff
@@ -4,17 +4,53 @@
...
+    update_condition: Optional[str] = None,
+    with_schema_evolution: bool = False,
 ):
+    # Aliases follow MERGE convention (and latesttable.py): t = target Delta table,
+    # s = source DataFrame being merged in. update_condition may reference columns
+    # as t.<col> / s.<col>.
+    builder = dt.alias("t").merge(
+        df.alias("s"),
+        f"t.{merge_col} = s.{merge_col}{' AND t.year = s.year' if include_year_condition else ''}",
+    )
```
johnflavin, line 49: Reviewer note: forgot to mention this change. It's a tiny consistency nit. This merge used the opposite convention for which table to call `s` and `t` as other parts of the code. It should change nothing functional.

extractor/hl7-transformer/src/hl7scout/hl7extractor/deltalake.py
```diff
@@ -165,6 +169,81 @@ def spark_activity_session(app_name: str, health_file: Optional[Path] = None):
...
+    if table_exists:
+        if CONTENT_HASH_COL not in spark.table(full_name).columns:
+            # One-time, metadata-only migration for tables created before this
+            # column existed. Explicit ALTER (rather than merge-time schema
+            # evolution) so t.content_hash is guaranteed resolvable in the matched
+            # condition and the projection below is uniform.
+            spark.sql(
+                f"ALTER TABLE {full_name} ADD COLUMNS ({CONTENT_HASH_COL} STRING)"
+            )
+        existing = spark.table(full_name).select(
```
kathrynalpert: Claude recommends:

> In `merge_report_df_into_table`, the `existing` DataFrame that feeds the `left_anti` join reads `(source_file, year, content_hash)` from the whole table with no filter — so every ingest reads and shuffles those three columns across *all* partitions, and even a single-file re-ingest drags every historical `year` through the anti-join. The MERGE right below it doesn't pay this cost: Delta already prunes it to the batch's `year` partitions from the source's min/max.
>
> Since the anti-join requires `year` equality, an incoming row can only match an existing row in the *same* year partition, so we can restrict `existing` to the batch's years without losing any match:
>
> ```python
> batch_years = [
>     row.year for row in df.select("year").distinct().collect()
>     if row.year is not None
> ]
> existing = (
>     spark.table(full_name)
>     .where(F.col("year").isin(batch_years))   # partition filter → Delta prunes dirs
>     .select(
>         F.col("source_file").alias("_existing_source_file"),
>         F.col("year").alias("_existing_year"),
>         F.col(CONTENT_HASH_COL).alias("_existing_content_hash"),
>     )
> )
> ```
>
> Correctness is unchanged: NULL-year rows are excluded from `batch_years` but never anti-match anyway (they always fall through to insert, as today), and an all-NULL-year batch yields an empty `existing` so everything merges — same as current behavior. The added cost is one `distinct().collect()` over `df`, which is already cached at the call site. The win scales with how many year-partitions the table holds vs. how many the batch touches — for a table spanning many years and a daily batch, that's the difference between scanning every year and scanning one.
>
> Non-blocking — but this `existing` scan is where I'd point the "operates at production ingest scale / no exact metrics yet" note: it's the one read that grows with total table size on every ingest, so it's worth a single real measurement at prod scale.

kathrynalpert (approved):
I think an ADR would be good, as this does constrain column types a bit, and I think that's a good component to have in docs

johnflavin:
Added a short ADR, numbered 0032 to not conflict with numbering in #523. (Though there will still be a minor merge conflict on the summaries in `CLAUDE.md`.)

extractor/hl7-transformer/src/hl7scout/hl7extractor/deltalake.py
johnflavin: Good change. Accepted.

PR #541, opened 2026-07-17
feat(gitops): publish charts to OCI and make spark-defaults chart-owned

andrey-embarklabs:
### Product
Phase 0 of the GitOps migration (ADRs 0030/0031). It makes Scout's Helm charts consumable as versioned OCI artifacts and lets a chart carry its own Spark configuration, so a GitOps-based deployment can pin an immutable chart version and stop hand-copying config out of the Ansible roles. There is no change for end users of a running Scout; this is deployment- and operator-facing groundwork, and the existing Ansible install is unchanged.

### Technical
A vertical slice through the `hl7-transformer` chart, in two independent pieces:

1. **Chart publishing.** A new `publish-charts` job in `.github/workflows/ci.yaml` runs `helm package` + `helm push` for charts whose directory changed and publishes them to `oci://ghcr.io/washu-tag/charts` at the ADR 0030 build-lane version `0.YYYYMMDD.<run>` (minted once per run on the `changes` job). It is gated exactly like the image `publish` job: `main` only, after `deploy-and-test` and `smoke-test` succeed, changed-charts-only. Only `hl7-transformer` is in the matrix now; the remaining charts are an additive follow-up. Only the chart `--version` is stamped at package time; `Chart.yaml` keeps its `0.0.0-dev` / `appVersion: latest` placeholders (appVersion-to-image coupling is phase 2).

2. **Chart-owned spark-defaults.** `spark-defaults.conf` can now be rendered by the chart from a `sparkDefaults` values block (a `_helpers.tpl` helper switched on an `on-prem`/`aws` service mode, ADR 0011), instead of only by the Ansible-rendered ConfigMap. It is opt-in (`sparkDefaults.enabled` defaults to `false`) so it does not double-render when a caller injects its own spark-defaults volume. The chart-owned on-prem render keeps the S3 credentials out of the ConfigMap: it uses the SDK-v2 `EnvironmentVariableCredentialsProvider` off the `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` env the pod already mounts via `s3-secret`.

The Ansible spark-defaults machinery (`scout_common/spark_config_setup`, `spark-defaults.conf.j2`, the extractor wiring) is left intact, with comments noting the config has been transferred to the chart; it retires separately when the Ansible service roles are removed at the ADR 0031 phase 5 cutover.

## Type of change
- [ ] New feature
- [x] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
No user roles, RBAC, or data-access paths are added or modified, and no REST API surface changes. The only new privilege is the `publish-charts` job's `packages: write`, scoped to that job alone (least privilege), authenticating with the workflow `GITHUB_TOKEN` (no PAT) passed to `helm registry login` via stdin (never argv).

##### Appsec
Reviewed. The chart-owned on-prem render removes the plaintext S3 access/secret key from the spark-defaults ConfigMap (they now come from the existing `s3-secret` env), so no credentials live in a ConfigMap on the chart path. Action pins are by commit SHA, matching the repo convention. Note the unchanged Ansible path still inlines those credentials into its own ConfigMap and retires separately (phase 5); this PR does not regress it.

### Performance
Runtime is unaffected. CI adds one job that runs only on `main`, only after `deploy-and-test` / `smoke-test`, and packages/pushes only charts that changed (typically one), so the added cost is a single lightweight package + push.

### Data
No changes to the data model, DICOM/HL7 handling, or any input-data path. The rendered `spark-defaults.conf` is behavior-preserving (see Testing).

### Backward compatibility
- **Ansible on-prem deploy: unchanged.** `sparkDefaults.enabled` defaults to `false`, so the chart does not render its own spark-defaults; the extractor role continues to render and mount its own ConfigMap exactly as before (verified: no duplicate volume).
- **Chart consumers:** charts are now additionally published as immutable OCI artifacts; existing git-branch consumers are unaffected until they opt in (a separate downstream repo PR).
- The chart's default image tag stays the published `:latest` in phase 0 (appVersion left at the placeholder); consumers pin `image.tag` for an immutable image. No API or dependency changes.

## Testing
- `helm lint` clean; `helm template` renders correctly with `sparkDefaults` enabled (both `on-prem` and `aws`) and disabled.
- Behavior equivalence of the chart-owned render vs the former Ansible `spark-defaults.conf.j2`, diffed with matching inputs: `aws` mode is line-identical; `on-prem` differs only by the two removed plaintext credential lines and the provider swap (the intended security fix), with all other lines identical.
- Opt-out path: with `sparkDefaults.enabled: false` and a caller-injected spark-defaults volume (the Ansible scenario), the chart renders no ConfigMap and passes through the caller's single volume, confirming no double-mount.
- OCI mechanics: full `helm package` + `helm registry login` + `helm push` + `helm pull` round-trip against a local OCI registry; the artifact lands at `.../charts/hl7-transformer:<version>` and pulls back with a matching digest. The packaged `Chart.yaml` keeps its in-tree placeholders (version is stamped only at package time).
- `.github/workflows/ci.yaml` validated (YAML parse + `actionlint`, no new findings).
- The existing `deploy-and-test` and `smoke-test` jobs are the end-to-end behavioral gate for the unchanged Ansible deploy path and must stay green.

## Note for reviewers
- Vertical slice: only `hl7-transformer` is wired into the `publish-charts` matrix; the other charts join in a follow-up (which should also weigh a dynamic changed-chart discovery matrix instead of per-chart `paths-filter` entries).
- The `publish-charts` job must remain in `ci.yaml`: `github.run_number` is scoped to the workflow file, so moving or renaming it would reset the sequence and could invert OCI tag ordering when phase 2 extends the job.
- The Ansible spark-defaults machinery is intentionally kept (comment-only diff) and retires at the phase 5 cutover, so the credential-handling improvement applies to the chart path only for now.
- The first publish of a new GHCR package via `GITHUB_TOKEN` may need a one-time package visibility / repo-link step.
- A separate downstream PR will repoint the `hl7-transformer` chart source from the git branch to the OCI chart and delete the hand-copied ConfigMap; that is when the config-staleness problem actually closes.

## Checklist
- [ ] My code adheres to the coding and style guidelines of the project (and I've run `pre-commit run --all-files`)


.github/workflows/ci.yaml
```diff
@@ -835,16 +786,53 @@ jobs:
...
-        with:
-          image-name: report-viewer
       - name: 'Push Image - xnat-plugin-installer'
         if: needs.changes.outputs.xnat-plugin-installer == 'true'
         uses: ./.github/actions/docker-push
         with:
           image-name: xnat-plugin-installer
+  # GitOps migration phase 0 (ADR 0030/0031): publish changed Helm charts as
+  # versioned OCI artifacts. Gated like the image `publish` job (main only,
+  # after deploy-and-test + smoke-test succeed). Vertical slice: hl7-transformer
```
kathrynalpert: Nit: This feels like an unnecessarily verbose comment

ansible/roles/extractor/tasks/deploy.yaml
```diff
@@ -96,6 +96,10 @@
       data:
         modality_mapping_codes.csv: '{{ lookup("file", modality_map_source_file) }}'
 
+# spark-defaults.conf is now also chart-owned in the hl7-transformer chart
+# (opt-in via sparkDefaults.*, ADR 0031 phase 0). This role keeps rendering its
+# own ConfigMap for the on-prem deploy path; it retires separately when the
```
kathrynalpert: IMO, the Ansible ConfigMap ought to be removed here and the Ansible deploy should swap to using the OCI chart. I don't see the benefit in keeping parallel paths until phase 5 when this would work today.

helm/extractor/hl7-transformer/templates/_helpers.tpl
```diff
@@ -60,3 +60,40 @@ Create the name of the service account to use
...
+Render spark-defaults.conf from the sparkDefaults values block. This is the
+former scout_common/spark-defaults.conf.j2, moved into the chart (ADR 0031 §4)
+and switched on .Values.sparkDefaults.mode (the ADR 0011 service mode).
+
+On-prem reads its S3 credentials from the AWS_ACCESS_KEY_ID /
+AWS_SECRET_ACCESS_KEY environment variables (the s3-secret the Deployment
+already mounts via envFrom) through the SDK-v2 EnvironmentVariableCredentialsProvider,
+rather than inlining the keys here, so no credentials live in the ConfigMap.
+*/}}
+{{- define "hl7-transformer.sparkDefaultsConf" -}}
```
kathrynalpert, line 74: Claude warns "the `sparkDefaults` render fails at *runtime* rather than at *template time* when it's enabled but under-configured. `hiveMetastoreUri`, `warehouseDir`, and `s3Endpoint` all default to `''`, so `sparkDefaults.enabled: true` with any of them unset produces a syntactically-valid ConfigMap carrying blank values, and the Spark job then fails when it hits Hive/S3, with no signal from `helm template`/`lint` pointing at the real cause. A typo in `mode` is worse: anything that isn't `"aws"` silently falls into the on-prem branch."

> Since these have no sensible defaults when enabled, I'd make them fail fast in `hl7-transformer.sparkDefaultsConf`:
> 
> ```gotemplate
> {{- define "hl7-transformer.sparkDefaultsConf" -}}
> {{- $s := .Values.sparkDefaults -}}
> {{- if not (has $s.mode (list "on-prem" "aws")) -}}
> {{- fail (printf "sparkDefaults.mode must be \"on-prem\" or \"aws\", got %q" $s.mode) -}}
> {{- end -}}
> {{- $aws := eq $s.mode "aws" -}}
> {{- if $aws }}
> spark.hadoop.fs.s3a.aws.credentials.provider software.amazon.awssdk.auth.credentials.WebIdentityTokenFileCredentialsProvider
> {{- else }}
> spark.hadoop.fs.s3a.endpoint {{ required "sparkDefaults.s3Endpoint is required when sparkDefaults.mode=on-prem" $s.s3Endpoint }}
> spark.hadoop.fs.s3a.aws.credentials.provider software.amazon.awssdk.auth.credentials.EnvironmentVariableCredentialsProvider
> {{- end }}
> spark.hadoop.fs.s3a.endpoint.region {{ $s.s3Region }}
> ...
> spark.hadoop.hive.metastore.uris {{ required "sparkDefaults.hiveMetastoreUri is required when sparkDefaults.enabled" $s.hiveMetastoreUri }}
> spark.sql.warehouse.dir {{ required "sparkDefaults.warehouseDir is required when sparkDefaults.enabled" $s.warehouseDir }}
> ```
> 
> (`required` fails on empty string as well as nil, so it does the right thing against the `''` defaults.)
> 
> The one thing this *can't* catch is the on-prem credential contract: the render assumes the pod carries `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` (via the consumer's `envFrom` → `s3-secret`), and Helm can't see runtime env at template time. I'd cover that with a line in the `sparkDefaults` values comment — something like *"on-prem mode reads S3 creds from `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`; the consumer must supply these via `envFrom` (e.g. `s3-secret`), and must not inject its own `spark-defaults` volume when enabled."* That last clause also documents the duplicate-volume-name footgun.

helm/extractor/hl7-transformer/values.yaml
```diff
@@ -113,3 +113,20 @@ affinity: {}
 spark:
   executor:
     memory: 1g
+
+# Optional chart-owned spark-defaults.conf, rendered by _helpers.tpl (see
+# "hl7-transformer.sparkDefaultsConf"). Opt-in (disabled by default) so it does
+# not double-render when a caller injects its own spark-defaults volume; the
+# Ansible on-prem path still renders its own ConfigMap, retired separately at the
```
kathrynalpert: Just to reiterate, I don't know why we wouldn't retire the Ansible ConfigMap here, and swap the ansible to use the OCI chart (as we've done for xnat, data generator, etc.)

.github/workflows/ci.yaml
andrey-embarklabs: Trimmed to two lines in 6a42579.

ansible/roles/extractor/tasks/deploy.yaml
andrey-embarklabs: Done in 6a42579. Dropped the spark_config_setup ConfigMap task and deleted the now-orphaned scout_common/spark_config_setup task + template, and the extractor deploys hl7-transformer from oci://ghcr.io/washu-tag/charts/hl7-transformer with a hl7_transformer_chart_path local override for dev/CI, matching the xnat/data-generator pattern. hl7log-extractor gets the same treatment in a follow-up so the whole role is consistent. One caveat: hl7_transformer_chart_version is a 0.0.0-dev placeholder until the chart first publishes from main; CI uses the local override, so it isn't exercised there.

helm/extractor/hl7-transformer/templates/_helpers.tpl
andrey-embarklabs: Implemented your suggestion in 6a42579. The helper now fails on a mode that isn't on-prem/aws and uses required for s3Endpoint (on-prem), hiveMetastoreUri, and warehouseDir when enabled, and I added the credential-contract note to the sparkDefaults values comment. Verified lint stays clean at the default (enabled: false) and each misconfiguration errors with a clear message.

helm/extractor/hl7-transformer/values.yaml
andrey-embarklabs: Agreed, done in 6a42579 (details in the deploy.yaml thread). The chart owns spark-defaults and the extractor sets enabled: true; the parallel Ansible ConfigMap is gone. I kept the chart default enabled: false so helm lint stays clean and it composes with callers that inject their own volume, but the extractor opts in.

ansible/group_vars/all/versions.yaml
```diff
@@ -182,3 +182,7 @@ xnat_chart_version: 2.0.0
 # data-generator Helm chart, published as an OCI artifact on GHCR (Renovate-monitored, ADR 0015)
 # renovate: datasource=docker depName=ghcr.io/washu-tag/charts/data-generator
 data_generator_chart_version: 1.0.0
+# Placeholder until the first hl7-transformer chart publishes from main (ADR 0030
+# build lane, 0.YYYYMMDD.<run>); bump to a published version. CI and dev use the
+# hl7_transformer_chart_path local override, so this is not exercised in CI.
+hl7_transformer_chart_version: 0.0.0-dev
```
kathrynalpert: I don't think we need a placeholder (OCI will be published with this PR), and I think it's a bit confusing to self-reference in the Renovate versions yaml.

kathrynalpert (approved):

ansible/group_vars/all/versions.yaml
andrey-embarklabs: Good call. Switched it to match the data-generator entry in a4bbfe9: a one-line description plus the # renovate: annotation for ghcr.io/washu-tag/charts/hl7-transformer, and dropped the placeholder prose. Renovate will manage the version once the chart first publishes from this PR; the chart-path / CI-override mechanics are documented in the extractor role rather than here. (Also rebased onto your main-merge.)

PR #548, opened 2026-07-20
fix(report-viewer): CSV Export Fix and Dark Mode

### LINKED ISSUE #535: CSV exported by report viewer service should reflect user's view
When I click "Download CSV" in the report viewer service, it outputs the full amount of columns (including `primary_report_identifier`) and the full amount of rows from the initial search. However, if the user has done additional filtering, what they very likely wanted to download was a CSV of their current rows and columns (plus maybe the unique ID for the reports). Debateably, even the sort order should be preserved, but that doesn't seem like a big deal.

### LINKED ISSUE #542: Report Viewer Dark Mode
## Problem
The viewer is hardcoded with light mode CSS colors. Inside OWUI's dark mode chat it's a glaring white panel.

## Proposed solution
Swap the hardcoded colors for CSS variables with a dark mode set under `@media (prefers-color-scheme: dark)`. No toggle or setting, just follow the browser theme.

## Alternatives considered
Match the OWUI theme (can differ from the browser setting). More plumbing would be required to pass the setting from OWUI to the report viewer. Following the browser theme seems reasonable.

## Additional context
Related: #534 (screenshot)

andylassiter:
### Product
Dark mode (#542): the viewer now follows the browser's light/dark theme automatically.

CSV export (#535): Download CSV now matches what you see, the filtered rows, chosen columns, and sort order, instead of dumping the whole search. The `primary_report_identifier` is always included so rows stay linked back to the Scout data lake.

### Technical
Dark mode: colors moved into CSS variables in a new `theme.css` (light `:root` defaults plus a `@media (prefers-color-scheme: dark)` block); the inline styles reference the vars. No new deps.

CSV: the `/csv` endpoint now takes the same sort/filter params as `/rows` plus a `columns` list. A shared `_build_where_and_order` (backend) and `buildViewQuery` (frontend) keep the two endpoints from drifting, and `primary_report_identifier` is always included.

Closes #542, #535

## Type of change
- [x] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
No change to the trust model. Export runs as the user through the existing path, same as the rest of the viewer.

##### Appsec
Export columns and sort are validated for identifier charset and quoted via `_quote_ident`; filter values are bound as params. No user values interpolated into SQL.

### Performance
Same query path and streaming response as before, just with the filter/sort/projection applied. No new scale concerns.

### Data
CSV always includes `primary_report_identifier` so rows stay linked back to the Scout data lake. 

### Backward compatibility
`/csv` gains optional query params. No data model or dependency changes.

## Testing
Tested on the dev cluster. Toggled between dark and light mode in Chrome and the viewer theme switched automatically to match. Sorted and filtered the table, then downloaded the CSV and it matched what the user sees, the same rows and columns, plus `primary_report_identifier.`

## Note for reviewers
Screenshots with fake data attached.

## Checklist

RV-LightMode

RV-DarkMode

RV-CSV-Download1

RV-CSV-Download2

### DIFF
--- a/report-viewer/frontend/src/App.tsx
+++ b/report-viewer/frontend/src/App.tsx
@@ -7,8 +7,8 @@ function App() {
       <div
         style={{
           fontFamily: '-apple-system, BlinkMacSystemFont, "Segoe UI", system-ui, sans-serif',
-          color: '#222',
-          background: '#fff',
+          color: 'var(--rv-fg)',
+          background: 'var(--rv-surface)',
           padding: '0.5rem',
           height: '100vh',
           boxSizing: 'border-box',
@@ -25,9 +25,9 @@ function App() {
     <div
       style={{
         fontFamily: '-apple-system, BlinkMacSystemFont, "Segoe UI", system-ui, sans-serif',
-        color: '#222',
+        color: 'var(--rv-fg)',
         height: '100vh',
-        background: '#fafafa',
+        background: 'var(--rv-bg)',
         display: 'flex',
         flexDirection: 'column',
         overflow: 'hidden',
@@ -36,8 +36,8 @@ function App() {
       <header
         style={{
           padding: '0.75rem 1.5rem',
-          background: '#fff',
-          borderBottom: '1px solid #e2e2e2',
+          background: 'var(--rv-surface)',
+          borderBottom: '1px solid var(--rv-border)',
           flex: '0 0 auto',
         }}
       >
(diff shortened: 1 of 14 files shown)


kathrynalpert (approved):
Claude had one finding:
>Loading spinner isn't themed — SearchDetailPage.tsx:420-421 keeps border: '2px solid #fde6c2' / borderTopColor: '#ea580c'. It's the only interactive color the PR leaves hardcoded (I grepped the whole src/ and cross-checked against the diff — everything else is converted or is an intentional pin). In dark mode you get a light-peach ring on a dark surface. Cosmetic, but it's the kind of straggler that's cheap to fold in now (e.g. var(--rv-border) track + an accent/orange top).

PR #550, opened 2026-07-21
feat(gitops): publish + OCI-deploy hl7log-extractor (phase 0 follow-up)

andrey-embarklabs:
Follow-up to #541. Extends the phase-0 chart pattern to the sibling hl7log-extractor chart so the extractor role deploys both charts the same way.

- publish-charts: add hl7log-extractor to the matrix and its changed-path filter.
- extractor role: deploy hl7log-extractor from oci://ghcr.io/washu-tag/charts/hl7log-extractor with a hl7log_extractor_chart_path local override (CI inventory points at the in-repo chart). Plain on-prem pin, not Renovate-tracked (self-built, matching the hl7-transformer pin).

hl7log-extractor has no spark-defaults, so this is a deploy-pattern change only.

### Technical
Follow-up to #541. Extends the phase-0 chart pattern to the sibling hl7log-extractor
chart so the extractor role deploys both charts the same way.
- publish-charts: adds hl7log-extractor to the matrix + a changed-path filter.
- extractor role: deploys hl7log-extractor from oci://ghcr.io/washu-tag/charts/hl7log-extractor
  with a hl7log_extractor_chart_path local override (CI uses the in-repo chart). Plain
  on-prem pin, not Renovate-tracked (self-built, matching hl7-transformer).

## Type of change
- [x] Improvement

## Testing
helm lint + package clean; ci.yaml actionlint clean; deploy-and-test exercises the chart.

## Note for reviewers
hl7log-extractor has no spark-defaults, so this is a deploy-pattern change only.

## Checklist

### DIFF
--- a/.github/ci_resources/inventory.yaml
+++ b/.github/ci_resources/inventory.yaml
@@ -115,9 +115,10 @@ k3s_cluster:
     base_dir: K3S_DATA_DIR/storage
     scout_repo_dir: WORK_DIR
     extractor_data_dir: '{{ scout_repo_dir }}/tests/ingest/staging_test_data'
-    # Deploy the in-repo hl7-transformer chart (unpublished on PRs); production
-    # pulls the OCI chart pinned to hl7_transformer_chart_version.
+    # Deploy the in-repo extractor charts (unpublished on PRs); production pulls
+    # the OCI charts pinned to the matching *_chart_version.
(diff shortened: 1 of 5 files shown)


kathrynalpert (approved):
Unfortunately, when you modify ci.yaml, it reruns all image scans (in case image build params changed), and it looks like there are some new vulns unrelated to this PR. Claude can usually review them and add the required ignore entries (if version changes are needed, let's discuss on Slack).

PR #554, opened 2026-07-21
feat(report-viewer): data-driven modality filter + chat confirm

### LINKED ISSUE #525: Users lose unsent chat when using report-viewer "propagate to chat" features
## Problem
When a user using the report-viewer clicks any option that passes a pre-composed message to the chat ("Filter via Chat", "Discuss in Chat"), any content but not yet sent is overwritten. In a scenario where a user has written a lengthy message but not yet sent it, that will lead to a lot of frustration in losing the message without a way to get it obviously get it back. I've confirmed that the browser's undo function _does_ work to restore the chat, but a lot of users won't realize that.

## Proposed solution
I'm not entirely sure. Maybe we warn the user that their unsent chat will be overwritten when they click one of these options if they have unsent changes?

Found in #483

### LINKED ISSUE #526: Modalities in filtering should be specific to the particular data
## Expected behavior
It would be more user friendly if the modalities shown to the user in the filter dialog were only the set of modalities present in the data. Currently, we always provide the user with a large list of modalities, even if only 1 or 2 of them are present in the data.

## Actual behavior
When a user is filtering their search in the report-viewer, they see the entire pre-defined list of modalities.

## Additional context

Image

Found in #483

### LINKED ISSUE #531: "Filter via Chat" is potentially unintuitive
My first experience with this button in the filter dialog was to click it with no filters selected. This closed the dialog without doing anything. My initial thought was that the button was broken. I think two possible changes could make this more logical to the end user:

1. If this button is clicked with no filters specified, tell the end user that some filters must be specified to pass them along to the chat.
2. Can we snap to the outgoing text buffer or put some sort of indicator on it that it has been updated? For a while, I didn't even realize it was updating my outgoing chat buffer.

Found in #483

andylassiter:
### Product
Three report-viewer improvements:
- #526: the Modality filter now only lists modalities actually present in the search cohort, instead of the full hardcoded list.
- #525: sending a filter or report to chat now asks for confirmation first, so you don't silently lose whatever you'd typed in the chat box.
- #531: "Filter in Chat" with no filters selected now tells you to pick a filter first (it used to just close and look broken), and the confirm dialog makes it obvious the action updates your chat.

### Technical
- #526: new `GET /api/searches/{id}/modalities` (twin of `/accessions`), fetched in parallel on viewer load. Defaults to the full list until it resolves, hides the row if the cohort has none, unions with the current selection so a checked value can't vanish.
- #525: a `ChatPromptProvider` confirm dialog gates every send to chat (the iframe is cross-origin so we can't detect a draft). Modal gained an X close button and Esc closes only the top modal.
- #531: renamed "Filter via Chat" to "Filter in Chat" and added an inline "select at least one filter first" guard instead of silently closing on an empty selection.

## Type of change
- [x] New feature
- [x] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
No change. The modalities query runs as the user (`X-Trino-User`), so OPA row filters apply exactly like `/rows` and `/accessions`.

##### Appsec
No new injection surface. `modality` is a literal, the saved SQL is persisted/validated, and IDs bind via `?`.

### Performance
One extra Trino `DISTINCT` on a single low-cardinality column per viewer load, run in parallel with the rows query and cached. Lighter than the rows query already running.

### Data
Empty/NULL modalities are dropped; a cohort with none hides the filter. `COLUMN_NOT_FOUND` (search doesn't project modality) returns an empty list, not a 502.

### Backward compatibility
Additive. New endpoint, no schema or existing-API changes.

## Testing
Backend unit tests for the new endpoint (happy path, empty, `COLUMN_NOT_FOUND`, 502, 404). Manually tested on dev: modality filter narrows to present values, chat confirm dialog wording/flow, dark + light mode.

## Note for reviewers
Combines #525 ,#526, and #531 (tested together). Screenshots attached from dev env.

## Checklist

RV-Modality

RV-FilterInChatModal

RV-DiscussInChatModal

RV-NoFilterSelected


report-viewer/src/scout_report_viewer/routes/searches.py
```diff
@@ -770,6 +776,45 @@ async def get_search_accessions(
...
+        f"SELECT DISTINCT s.modality "
+        f"FROM ({ds['sql']}) s "
+        f"WHERE s.modality IS NOT NULL "
+        f"ORDER BY s.modality"
+    )
+    try:
+        with metrics.time_trino("modalities_query"):
+            # safe: source_sql is persisted validated SQL, IDs bind via ?; OPA is the AuthZ boundary
+            # nosemgrep: python.sqlalchemy.security.sqlalchemy-execute-raw-query.sqlalchemy-execute-raw-query
+            _cols, rows = await trino_client.execute(
```
kathrynalpert, line 801: Is it really necessary to rerun the query to get the list of modalities? Would it make more sense to compute + store this from the original query?

kathrynalpert (approved):

report-viewer/src/scout_report_viewer/routes/searches.py
cmoore01: It seems pretty unlikely, but _technically_, the data could change over time such that a cached list of modalities for a given search could become out of date. Is that worth thinking about though?

andylassiter: I was considering that but we run the query live so storing the modalities from the original query could be out of sync with what is being shown to the user.

PR #564, opened 2026-07-27
fix(launchpad): Expose favicon through auth on launchpad login

### LINKED ISSUE #557: Favicon is blocked by auth
We never notice this because the browser is caching the favicon, but it's something new users would run into. When you initially load Scout, there's no favicon for your tab because it can't make it through Scout's auth:

Image

Once you log in, it's fine.

cmoore01:
### Product
First-time users (and anyone on a fresh browser) now see the Scout favicon in
their browser tab on the initial, pre-login page load. Previously the tab showed
no icon until after authentication — returning users never noticed because the    
browser caches the favicon after first login, but new users always hit it.

### Technical
The launchpad serves `/favicon.ico`, and the launchpad ingress sits behind the
oauth2-proxy ForwardAuth gate, so an unauthenticated request for `/favicon.ico`
is 401'd instead of returning the icon. This adds a gated, exact-path,
host-scoped route (`faviconPublicRoute`) that serves *only* `/favicon.ico` from
the launchpad service with the `security-headers` middleware but **without** the
oauth2-proxy `auth`/`error` middlewares.

Implemented as an **annotated Ingress**, not a Traefik `IngressRoute`: Scout's
shared middlewares live in `kube-system`, and Traefik's CRD provider blocks
cross-namespace middleware references (`allowCrossNamespace=false`), so an
IngressRoute in the launchpad namespace can't reference
`kube-system/security-headers` (Traefik disables the router at runtime). The
Ingress `router.middlewares` annotation is not subject to that restriction — the
launchpad Ingress already references the same `kube-system-*` middlewares this
way. Off by default in the chart (`faviconPublicRoute.enabled: false`); the
launchpad role enables it and passes the host.

## Type of change
- [ ] New feature
- [ ] Improvement
- [x] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization
This PR intentionally makes one path — `/favicon.ico` on the launchpad host —
reachable without authentication, so it warrants a careful read. Scoping that
keeps the exposure to a single, non-sensitive static asset:
- **Exact path + specific host** (`pathType: Exact`, explicit `host`), so the
  auth-free router matches only `/favicon.ico` on the launchpad host. It is
  **not** a prefix and **not** a wildcard host — no other path or host can fall
  through it. (Comments in the template call this out to prevent a future
  widening to `Prefix`.)
- **Priority is a tiebreaker, not a traffic-widener** — it only decides that the
  favicon router wins over the launchpad `PathPrefix(/)` router *for that one
  path*. Every other path still routes to the auth-gated launchpad router;
  no existing route loses its oauth2-proxy gate.
- **The asset is already public** — `launchpad/public/favicon.ico` is shipped in
  the repo; serving it unauthenticated discloses nothing. Worst case is "the
  favicon is fetchable without a session," which is the intended outcome.
- No roles, REST paths, or data access are added or modified.

##### Appsec
The `security-headers` middleware is retained on the route; only the
auth/error middlewares are dropped, and only for the exact favicon path. Exact
matching prevents path-traversal/prefix abuse of the auth-free router.

### Performance
Negligible — one additional Traefik router for a single static file. No new
compute, storage, or upstream load.

### Data
N/A — no data handling, parsing, or schema involvement.

### Backward compatibility
Additive and gated. New chart value `faviconPublicRoute` (default disabled in
the chart; enabled by the launchpad role). No changes to the data model, APIs,
or dependencies. Behavior for all other paths is unchanged.

## Testing
- **Bug reproduced pre-fix** on dev01
- **Caught a runtime failure in the first approach:** the initial IngressRoute
  was rejected by Traefik (`middleware kube-system/security-headers is not in
  the IngressRoute namespace scout-core`); switched to the annotated Ingress.
- **Post-deploy verification (to run on dev01 after `make install-launchpad`):**
  - Checked with curl
  - Fresh incognito browser → favicon visible in the tab before login.
  - Small sanity check on admin path to make sure auth still working as expected

## Note for reviewers
This deliberately exposes `/favicon.ico` unauthenticated — please sanity-check
the exact-path + host scoping in `favicon-ingress.yaml`.

## Checklist
- [x] My code adheres to the coding and style guidelines of the project (run `pre-commit run --all-files`)
- [ ] I have added or updated technical documentation, including an ADR, if appropriate 


helm/launchpad/templates/favicon-ingress.yaml
```diff
@@ -0,0 +1,49 @@
...
+metadata:
+  name: {{ include "launchpad.fullname" . }}-favicon-public
+  labels:
+    {{- include "launchpad.labels" . | nindent 4 }}
+  annotations:
+    traefik.ingress.kubernetes.io/router.priority: {{ .Values.faviconPublicRoute.priority | quote }}
+    traefik.ingress.kubernetes.io/router.middlewares: >-
+      {{ .Values.faviconPublicRoute.securityHeadersMiddleware.namespace }}-{{ .Values.faviconPublicRoute.securityHeadersMiddleware.name }}@kubernetescrd
+spec:
+  ingressClassName: traefik
```
andylassiter: Claude finding:

> Hardcoded `apiVersion`/`ingressClassName` vs. the main ingress. ingress.yaml version-gates `apiVersion` and uses `.Values.ingress.className`; the favicon template hardcodes `networking.k8s.io/v1` and `ingressClassName: traefik`.

andylassiter (approved):

PR #567, opened 2026-07-28
ci(ingest-tests): warm the Gradle cache so the ingest e2e runs offline

andrey-embarklabs:
### Technical
The `deploy-and-test` "Run tests" step (the ~18m long pole on every PR) ran
`gradlew clean test` in an ephemeral pod, so each run re-downloaded the Gradle
distribution and the full Spark4 test classpath from Maven Central and the slow
dcm4che / nrgxnat mirrors (~5m cold start inside the timed step).

- A "Warm Gradle cache" step resolves that classpath on the runner (cached across
  runs by `setup-java`), then hands a world-writable copy to the in-cluster test
  Job via a hostPath mount at `/gradle-cache`.
- The Job now runs `gradlew test --offline` and drops `clean` (the mount is a fresh
  checkout with no prior build output), so the cold download no longer happens
  inside the timed test step.

## Type of change
- [x] Improvement

### Performance
Removes the ~5m per-run dependency download from the timed "Run tests" step.
### Backward compatibility
No app or test changes. The same tests run the same way; only how the in-cluster
test runner obtains its dependencies changes.

## Testing
This PR's own `deploy-and-test` run exercises the offline test runner.

## Checklist
- [x] Adheres to style (`pre-commit run --all-files`)


cmoore01:
The gradle caching looks like a great idea, but I'm a bit iffy on abbreviating the tests on CI (both on which tests run, but also running slimmed down versions of some tests). That's something we've considered doing long term, for sure, but right now the runtimes don't seem that long such that losing the coverage on PRs is worth it to me.

andrey-embarklabs:
> The gradle caching looks like a great idea, but I'm a bit iffy on abbreviating the tests on CI (both on which tests run, but also running slimmed down versions of some tests). That's something we've considered doing long term, for sure, but right now the runtimes don't seem that long such that losing the coverage on PRs is worth it to me.

good call, narrowed this one down to just the gradle cache + --offline change and dropped the test abbreviation, so PRs keep full ingest coverage. if we do want to trim the ~36m pipeline later without losing coverage, the bigger lever is actually the stack deploy: deploy-and-test spends ~13m just waiting on cassandra/es/temporal readiness before the tests even start. happy to look at that separately.

cmoore01 (approved):
The PR title is a bit outdated now

Issue #476, opened 2026-07-01
Integrate Conventional Commits tooling / enforcement into Scout


johnflavin:
Enforce commit messages following [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) style. Not sure exactly what this will look like, but could be

- A note for humans (and robots) in a `CONTRIBUTING.md` file
- A note for robots in the `CLAUDE.md` file
- [`commitlint`](https://commitlint.js.org) in `pre-commit` and/or CI
- Update renovate / depandabot settings to produce conventional commit messages / PR names

andylassiter:
Stumbled on [conventional-commits-action](https://github.com/marketplace/actions/conventional-commits-action) in the GitHub Action Marketplace. There are [other similar actions in the marketplace](https://github.com/marketplace?query=conventional+commits&type=actions) too.

PR #568, opened 2026-07-28
ci: enforce Conventional Commit PR titles

andrey-embarklabs:
Adds a PR Title Lint check (amannn/action-semantic-pull-request) that validates every PR title against the Conventional Commit spec. Squash merges use the PR title as the commit subject, so this makes the commit type a reliable, machine- parseable contract, the input the ADR 0030 release lane will use to compute the version (fix -> patch, feat -> minor, `!` -> major) and changelog.

First step of Phase 1 (release automation); release-please wiring builds on this. Uses the default Conventional Commit types; runs on synchronize too so the check reports on every head SHA and can be a required status check.

### Technical
Adds a PR Title Lint check validating every PR title against the Conventional
Commit spec. Squash merges use the PR title as the commit subject, so this makes
the commit type a reliable, machine-parseable contract, the input the ADR 0030
release lane computes the version (fix→patch, feat→minor, `!`→major) and changelog
from. First step of Phase 1 (release automation); release-please wiring builds on it.

## Type of change
- [x] Improvement

### Backward compatibility
Enforcement only starts once the check is marked required (see reviewer note). Uses
default Conventional Commit types; the one historical bare `security:` title becomes
`fix(security):`/`feat(security):` (already the repo's pattern).

## Testing
Workflow YAML validated; the action reads the title from the PR event payload (no
checkout). This PR's own title is the first live case.

## Note for reviewers
Two admin steps make this real (they can't be done in-repo):
1. Mark **PR Title Lint / lint-title** a required status check on `main`.
2. Set merge method to **squash-only** and enable **require linear history** on `main`
   (ADR 0030: squash keeps the subject = PR title, and the build lane's run_number
   ordering guard needs linear history).

### DIFF
--- a/.github/workflows/pr-title-lint.yaml
+++ b/.github/workflows/pr-title-lint.yaml
@@ -1,0 +1,24 @@
+name: PR Title Lint
+
+# PR titles are the source for the release lane's version + changelog (ADR 0030):
+# `fix:` -> patch, `feat:` -> minor, a `!` breaking marker -> major. Squash merges
+# use the PR title as the commit subject, so this gates exactly what the release
+# automation will parse. Validates against the default Conventional Commit types
+# (feat, fix, docs, style, refactor, perf, test, build, ci, chore, revert).
+on:
+  pull_request:
+    # `synchronize` re-runs on every push so the check reports on each head SHA
+    # (required-check status must exist on the SHA being merged); `edited` re-runs
+    # when the title itself changes.
+    types: [opened, edited, synchronize, reopened]
+
+permissions:
+  pull-requests: read
+
+jobs:
+  lint-title:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@48f256284bd46cdaab1048c3721360e808335d50 # v6.1.1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


johnflavin:
> Mark PR Title Lint / lint-title a required status check on main.

This doesn't make sense to me, as PR Title Lint doesn't run on `main`, only on PRs. It _can't_ run on `main` as by that point we've already committed. Enforcement on PRs is the only way this can ever run.

.github/workflows/pr-title-lint.yaml
```diff
@@ -0,0 +1,24 @@
...
+permissions:
+  pull-requests: read
+
+jobs:
+  lint-title:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@48f256284bd46cdaab1048c3721360e808335d50 # v6.1.1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
johnflavin, line 24: Why does this need a GH token if it doesn't need to clone and doesn't post back a status?

andylassiter:
Would this close #476?

andrey-embarklabs:
> Would this close #476?

partially, this is the CI-enforcement piece. #476 also mentions CONTRIBUTING/CLAUDE.md notes, a pre-commit commitlint, and renovate/dependabot title config, which this doesn't touch. i can add Closes #476 + file quick follow-ups for those, or leave it open to track them, your call.

.github/workflows/pr-title-lint.yaml
andrey-embarklabs: the action requires it, it sets up an authenticated octokit client on startup and errors if GITHUB_TOKEN is unset (their README notes this for fork PRs). we don't use the WIP-status feature (the only piece needing pull-requests: write), so read-only + the default token is the least it'll run with

andrey-embarklabs:
> > Mark PR Title Lint / lint-title a required status check on main.
> 
> This doesn't make sense to me, as PR Title Lint doesn't run on `main`, only on PRs. It _can't_ run on `main` as by that point we've already committed. Enforcement on PRs is the only way this can ever run.

yeah, agreed, it only runs on PRs. poor wording on my part: I meant add it to main's branch-protection required-checks list, which is the setting that blocks a PR from merging until the check passes. it runs on the PR, never on main itself.

johnflavin (approved):

PR #570, opened 2026-07-29
feat(xnat): chart 3.0.0 adoption - chart-native plugin install, deploy hardening, dev/test tooling

kathrynalpert:
### Product
Lets Scout deploy the optional XNAT imaging platform (`enable_xnat`) on the upstream NrgXnat Helm chart **3.0.0**, with the plugin set (OpenID SSO, container service, DICOM query-retrieve, etc.) installed by the chart itself, plus tooling to stand up and validate multi-node scenarios on a dev cluster. Most of this is gated behind `enable_xnat` and the `xnat_dev_*` dev flags; the only non-gated parts are the chart version pin and the removal of a now-unused CI image.

### Technical
- **Adopts XNAT chart 3.0.0**: plugins install natively through the chart (`url`/`coordinates`/`file`/`s3`/`image`), console logging via the chart's `loggingConfig`, and the dev WAR/plugin side-load runs through the chart's S3 support (`devWar` / `devPlugins` / `pluginInstaller.s3`).
- **Maven `coordinates` are release-only**: the chart resolves a coordinate to a URL at render time (no network), so a `-SNAPSHOT` coordinate — whose filename is timestamped and known only to `maven-metadata.xml` — is rejected. Scout's old `source.snapshots` opt-in and the Nexus MIXED-versionPolicy plumbing are removed; ship snapshots via `url`/`s3`/`file`/`devPlugins`.
- **Retires Scout's bespoke `xnat-plugin-installer` image** and all its CI/release wiring (ADR 0027 → Superseded).
- **Deploy correctness / hardening:** container-service on a Kubernetes backend (no `docker.sock`), a valid alphanumeric `siteId` default, a same-origin CSP so XNAT can frame itself, a Traefik sticky-session cookie for multi-replica login affinity, and Keycloak `xnat`-client redirect fix.
- **Dev/test tooling (flag-gated):** local WAR + plugin side-load staged to MinIO, an optional external ActiveMQ broker, inventory-driven replica count, and re-staging that re-rolls the StatefulSet.

## Type of change
- [x] Work behind a feature flag
- [x] New feature
- [ ] Improvement
- [x] Bug fix
- [x] Documentation update
- [ ] Test update

##### Authorization
XNAT authenticates via Keycloak OpenID (the realm `xnat` client) behind the oauth2-proxy edge gate (ADR 0003/0026) — posture unchanged. This PR only adjusts the `xnat` client's allowed redirect URIs so SSO login/logout complete; no new realm roles and no change to the OPA/Trino authorization model. Container-service now runs as Kubernetes Jobs via a scoped ServiceAccount (chart RBAC) instead of the host Docker socket.

##### Appsec
- Same-origin framing via a dedicated Traefik `security-headers-sameorigin` middleware (`frame-ancestors 'self'`); the global `frame-ancestors 'none'` is unchanged.
- No secrets in templates — the realm uses Jinja vars; the dev ActiveMQ broker uses a throwaway credential on synthetic dev clusters only.
- The sticky-session cookie is per-browser affinity (Traefik), not an auth control.

### Performance
XNAT is single-replica by default. The replica-count knob + external ActiveMQ exist for race/replay testing on dev clusters; a genuine multi-node deployment additionally needs RWX `archive`/`build` storage (documented in the role README). No production performance change.

### Data
No data-model changes. The chart 3.0.0 adoption requires a one-time StatefulSet recreate (see Backward compatibility); the CNPG Postgres DB and the named data PVCs are separate objects and are preserved.

### Backward compatibility
**Two one-time breaking changes for existing XNAT installs, both requiring operator action:**

1. **Chart 3.0.0 moves the StatefulSet selector** off the version-pinned labels to `app.kubernetes.io/{name,instance}`. StatefulSet selectors are immutable, so an existing install cannot be upgraded in place — delete the StatefulSet once (`kubectl -n <ns> delete statefulset xnat`), then deploy. Pods restart; Postgres + data PVCs are untouched; future upgrades roll normally. Chart 3.0.0 is **published** to `oci://ghcr.io/nrgxnat/charts/xnat` (nrgxnat/helm-charts#38).
2. **`xnat-plugin-installer` image removed.** Plugin install + console logging are now chart-native. `ghcr.io/washu-tag/xnat-plugin-installer` is no longer built/published and the `xnat_plugin_installer_image_*` role vars are gone.

## Testing
Validated end-to-end on dev04:
- Fresh deploy on chart 3.0.0 (from local disk): pods reach Ready, ROOT.war deploys, every plugin installs via the chart, the XNAT REST API answers, and container-service registers its Kubernetes DockerServer.
- Dev side-load — local WAR + plugin JARs stage to MinIO and are pulled by the chart's S3 init containers; re-staging re-rolls the StatefulSet.
- Loaded a plugin via each chart source and confirmed all landed in `/data/xnat/home/plugins/`: `coordinates` (jFrog fat-jar through the Nexus Maven proxy), `url` (a GitHub release — the nominally-air-gapped dev cluster additionally needed public-CA trust + a `githubusercontent.com` DNS forward to reach the release CDN, neither shipped in this PR), and `file` (jar → Secret → chart mount).
- External ActiveMQ messages received and delivered.

## Note for reviewers
N/A

## Checklist
- [x] My code adheres to the coding and style guidelines of the project (ran `pre-commit` on changed files; run `--all-files` before un-drafting)
- [x] I have added or updated user documentation, if appropriate (role READMEs)
- [x] I have added or updated technical documentation, including an architectural decision record, if appropriate (runbook `docs/internal/xnat-develop-testing.md`; ADR 0027 → Superseded — a formal successor ADR is a candidate follow-up)
- [x] I have added unit tests, if appropriate (`xnat_maven`)
- [ ] I have added end-to-end tests, if appropriate (validated manually on a dev cluster; none automated)


ansible/group_vars/all/versions.yaml
```diff
@@ -169,15 +169,22 @@ orthanc_image: jodogne/orthanc-plugins
...
-xnat_image_tag: 1.10.0
+# GHCR monorepo lane (ghcr.io/nrgxnat/xnat, set in the xnat role defaults); the
+# Docker Hub xnatworks/xnat-web images are no longer pulled.
+# TEMPORARY PIN: 1.10.1-SNAPSHOT is the only 1.10 tag published to GHCR. The
+# chart's own appVersion (1.9.3.7) is a 1.9 image, and Scout's plugins are built
+# against 1.10 (iq-plugin-...-xnat1.10, openid 1.6.0), so 1.9 is not an option.
+# Repin to a 1.10 release tag once one is published — Renovate cannot bump a
+# -SNAPSHOT, so this pin is effectively unmonitored until then.
+# renovate: datasource=docker depName=ghcr.io/nrgxnat/xnat
+xnat_image_tag: 1.10.1-SNAPSHOT
```
johnflavin, line 180: I don't like having to pin a snapshot version here. But I wasn't able to launch XNAT with the xnatworks 1.10.0 image, but I was with the nrgxnat 1.10.1-SNAPSHOT image which fixed some bugs I saw around file permissions and ownership. (There isn't an nrgxnat 1.10.0 or I would have used it.)

ansible/roles/xnat/defaults/main.yaml
```diff
@@ -76,3 +86,58 @@ xnat_extra_config: {}
...
+xnat_dev_war: '' # path to a local ROOT.war (empty = feature off)
+xnat_dev_plugins: [] # local plugin .jar paths (replace vs add: set xnat_plugins accordingly)
+xnat_dev_bucket: xnat-dev # MinIO bucket holding the staged dev artifacts
+# The chart pulls the staged artifacts with the AWS CLI (S3-compatible, pointed at
+# MinIO via endpointUrl), replacing the mc client the role used to run in its own
+# init containers. The chart's own default is public.ecr.aws/aws-cli/aws-cli, but
+# that registry isn't in Scout's air-gapped Harbor proxy set (harbor_proxies) —
+# docker.io is — so default to Docker Hub's mirror of the same tool, and on-prem
+# air-gapped clusters work without a per-inventory override.
+xnat_dev_awscli_image: amazon/aws-cli:latest
```
johnflavin, line 108: I'm not sure about this. I am guessing it is harmless so long as the image is an actual mirror of the `public.ecr.aws` image. But I'd prefer to add that registry to the harbor proxy and pull the real thing rather than override.

Maybe this is a "Nit"? Not sure.

ansible/roles/xnat/defaults/main.yaml
```diff
@@ -76,3 +86,58 @@ xnat_extra_config: {}
...
+# XNAT runs an embedded (per-JVM) broker by default, so multi-replica pods each
+# hold their OWN message queue and don't coordinate cluster events. Set true to
+# deploy a standalone ActiveMQ Artemis broker (tasks/dev_activemq.yaml) and point
+# every replica at it via spring.activemq.* (values.yaml.j2 -> xnat-conf.properties),
+# so cluster events fan out over one shared queue — useful for multi-node
+# race/replay testing. Not for prod: a single ephemeral broker, no HA or
+# persistence (and true multi-node also needs RWX archive/build storage, per the
+# container-service note above).
+xnat_dev_activemq: false
+xnat_dev_activemq_image: apache/activemq-artemis:latest
```
johnflavin, line 139: Do we really need to override the image if we aren't deploying activemq? And why a `latest` image rather than a pin?

johnflavin (approved):
I've deployed this to my dev stack and found that it worked, but only with some tweaks that I have pushed up. I also have _not_ deployed with multiple replicas, activemq, or container service, so I can't vouch for those.

andylassiter (approved):

PR #574, opened 2026-07-30
ci: publish build-lane image tags and capture digests (phase 2)

andrey-embarklabs:
First slice of Phase 2 (ADR 0030 build lane). Today images publish only the derived tag (mostly `latest`) and the pushed digest is thrown away, so nothing can pin an image immutably. This adds the foundation the build manifest needs, without changing what deploys.

- docker-push: new optional `version` input. When set, it additionally tags and pushes the immutable build-lane tag (0.YYYYMMDD.<run>, minted once in the changes job), captures the registry digest, and writes "<name> <ref>@<digest>" as an `image-digest-<name>` artifact for the manifest job to aggregate. When empty (demo publishes) it behaves exactly as before, so publish-demo is untouched.
- publish (main only): pass the build-lane version to all nine image pushes.

Keeps the existing build-tar flow (scan-images + in-cluster image import both consume that tar) rather than switching to push-by-digest. Dual-publishing the derived tag alongside the build tag is intentional until the ADR 0031 cutover. No new permissions (packages:write already present); no deploy behavior changes.

### Technical
First slice of Phase 2 (ADR 0030 build lane). Images currently publish only the
derived tag and discard the pushed digest, so nothing can pin an image immutably.

- docker-push: optional `version` input. When set, additionally tags+pushes the
  immutable 0.YYYYMMDD.<run> tag, captures the registry digest, and uploads an
  `image-digest-<name>` artifact ("<name> <ref>@<digest>") for the manifest job.
  Empty (demo publishes) = unchanged behavior; publish-demo is untouched.
- publish (main only): pass the build-lane version to all nine image pushes.

Keeps the build-tar flow (scan-images + in-cluster import consume the tar) rather
than push-by-digest. Dual-publishing the derived tag alongside the build tag is
intentional until the 0031 cutover. No new permissions; no deploy changes.

## Type of change
- [x] Improvement

## Note for reviewers
This is the foundation the build manifest (a later PR) aggregates. Digest is parsed
from `docker push` output; keeping the tar was decision 1 from the Phase 2 design
(push-by-digest would break the tar consumers). Only exercises on merge to main
(publish is main-gated).

### DIFF
--- a/.github/actions/docker-push/action.yaml
+++ b/.github/actions/docker-push/action.yaml
@@ -1,9 +1,23 @@
 name: 'Docker push'
-description: 'Downloads docker image and pushes to ghcr'
+description: >-
+  Loads the built image tar and pushes it to ghcr at the derived tag. When a
+  build-lane version is given, also publishes an immutable 0.YYYYMMDD.<run> tag
+  and records the pushed digest for the ADR 0030 build manifest.
 inputs:
   image-name:
     description: 'Name for docker image'
     required: true
+  version:
+    description: >-
+      Build-lane version (0.YYYYMMDD.<run>) to publish alongside the derived
+      tag. Leave empty (e.g. demo publishes) to push only the derived tag and
+      record no digest.
+    required: false
+    default: ''
+outputs:
+  digest:
+    description: 'Registry digest (sha256:...) of the pushed image; empty when no build-lane version was given.'
+    value: ${{ steps.push.outputs.digest }}
 runs:
   using: 'composite'
   steps:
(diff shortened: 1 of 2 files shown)


.github/actions/docker-push/action.yaml
```diff
@@ -13,11 +27,48 @@ runs:
...
+        fi
+
+        # Also publish the immutable build-lane tag so the build manifest can pin
+        # this image by name:tag@digest. It aliases the same image, so the registry
+        # returns the same content digest.
+        repo="${loaded_ref%:*}"
+        build_ref="${repo}:${VERSION}"
+        docker tag "${loaded_ref}" "${build_ref}"
+        push_out="$(docker push "${build_ref}")"
+        digest="$(printf '%s\n' "${push_out}" | sed -n 's/.*digest: \(sha256:[0-9a-f]\{64\}\).*/\1/p' | head -1)"
```
andylassiter: Minor finding from Claude.
> This parses human-facing CLI output, which is not a stable contract. The hard-fail guard below it means a Docker CLI format change would break the build loudly (good — not silently wrong), but a more robust source of truth is docker buildx imagetools inspect "${build_ref}" --format '{{.Manifest.Digest}}' or crane digest. Given the PR explicitly chose to keep the tar flow and the parse is defensively guarded, this is acceptable as-is

andylassiter (approved):

.github/actions/docker-push/action.yaml
andrey-embarklabs: good call, hardened it. reads the digest back from the registry now via docker buildx imagetools inspect "$ref" --format '{{json .Manifest}}' | jq -r .digest instead of parsing the push output. one heads-up: {{.Manifest.Digest}} prints the default view with our buildx (0.30.x), so I used the json form to drill in cleanly. jq and buildx are both already in the pipeline.

PR #579, opened 2026-07-30
feat(temporal): make default namespace workflow retention configurable

### LINKED ISSUE #432: Temporal Namespace retention should be configurable
Note: I am talking about the namespace temporal uses internally for managing workflows, _not_ the K8s namespace under which temporal is deployed.

We use only a `default` namespace in temporal that gets a default of 3 days for workflow retention. I would like this to be much higher on my dev stack so that if I come back to temporal after a week my historical workflows are still there. It might end up needing to be longer in a production environment for auditing reasons.

cmoore01:
## Description

### Product
Temporal's workflow-history retention is now configurable per environment instead of fixed at 3 days. Operators can raise it on dev so historical workflows survive time away, and on prod for longer audit windows. The default is unchanged (3 days), so existing deployments see no behavior change.

### Technical
Adds a small Scout-owned Helm chart, `temporal-bootstrap`, deployed as its own release. Its namespace-retention Job runs `temporal operator namespace update -n default --retention <value>` as a post-install/post-upgrade hook, driven by the new `temporal_default_namespace_retention` var (default `3d`).

A hook Job is required (rather than just declaring retention in the Temporal chart values) because the upstream chart's namespace-setup job is create-only (`describe || create`) — it never updates retention on an already-existing namespace. The hook runs on every release, so retention converges on both fresh and existing clusters and self-corrects drift.

It's modeled on the `scout-dashboards` import chart (declarative config + reconciling hook Job) rather than a bespoke Ansible Job, so it carries over verbatim to a Flux `HelmRelease` under the ADR 0031 GitOps migration with no rewrite. The upstream Temporal chart and its values are left untouched — the bootstrap chart is the single retention mechanism. Wired via role defaults plus three `deploy.yaml` tasks (hash → checksum → `deploy_helm_chart`) after the schema-wait.

## Type of change
- [X] New feature
- [ ] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

## Impact

### Security 

##### Authorization
No new user-facing roles or authz surface. The hook Job talks to the *internal* Temporal frontend (`temporal-internal-frontend:7236`), which bypasses the OIDC authz on the public frontend — the same pattern Scout's existing prospective-indexing setup Job already uses. It runs with a hardened securityContext (non-root uid 1000, `readOnlyRootFilesystem`, all caps dropped, seccomp `RuntimeDefault`). Retention is a namespace-level operational setting;                                                                                                                                                                                                                                                                                                    no REST paths or user data are involved.

##### Appsec
Minimal-privilege Job on a configurable admin-tools image; no cross-namespace references (release and Job live in the extractor namespace). A `required` guard on `temporalAddress` prevents an empty address silently falling back to localhost.

### Performance
Negligible at runtime — one short-lived hook Job per Temporal release (the describe-until-exists wait caps at ~60s). The real cost is data-side: longer retention increases Cassandra (history) and Elasticsearch (visibility) storage roughly linearly, so long prod audit windows are a storage-capacity input, not a runtime concern.

### Data
Retention governs how long completed workflow histories persist before Temporal's scavenger deletes them. No parsing or data-shape handling. Default 443d is unchanged.

### Backward compatibility
Additive and default-preserving. New separate Helm release (`temporal-bootstrap`); the upstream Temporal chart/values are unchanged. No data-model, API, or dependency changes. The hook is idempotent — safe to re-run deploy.

## Testing
- `helm lint` / `helm template` / `helm show chart` on the chart (renders clean; Chart.yaml description intact; `namespace.retention` override applies).
- Dev deploy on dev01: `temporal operator namespace update` applied and `namespace describe` reflected the value.
- **Verified `retention=60d` is accepted with no max-retention error**, confirming the modern Temporal server (chart ~1.2.x) has no 30-day cap.
- Idempotent re-run: the hook re-runs on each release and converges an existing namespace (the case the upstream create-only job misses).

## Note for reviewers
- Deliberately mirrors `scout-dashboards` (declarative + reconciling hook Job) so it ports to a Flux `HelmRelease` (ADR 0031) with no rewrite; chose this over an Ansible-inline Job / `extraObjects` because the create-only upstream job needs an update-on-every-release mechanism and a chart is the natural GitOps home. 
- Correction from early design discussion: modern Temporal has **no** max-retention cap (verified in `common/dynamicconfig/constants.go` — only `system.namespaceMinRetention*` remains; `60d` tested), so there is no `maxRetentionDays` to raise. Two commits: the feat, plus a `docs` commit fixing an earlier comment that wrongly referenced it.
- The prospective-indexing schedule remains a bespoke Ansible Job for now; it can move into this chart as a follow-up (this PR establishes the pattern).

## Checklist


ansible/roles/temporal/defaults/main.yaml
```diff
@@ -19,5 +19,21 @@ temporal_resources_default:
...
+# the minimum (system.namespaceMinRetentionGlobal/Local, ~24h).
+temporal_default_namespace_retention: 3d
+
+# temporal-bootstrap chart: idempotent Temporal setup applied via Helm-hook
+# Jobs, deployed as a separate release (mirrors the scout-dashboards model).
+# Currently enforces the namespace retention above; future one-off setup can be
+# added as additional hook Jobs in the chart.
+temporal_bootstrap_helm_chart_ref: '{{ scout_repo_dir }}/helm/temporal-bootstrap'
+temporal_admin_tools_image_repository: temporalio/admin-tools
+temporal_admin_tools_image_tag: latest
```
andylassiter: Should this be `latest`? We have `temporal_version: ~1.2.0` set in `versions.yaml` (but that is the helm chart version not the image version). Claude is saying Helm chart v1.2.0 resolves to admin tools image 1.31.0. Also maybe this should be moved to `versions.yaml` and tracked with renovate?

andylassiter (approved):
`temporal-bootstrap` helm chart needs to be added to our CI now after Andrey's GitOps PRs.

```
Fix: add the three entries mirroring scout-dashboards:
# outputs:
temporal-bootstrap-chart: ${{ steps.filter.outputs.temporal-bootstrap-chart }}
# filters:
temporal-bootstrap-chart:
  - 'helm/temporal-bootstrap/**'
# publish-charts matrix:
- chart-name: temporal-bootstrap
  chart-dir: helm/temporal-bootstrap
  changed: temporal-bootstrap-chart
```

ansible/roles/temporal/defaults/main.yaml
```diff
@@ -19,5 +19,21 @@ temporal_resources_default:
...
+# chart's prior default. Bump on dev for durability, or on prod for audit;
+# modern Temporal (chart ~1.2.x) has no max-retention cap -- the only bound is
+# the minimum (system.namespaceMinRetentionGlobal/Local, ~24h).
+temporal_default_namespace_retention: 3d
+
+# temporal-bootstrap chart: idempotent Temporal setup applied via Helm-hook
+# Jobs, deployed as a separate release (mirrors the scout-dashboards model).
+# Currently enforces the namespace retention above; future one-off setup can be
+# added as additional hook Jobs in the chart.
+temporal_bootstrap_helm_chart_ref: '{{ scout_repo_dir }}/helm/temporal-bootstrap'
```
johnflavin, line 34: I believe we've moved from deploying local helm chart directories to publishing helm charts as `oci` and deploying from those.

(Checking...)

Ok, we've done this partially. All the local charts get built in CI and published (that's Andrey's Phase 0) but only a few of them currently get deployed from the published chart. So we should add this to be published.

johnflavin (approved):

helm/temporal-bootstrap/values.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+nameOverride: ''
+fullnameOverride: ''
+
+# admin-tools image providing the `temporal` CLI. Should track the Temporal
+# server version.
+image:
+  repository: temporalio/admin-tools
+  tag: latest
+  pullPolicy: IfNotPresent
```
johnflavin: 1. This shouldn't be `latest` for a couple reasons.
1a. Just generally we should pin the version of the images we pull.
1b. The comment above says it should track the temporal server version. I'm pretty sure that is pinned elsewhere, either directly in an image pin or by pinning a chart version which itself pins an image, so we could likely reuse that pin wherever it lives.
2. This is more speculative, but doesn't the underlying temporal chart have this version defined somewhere? Could we reuse that? I'm really not sure if it works that way, if we could reach into a chart we're depending on to get values out of it, but it might be nice. Either way, 1 is probably the best solution.

helm/temporal-bootstrap/values.yaml
```diff
@@ -0,0 +1,44 @@
...
+# authz on the public frontend, matching Scout's other in-cluster setup Jobs.
+# Required — the deploying role sets this to the in-namespace FQDN.
+temporalAddress: ''
+
+namespace:
+  name: default
+  # Workflow-history retention. Accepts the Temporal CLI duration form
+  # ("3d", "30d", "90d", ...). Modern Temporal has no max-retention cap; the
+  # only bound is the minimum (system.namespaceMinRetentionGlobal/Local, ~24h).
+  retention: 3d
```
johnflavin: > The default is unchanged (3 days), so existing deployments see no behavior change

I'm not convinced by this. I think we have a need across all our deployments for longer retention, so preserving the old behavior doesn't seem like a benefit to me. It's not like there is a risk of breakage, the whole point of the job is to be able to change an existing value, so why not just bump it up to a value that would be more useful?

For our deployments I'd like it to be at least a few weeks, maybe over a month. If we want the default to be lower still a week seems like the minimum amount of time we'd want.

helm/temporal-bootstrap/templates/namespace-retention-job.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+Runs as a post-install/post-upgrade hook with before-hook-creation delete, so
+it re-executes on every `helm upgrade` -- the same reconciling pattern as
+scout-dashboards' import Job, and it carries over verbatim to a Flux HelmRelease
+(ADR 0031) with no rewrite.
+*/ -}}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ include "temporal-bootstrap.fullname" . }}-namespace-retention
```
johnflavin, line 19: This `name` is a problem. If we run this deploy and render this chart, it will create a job. Whether that job is successful or fails, it lives for its TTL. If we render the chart again, it will fail to render because the job with this name already exists.

I had initially wanted to say "use `generateName` instead". But I had a suspicion that helm does not support it, and I found evidence that they don't: helm/helm#3348. That issue does point to a workaround in some of the comments: "use `{{ .Release.Revision }}` on top of the job name", i.e. use
```
name: {{ include "temporal-bootstrap.fullname" . }}-namespace-retention-{{ .Release.Revision }}
```
That way helm will generate a unique per-invocation name which avoids collisions.

But you don't have to take my word for it! 📖 🌈 
You should ask Claude if this makes sense.

helm/temporal-bootstrap/values.yaml
cmoore01: It's just a placeholder here though, right? We're specifying the version in `deploy.yaml`.

johnflavin: > It's just a placeholder here though, right? We're specifying the version in `deploy.yaml`.

Ok, I can see that. So maybe this version doesn't need to be pinned here; we have it as a placeholder that gets pinned externally in Scout's ansible renovate stuff. But does it need to be tied to the underlying temporal chart/image version? Like independently of Scout's ansible, just as a chart on its own, this `latest` version could drift from the version of temporal that is installed in an incompatible way and break things. Do we need to / is there a way to enforce that version compatibility in the chart? For instance, I notice the `Chart.yaml` doesn't define the temporal chart as a dependency. Should it? If it did, would this chart then have access to the version information in that dependent chart and be able to force compatibility? And would that even be a good idea, given that we're ansible deploying these things?

helm/temporal-bootstrap/templates/namespace-retention-job.yaml
cmoore01: Could you double check this please? I don't think it's quite right. We're specifying `"helm.sh/hook-delete-policy": before-hook-creation` so it already cleans up the existing job when rerunning, as https://github.com/washu-tag/scout/blob/main/helm/scout-dashboards/templates/import-job.yaml does.

helm/temporal-bootstrap/values.yaml
cmoore01: Good point, I'll bump it to 30d.

helm/temporal-bootstrap/templates/namespace-retention-job.yaml
johnflavin: Yep, I think you’re correct. Sorry for the confusion, can disregard.

johnflavin (approved):

ansible/roles/temporal/defaults/main.yaml
cmoore01: Addressed, thanks

helm/temporal-bootstrap/values.yaml
cmoore01: Thanks, I think we landed on a better method

PR #598, opened 2026-08-04
ci: authenticated ghcr carry for private packages in build_haul

andrey-embarklabs:
ghcr_digest gains a github_token param: with a token it authenticates (GHCR accepts the base64-encoded token as the bearer), so first-push private chart and image packages resolve; without one the lookup stays anonymous. main() reads GITHUB_TOKEN from the env and the publish-haul render step passes it. Removes one of the two chart-enablement blockers (private packages); the other is publishing every chart at least once.

### Technical

`build_haul` carries unchanged components by looking up their current registry digest. That lookup was anonymous, so a first-push **private** ghcr package (which is how not-yet-public charts start life) would 403 and the carry would fail. This adds an authenticated path:

- `ghcr_digest` gains a `github_token` parameter. With a token it authenticates (GHCR accepts the base64-encoded token as the bearer), so private chart/image packages resolve; without one the lookup stays anonymous (public packages only), unchanged.
- `main()` reads `GITHUB_TOKEN` from the environment, and the `publish-haul` render step now passes `GITHUB_TOKEN: ${{ github.token }}`.

This removes one of the two chart-enablement blockers documented in `components.txt` (private packages). The other (publishing every chart at least once) is separate.

## Type of change
- [x] Improvement

## Testing
Added `test_ghcr_digest_authed_uses_base64_bearer_and_skips_token_endpoint`: the authed path sends `Authorization: Bearer <base64(token)>` on the manifest HEAD and does not hit the anonymous token endpoint. Full offline suite: 34 passed.

## Note for reviewers
The authed path is unit-tested with an injected opener but not integration-tested against a real private washu-tag package (no token on hand). It follows GHCR's documented base64-token-as-bearer convention. `build_haul` isn't wired into a running job yet (publish-haul is still the inert draft), so this is preparatory.

## Checklist
- [x] My code adheres to the coding and style guidelines of the project (`pre-commit` black/prettier)

### DIFF
--- a/.github/workflows/publish-haul.yaml
+++ b/.github/workflows/publish-haul.yaml
@@ -67,6 +67,10 @@ jobs:
           merge-multiple: true
 
       - name: Render the Hauler manifest
+        env:
+          # Authenticated carry of unchanged components so private (first-push)
+          # chart/image packages resolve; build_haul reads this from the env.
+          GITHUB_TOKEN: ${{ github.token }}
         run: |
           python3 tooling/manifest/build_haul.py \
             --digests-dir digests \
--- a/tooling/manifest/build_haul.py
+++ b/tooling/manifest/build_haul.py
@@ -15,7 +15,9 @@ is agnostic to the choice, so only ``carry`` below changes.
 from __future__ import annotations
 
 import argparse
+import base64
 import json
+import os
 import sys
 import urllib.request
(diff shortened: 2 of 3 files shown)


andrey-embarklabs:
closing in lieu of https://github.com/washu-tag/scout/pull/601

PR #601, opened 2026-08-05
ci: carry unchanged components from the predecessor haul

andrey-embarklabs:
Prove-out showed a fixed carry tag can't work: superset and keycloak have no :latest, and superset has no build-lane tag either. Replace the live tag-lookup carry with predecessor-haul carry: build_haul reads the previous build's haul manifest (--predecessor, which the workflow oras-pulls) and carries each unchanged component at the digest it had there. Removes ghcr_digest and the network carry path, so build_haul is now pure/offline. The first build has no predecessor, so it must be a full rebuild (every component fresh). Roadmap step 5a; supersedes the carry-auth spike (ci/ghcr-digest-auth).

### Technical
Prove-out showed a fixed carry tag can't work (superset/keycloak have no `:latest`; superset has no build-lane tag). This replaces the live tag-lookup carry with **predecessor-haul carry**: `build_haul` reads the previous build's `haul.yaml` (`--predecessor`, which the haul job oras-pulls) and carries each unchanged component at the digest it had there. Removes `ghcr_digest` and the network path, so `build_haul` is pure/offline. The first build has no predecessor, so it must be a full rebuild.

## Type of change
- [x] Improvement
## Testing
35 offline tests, including a render→parse round-trip and fail-closed when a component is neither fresh nor in the predecessor.
## Note for reviewers
Supersedes `ci/ghcr-digest-auth` (the carry-auth spike) — that PR can be closed; auth is no longer needed in `build_haul` since the job pulls the predecessor.
## Checklist
- [x] Code adheres to guidelines
- [x] Commented hard-to-understand areas
- [x] Unit tests


.github/workflows/ci.yaml
```diff
@@ -1103,3 +1103,93 @@ jobs:
...
         with:
           image-name: report-viewer
+
+  publish-haul:
+    # Build-lane haul (ADR 0033): after images publish, render the manifest from
+    # this run's fresh digests plus the previous build's haul (carry), sync it into
+    # a .tar.zst bundle, and publish the bundle + the manifest so the next run can
+    # carry from it. continue-on-error so it never gates a release; it fails closed
+    # (harmless) on partial builds until a one-time bootstrap seeds the first
+    # predecessor. Sync is unsigned for now; `--key` verification is a follow-up
```
johnflavin, line 1154: > continue-on-error so it never gates a release

Why not block release on error? If this errors it means we didn't build/publish the manifest that in some sense _is_ the release, right? Or am I not understanding what this is doing?

johnflavin (commented):
I don't see why we would merge this before #602. This one spends some comment space and code logic maneuvering around the fact that there is no "predecessor" on the first run. But if we merge #602 first we could do a manual dispatch to seed the first run before this one ever runs automatically, which means the predecessor would be there and we can skip all the workarounds.

andrey-embarklabs:
> I don't see why we would merge this before #602. This one spends some comment space and code logic maneuvering around the fact that there is no "predecessor" on the first run. But if we merge #602 first we could do a manual dispatch to seed the first run before this one ever runs automatically, which means the predecessor would be there and we can skip all the workarounds.

The dependency actually runs the other way: #602's bootstrap calls
`build_haul --predecessor`, which this PR introduces, so #602 can't run (or be
dispatched to seed) until this one is merged. That's why #602 is stacked on this
branch.

.github/workflows/ci.yaml
andrey-embarklabs: that's just to bootstrap it first. we'll have a gate after for sure

johnflavin (approved):

PR #602, opened 2026-08-05
ci: add the one-time build-lane haul bootstrap dispatch

andrey-embarklabs:
The steady-state haul job carries unchanged components from the previous haul, but the first haul has no predecessor and publish only pushes changed images, so no normal run has all 8 fresh. This workflow_dispatch resolves each image's current digest (superset/keycloak from a release tag since they have no :latest), signs it, and builds + publishes the first full haul to seed the predecessor.

No retag: docker buildx imagetools create rewraps the source into a new index with a different digest (verified locally), so we pin the source tag by @digest instead. Dispatch after the predecessor-carry PR (#601) lands build_haul --predecessor. Roadmap step 5b-2.

### Product

No product/user-facing change. Operational tooling to seed the air-gap build-lane haul once.

### Technical

One-time `workflow_dispatch` that produces the first build-lane haul (ADR 0033). The steady-state `publish-haul` job carries unchanged components from the previous haul, but the first haul has no predecessor, and `publish` only pushes changed images, so no normal run ever has all 8 images fresh. This job resolves each image's current registry digest, cosign-signs it, records it as fresh, and builds + publishes the first full haul (bundle + manifest, `:main` advanced last) so subsequent builds have a predecessor to carry from.

superset and keycloak have no `:latest` tag, so they seed from a release tag (dispatch inputs, default `5.0.0` / `26.6.3`). No retag is done: `docker buildx imagetools create` rewraps the source into a new image index with a different digest (verified locally), which would break digest fidelity, so the source tag is pinned by `@digest` instead (hauler copies by digest, the tag is only a label).

## Type of change
- [x] New feature

##### Authorization
None. No user roles or data access change.

##### Appsec
Uses the keyed cosign pair (`COSIGN_PRIVATE_KEY` / `COSIGN_PASSWORD`) to sign each seeded image and the bundle; secrets are passed via `env:`, never interpolated into `run:` scripts. Dispatch inputs are consumed as double-quoted shell values, so a crafted tag cannot inject commands. Signing is keyed with no transparency log, so private digests stay off public Rekor.

### Performance
One-time manual dispatch. No effect on normal CI.

### Data
N/A.

### Backward compatibility
Additive: a new `workflow_dispatch`-only workflow that never runs automatically. Nothing else references it.

## Testing
- `bootstrap-haul.yaml` parses; pinned `prettier` passes; no `${{ }}` in any `run:` block.
- Validated the render locally against the `--predecessor` `build_haul` (from #601): 8 images pinned, superset at `5.0.0`, keycloak at `26.6.3`, rest `:latest`, all `@sha256:`.
- Validated locally that `imagetools create` does not preserve the digest (the reason this pins the source tag instead).
- The full sign + sync + publish path runs when dispatched (after #601 merges).

## Note for reviewers
Dispatch-only and inert until run. Run it **after #601 merges** (it provides `build_haul --predecessor`). The sync/save/publish tail duplicates the `publish-haul` job in ci.yaml; a `.github/actions/publish-haul` composite can dedup both in a follow-up.

## Checklist

### DIFF
--- a/.github/workflows/bootstrap-haul.yaml
+++ b/.github/workflows/bootstrap-haul.yaml
@@ -1,0 +1,116 @@
+# One-time bootstrap for the build-lane haul (ADR 0033, roadmap step 5b-2).
+#
+# The steady-state publish-haul job (in ci.yaml) carries unchanged components
+# from the previous haul, but the FIRST haul has no predecessor, and the publish
+# job only pushes CHANGED images, so no normal run ever has all 8 fresh. This
+# dispatch resolves every image's current digest, signs it, and builds + publishes
+# the first full haul so subsequent builds have a predecessor to carry from.
+#
+# Run once (again only to re-seed from scratch). Dispatch after the predecessor-
+# carry + haul-job PR is merged (it provides build_haul --predecessor).
+#
+# The sync/save/publish tail duplicates the publish-haul job; if that grows, lift
+# it into a `.github/actions/publish-haul` composite used by both.
+name: Bootstrap build-lane haul
+
+on:
+  workflow_dispatch:
+    inputs:
+      superset_tag:
+        description: 'superset has no :latest; seed from this release tag'
+        default: '5.0.0'
+      keycloak_tag:
+        description: 'keycloak has no :latest; seed from this release tag'
+        default: '26.6.3'
+
+permissions:
+  contents: read
+  packages: write
+
+env:
+  REGISTRY: ghcr.io
+  MANIFEST_REPO: ghcr.io/washu-tag/manifests/scout-manifest
+  BUNDLE_REPO: ghcr.io/washu-tag/manifests/scout
+
+jobs:
+  bootstrap:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+        with:
+          python-version: '3.12'
+      - name: Login to GHCR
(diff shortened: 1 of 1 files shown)


.github/workflows/bootstrap-haul.yaml
```diff
@@ -0,0 +1,116 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      superset_tag:
+        description: 'superset has no :latest; seed from this release tag'
+        default: '5.0.0'
+      keycloak_tag:
+        description: 'keycloak has no :latest; seed from this release tag'
+        default: '26.6.3'
```
johnflavin, line 24: I'm a little unsure about embedding these versions here. I don't think anything auto-updates them and they're quite likely to drift from the deployed artifact versions.

.github/workflows/bootstrap-haul.yaml
```diff
@@ -0,0 +1,116 @@
...
+        uses: docker/login-action@371161bbe7024a29a25c5e19bfcbc0804fe9ad2c # v4.5.2
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
+      - name: Install hauler and oras
+        env:
+          HAULER_VERSION: '2.0.2'
+          ORAS_VERSION: '1.2.0'
```
johnflavin, line 53: Similarly, I'm not sure about putting the versions here. This is less impactful than the other since these are just build components, not deployed, but it's still another version embedded somewhere that will never be updated.

johnflavin (approved):
Just out of curiosity, I want to see what happens if you Rebase Stack to get it up-to-date with main rather than merge changes in.

PR #609, opened 2026-08-06
ci: verify component signatures at haul sync, with a completeness guard

andrey-embarklabs:
Turn on cosign verification in the haul job's `hauler store sync` (--key from COSIGN_PUBLIC_KEY), now that images and charts are signed at publish. hauler SKIPS an image it can't verify and still exits 0, so a naive --key would silently drop a component and publish an incomplete haul; guard against that by failing the step if the sync log reports any verification miss.

Validated locally against a throwaway registry: --key verifies a signed image, rejects an unsigned one, and the guard trips on the skip. The job stays continue-on-error, so a miss is non-blocking and visible rather than a wrong publish. Log strings are hauler 2.0.2's; recheck on a hauler bump.

### Technical
Turns on cosign verification in the haul job's `hauler store sync` (`--key` from `COSIGN_PUBLIC_KEY`), now that images + charts are signed at publish, so a tampered component can't be bundled.

Important hauler behaviour I verified locally: `hauler store sync --key` **skips** an image it can't verify and **still exits 0** ("processing completed successfully"), so a naive `--key` would silently drop a component and publish an *incomplete* haul. This adds a completeness guard, the step fails if the sync log reports any verification miss.

## Type of change
- [x] New feature (supply-chain: sync-time signature verification)

## Testing
Validated against a throwaway registry: `--key` logged "signature verified" and added a signed image, rejected an unsigned one ("no signatures found"), and the guard grep trips on the skip / passes a clean log. YAML + prettier clean.

## Note for reviewers
Depends on every carried image being signed (the bootstrap signed the seed; the sign-at-publish steps sign future builds). The job stays `continue-on-error`, so a verification miss is non-blocking and visible rather than a wrong publish. The guard's log strings are hauler 2.0.2's, recheck them on a hauler bump.

## Checklist
- [x] Adheres to guidelines
- [x] Commented (the skip-and-exit-0 gotcha + guard)
- [ ] Docs / tests (N/A, workflow)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1226,9 +1226,19 @@ jobs:
         env:
           COSIGN_PRIVATE_KEY: ${{ secrets.COSIGN_PRIVATE_KEY }}
           COSIGN_PASSWORD: ${{ secrets.COSIGN_PASSWORD }}
+          COSIGN_PUBLIC_KEY: ${{ secrets.COSIGN_PUBLIC_KEY }}
         run: |
           set -euo pipefail
-          hauler store sync -f haul.yaml
+          # Verify each component's cosign signature as it is pulled. hauler SKIPS
+          # an image it can't verify and still exits 0, so guard against a silently
+          # incomplete haul: fail if the sync log reports any verification miss.
+          # (Log strings are hauler 2.0.2's; recheck them when bumping hauler.)
+          printf '%s' "${COSIGN_PUBLIC_KEY}" > cosign.pub
+          hauler store sync -f haul.yaml --key cosign.pub 2>&1 | tee sync.log
+          if grep -qiE "signature verification failed|no signatures found" sync.log; then
+            echo "::error::a component failed cosign verification and was skipped; the haul would be incomplete"
+            exit 1
+          fi
           hauler store save -f "scout-${VERSION}.tar.zst"
           # Push both immutable :version artifacts and sign the bundle, THEN advance
           # the :main pointers last, so :main only moves after a complete, signed
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1236,14 +1236,14 @@ jobs:
         env:
           COSIGN_PRIVATE_KEY: ${{ secrets.COSIGN_PRIVATE_KEY }}
           COSIGN_PASSWORD: ${{ secrets.COSIGN_PASSWORD }}
-          COSIGN_PUBLIC_KEY: ${{ secrets.COSIGN_PUBLIC_KEY }}
         run: |
           set -euo pipefail
           # hauler 2.0.2 SKIPS an image it can't verify yet still exits 0, so an
           # exit-code or failure-string check fails OPEN. Fail CLOSED: require one
           # "signature verified" line per image in haul.yaml (dash-anchored count so
           # metadata.name isn't included). (Strings are 2.0.2's; recheck on bump.)
-          printf '%s' "${COSIGN_PUBLIC_KEY}" > cosign.pub
+          # Derive the public key from the signing key -- no separate COSIGN_PUBLIC_KEY secret.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -1219,9 +1219,19 @@ jobs:
...
+          # Verify each component's cosign signature as it is pulled. hauler SKIPS
+          # an image it can't verify and still exits 0, so guard against a silently
+          # incomplete haul: fail if the sync log reports any verification miss.
+          # (Log strings are hauler 2.0.2's; recheck them when bumping hauler.)
+          printf '%s' "${COSIGN_PUBLIC_KEY}" > cosign.pub
+          hauler store sync -f haul.yaml --key cosign.pub 2>&1 | tee sync.log
+          if grep -qiE "signature verification failed|no signatures found" sync.log; then
+            echo "::error::a component failed cosign verification and was skipped; the haul would be incomplete"
+            exit 1
+          fi
```
johnflavin: This fails open, meaning if the log string does change in a future version we wouldn't catch failures and would silently accept invalid signatures. Is there a way to make this fail closed? Is there a log string indicating success that we could key off of, otherwise we treat it as a failure?

.github/workflows/ci.yaml
```diff
@@ -1219,9 +1219,19 @@ jobs:
...
+          COSIGN_PUBLIC_KEY: ${{ secrets.COSIGN_PUBLIC_KEY }}
         run: |
           set -euo pipefail
-          hauler store sync -f haul.yaml
+          # Verify each component's cosign signature as it is pulled. hauler SKIPS
+          # an image it can't verify and still exits 0, so guard against a silently
+          # incomplete haul: fail if the sync log reports any verification miss.
+          # (Log strings are hauler 2.0.2's; recheck them when bumping hauler.)
+          printf '%s' "${COSIGN_PUBLIC_KEY}" > cosign.pub
+          hauler store sync -f haul.yaml --key cosign.pub 2>&1 | tee sync.log
```
johnflavin, line 1248: I don't know what this `sync` operation does. But it doesn't look like we undo whatever that is in the case of an invalid key, so hopefully hauler will not perform the `sync` if things aren't properly signed?

And in that case does `hauler sync` exit with a non-zero code? That may be a more reliable signal of failed verification than digging through its stdout.

johnflavin (commented):
Don't take these questions too seriously, this might be fine, I really don't know how hauler works or what it is supposed to do.

.github/workflows/ci.yaml
andrey-embarklabs: yeah, good catch, that grep fails open. flipped it to fail closed off the success string instead. hauler logs signature verified for image [<ref>] once per image on a good sig, so i grep -c those and require the count to equal the number of images in haul.yaml (the - name: entries, dash-anchored so it doesn't also count metadata.name). if any component fails or gets skipped it logs "verification failed" instead of "verified", the count comes up short, and we bail. so the default is now rejection rather than acceptance, and we're not depending on the failure wording staying stable anymore.

.github/workflows/ci.yaml
andrey-embarklabs: nope, and that's exactly the trap. checked it both ways, read the v2.0.2 source and ran the binary: a bad or missing sig under --key isn't fatal. cosign fails, hauler logs it, does continue to the next image, and the sync still returns nil so the process exits 0. main.go only exits 1 on a returned error and the verify path never returns one. so the exit code tells us nothing for the verify case. and yeah, it leaves a partial store behind: the unverified image is skipped right before it'd be added, so the store ends up missing that component while the command reports success. when i ran it against an unsigned image the store came out with no index.json and 0 blobs vs 65 blobs on a clean run, both exit 0. that's why i'm not leaning on the exit code and went with the verified-line count. the hard failures (network, store write) DO exit non-zero and set -e + pipefail already catch those; it's specifically the verify-skip that stays 0.

johnflavin (approved):

PR #610, opened 2026-08-06
feat: couple scout-dashboards appVersion to the superset image version

andrey-embarklabs:
scout-dashboards deploys ghcr.io/washu-tag/superset, but its appVersion was "latest" while it pinned image.tag: 5.0.0 explicitly, so the chart and the image it deploys could drift apart. Set appVersion to the superset image version (5.0.0), default image.tag to {{ .Chart.AppVersion }} (adding the fallback in the template), so the chart and its image move together. The deployed image is unchanged (superset:5.0.0, verified via helm template); the k8s version label is now accurate (was "latest") and consumers can drop the manual tag pin.

First chart in the appVersion->image coupling (ADR 0030). A Renovate manager to keep this appVersion current with apache/superset rides in the superset-tracking PR (ci/renovate-track-superset).

### Technical
scout-dashboards deploys `ghcr.io/washu-tag/superset` but its `appVersion` was `"latest"` while it pinned `image.tag: 5.0.0` explicitly, so the chart and the image it deploys could drift. This couples them:
- `Chart.yaml` `appVersion`: `latest` → `5.0.0` (the superset image version)
- `values.yaml` `image.tag`: `5.0.0` → `''` (defaults to appVersion)
- template: `image.tag` → `image.tag | default .Chart.AppVersion`

The deployed image is unchanged (`superset:5.0.0`, verified with `helm template`); the `app.kubernetes.io/version` label is now the real `5.0.0` (was `latest`), and consumers can drop the manual tag pin.

## Type of change
- [x] New feature (first chart in the ADR 0030 appVersion→image coupling)

## Testing
`helm template` renders `ghcr.io/washu-tag/superset:5.0.0` and `app.kubernetes.io/version: "5.0.0"`; `helm lint` passes.

## Note for reviewers
Deploy render is a no-op (still `superset:5.0.0`); the only manifest change is the accurate version label. A Renovate manager to keep this `appVersion` in lockstep with `apache/superset` rides in `ci/renovate-track-superset` (grouped with the superset image managers).

## Checklist
- [x] Adheres to guidelines
- [x] Commented (why tag is empty)
- [ ] Docs / tests (N/A, chart plumbing)


johnflavin (approved):
I think this, #606, and #608 should have been a single PR

andrey-embarklabs:
> I think this, #606, and #608 should have been a single PR

Fair. I split them so the chart change, the CI hardening, and the Renovate
config could be reviewed and reverted independently

PR #629, opened 2026-08-10
fix(ci): publish images with a missing tag, and block releases that lack them

kathrynalpert:
### Product

No user-facing change. Release plumbing: a release can no longer be tagged without the images it names.

### Technical

Background: `v4.1.0` was tagged with no `4.1.0` images for five of the six Scout-versioned images. The `build-and-upload` workflow builds when an image's paths changed, when the `ci` filter fires (meaning CI logic itself changed), or if its derived tag is absent from the registry. However, `publish` gated only on the change condition. Because the `v4.1.0` release was fixed forward on a commit touching two images, the other five were built at `4.1.0` and scanned, but then dropped by the publish step. And `release.yaml`, which only waits for CI to be green, went ahead and tagged the commit, despite the publish gap.

Changes:
- **`publish`** now gates each push on `paths changed || tag absent`
- **`release.yaml`** asserts every Scout-versioned tag resolves before `gh release create`
- **The registry probe** in `check-image-exists` now asks about the one tag of interest, instead of paging through all package versions. (The old list-and-grep saw only the newest page — it reports `launchpad:4.0.0` absent today.) An ambiguous answer (500, 429, etc) now fails loudly instead of defaulting to "image missing", which under the new gate would overwrite a live tag.

Both callers share `.github/scripts/ghcr-tag-published.sh`, and `release.yaml`'s hardcoded image roster becomes `SCOUT_VERSIONED_IMAGES` / `UPSTREAM_VERSIONED_IMAGES` so the list exists once.

## Type of change

- [x] Bug fix

## Impact

CI only

## Testing

Exercised against live GHCR: `launchpad:4.0.0` → true (the case the old query gets wrong), `hl7-listener:4.1.0` → true, the five missing `4.1.0` tags → false, never-pushed repo → false, forced network failure → exit 1. The release guard at `VERSION=4.1.0` reproduces the current broken state and blocks.

## Note for reviewers

N/A

## Checklist

- [x] Style guidelines (`pre-commit` on changed files)
- [x] Commented, particularly the non-obvious gating
- [x] Technical documentation (CLAUDE.md image-wiring list)

### DIFF
--- a/.github/scripts/ghcr-tag-published.sh
+++ b/.github/scripts/ghcr-tag-published.sh
@@ -1,0 +1,65 @@
+#!/usr/bin/env bash
+#
+# Is ghcr.io/washu-tag/<image>:<tag> published?
+#
+# Usage:
+#   ghcr-tag-published.sh <image-name> <tag>   # prints "true" or "false"
+#
+# Prints the answer and exits 0 when the registry gave a definitive one; prints
+# an ::error:: and exits non-zero when it did not, so no caller has to guess
+# from an ambiguous result. Callers decide what a "false" means to them --
+# ci.yaml builds and publishes, release.yaml refuses to cut the release.
+#
+# Asks about the one tag rather than listing package versions: the packages API
+# has no tag filter, and images carry 100+ versions (untagged ones are never
+# pruned), so a list-and-grep only ever saw the newest page.
+set -euo pipefail
+
+IMAGE="${1:?usage: $0 <image-name> <tag>}"
+TAG="${2:?usage: $0 <image-name> <tag>}"
+
+repo="washu-tag/${IMAGE}"
+# Retry first, so a hard failure below means a real problem rather than a blip:
+# curl retries timeouts, connection failures, 408/429/5xx -- never 404.
+retry=(--retry 3 --retry-connrefused --retry-delay 2)
+
+token_body="$(mktemp)"
+trap 'rm -f "${token_body}"' EXIT
+
+# ghcr mints a pull token per repository. An image that has never been pushed
+# has no repository, so the token endpoint answers 403 DENIED instead of issuing
+# one -- that, not a 404 on the manifest, is what a brand-new image looks like.
+token_code="$(curl -sS "${retry[@]}" -o "${token_body}" -w '%{http_code}' \
+  "https://ghcr.io/token?service=ghcr.io&scope=repository:${repo}:pull")" || {
+  echo "::error::could not reach the ghcr token endpoint for ${repo}" >&2
+  exit 1
+}
+case "${token_code}" in
+  200) ;;
+  403)
+    echo false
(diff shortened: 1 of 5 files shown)


.github/scripts/ghcr-tag-published.sh
```diff
@@ -0,0 +1,65 @@
...
+# pruned), so a list-and-grep only ever saw the newest page.
+set -euo pipefail
+
+IMAGE="${1:?usage: $0 <image-name> <tag>}"
+TAG="${2:?usage: $0 <image-name> <tag>}"
+
+repo="washu-tag/${IMAGE}"
+# Retry first, so a hard failure below means a real problem rather than a blip:
+# curl retries timeouts, connection failures, 408/429/5xx -- never 404.
+retry=(--retry 3 --retry-connrefused --retry-delay 2)
```
johnflavin: Should probably set a timeout: `--max-time 10`

.github/scripts/ghcr-tag-published.sh
```diff
@@ -0,0 +1,65 @@
...
+
+token_body="$(mktemp)"
+trap 'rm -f "${token_body}"' EXIT
+
+# ghcr mints a pull token per repository. An image that has never been pushed
+# has no repository, so the token endpoint answers 403 DENIED instead of issuing
+# one -- that, not a 404 on the manifest, is what a brand-new image looks like.
+token_code="$(curl -sS "${retry[@]}" -o "${token_body}" -w '%{http_code}' \
+  "https://ghcr.io/token?service=ghcr.io&scope=repository:${repo}:pull")" || {
```
johnflavin: This request is unauthenticated, so a 403 could mean either "no image exists" or "no _public_ image exists". That is the default state for any new image we publish, so unless we explicitly flip them to public this would see them as unpublished every time. See [GHCR docs](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#pushing-container-images):
> When you first publish a package, the default visibility is private.

This is a regression, since the old workflow used the authenticated `gh api` instead of `curl`.

johnflavin (approved):

.github/scripts/ghcr-tag-published.sh
kathrynalpert: Good point!

.github/scripts/ghcr-tag-published.sh
kathrynalpert: It's not really a regression, since the old workflow didn't have package permissions, but it's probably cleaner to just authenticate. I did waffle on whether we'd actually want a failure here to notify us that a package isn't public, since presumably we'd always want that, but I decided if we wanted to test things were public, we should explicitly do so.

(As an aside, I'm not crazy about this curl approach, but I prefer it to paging through `gh api` results!)

PR #630, opened 2026-08-10
feat(extractor): chart-own the modality mapping

andrey-embarklabs:
### Technical
Phase 0 (ADR 0031) moves derived/site config into the charts so a change ships as a chart version bump rather than a hand-copied manifest. `spark-defaults` already moved; the modality-code map was the last config Ansible created out-of-band (a `kubernetes.core.k8s` ConfigMap built from a 22k-line CSV, mounted via injected values).

The `hl7-transformer` chart now owns it: the default CSV moves into `files/modality_mapping_codes.csv`, a chart ConfigMap template renders it (gated on `modalityMapping.enabled`, default true), and the deployment mounts it at `/config/modality_mapping_codes.csv` exactly as before. A custom map is now a chart value (`modalityMapping.data`) instead of an Ansible-created ConfigMap. Ansible drops the ConfigMap task and the injected volume/mount; `modality_map_source_file` defaults empty (chart-owned) and, when set, is validated and passed through as `modalityMapping.data`.

## Type of change
- [x] Improvement

### Backward compatibility
Behavior-preserving. `helm template` verified the rendered ConfigMap is byte-identical to the previous map (Ansible's `lookup('file')` also stripped the trailing newline). Operational note: on upgrade of an existing cluster the old out-of-band `modality-mapping` ConfigMap is left orphaned (nothing mounts it) and can be deleted; the chart now owns `<release>-hl7-transformer-modality-mapping`.

## Testing
`helm lint` clean; `helm template` verified byte-identical default map, correct mount/volume, working `modalityMapping.data` override, and clean `enabled=false` gating.


ansible/roles/extractor/defaults/main.yaml
```diff
@@ -48,7 +48,10 @@ hl7logs_root_dir: /data
 # Modality mapping file (source file on control machine)
```
kathrynalpert: Should this old comment be removed?

helm/extractor/hl7-transformer/values.yaml
```diff
@@ -110,6 +110,18 @@ tolerations: []
...
+# Chart-owned modality-code -> modality mapping (ADR 0031). The chart bundles the
+# default map at files/modality_mapping_codes.csv and mounts it at
+# /config/modality_mapping_codes.csv, where the transformer reads it (the
+# MODALITY_MAP_PATH constant). Enabled by default because the transformer needs
+# the map; a bare install works from the bundled default. To use a custom map,
+# set `data` to the full CSV (overrides the bundled default), or disable this and
+# mount your own. A map update ships as a chart version bump (no hand-copied
+# ConfigMap), which is the point of chart-ownership.
+modalityMapping:
+  enabled: true
```
kathrynalpert: Why is this default enabled while spark defaults is default disabled? I'm fine with either, but I think they should be consistent.

helm/extractor/hl7-transformer/values.yaml
```diff
@@ -110,6 +110,18 @@ tolerations: []
...
+  data: ''
```
johnflavin, line 121: I don't understand why there needs to be an `enabled` switch gating this. The only explanation is in the comment which says 
> ...or disable this and mount your own [csv].

But how would I do that? If I don't provide a path to mount the mapping file here when the chart is deployed, I don't know of a way I can retroactively mount one into the pod later. And why would I want to, if I could just mount the file here now?

I don't see a need for this `modalityMapping.enabled` value and all the checks for it around the rest of the chart. IMO we should remove it. If there is no need now but might be one later, we could easily add it then. But maybe I'm missing a use case you have at Embark that I can't see.

kathrynalpert (approved):
Couple other requests that didn't really fit inline:
1. Stale docs. docs/source/operate/ingest.md:27 and docs/source/operate/inventory.md:821 both name the deleted path and describe Ansible building the ConfigMap. 
2. Nothing triggers a rollout when the map changes, seems like a `checksum/modality-mapping` pod annotation is worth adding (spark-defaults has the same gap)
3. Orphan cleanup. The PR body tells operators to go delete the old modality-mapping ConfigMap by hand. A kubernetes.core.k8s: state: absent task would make the upgrade self-cleaning.

johnflavin:
> Nothing triggers a rollout when the map changes

Is this necessary? If the configmap changes I _think_ k8s will synchronize those changes into the pod. Nothing in the hl7-transformer caches that mapping file, it is read anew every time the activity runs. https://github.com/washu-tag/scout/blob/main/extractor/hl7-transformer/src/hl7scout/hl7extractor/deltalake.py#L287

helm/extractor/hl7-transformer/values.yaml
andrey-embarklabs: good call, agreed. dropped modalityMapping.enabled in dee51d08, so the ConfigMap + mount always render now and modalityMapping.data stays for the override case. the transformer always needs the map and you can't retroactively mount one anyway, so there was no real use for the toggle. easy to add back if a bring-your-own-mount case ever comes up.

andrey-embarklabs:
> > Nothing triggers a rollout when the map changes
> 
> Is this necessary? If the configmap changes I _think_ k8s will synchronize those changes into the pod. Nothing in the hl7-transformer caches that mapping file, it is read anew every time the activity runs. https://github.com/washu-tag/scout/blob/main/extractor/hl7-transformer/src/hl7scout/hl7extractor/deltalake.py#L287

you're right the app re-reads the map every activity run, but there's a wrinkle: it's a subPath mount, and subPath ConfigMap mounts don't get the live update k8s does for whole-volume mounts, so a running pod won't pick up a data change on its own. in the normal flow it still propagates because a map change is a chart-version bump, which rolls the deployment and the fresh pod reads the new map. the gap kathryn's pointing at is a data-only change on the same chart version. so i added checksum/modality-mapping (and the same for spark-defaults, same subPath gap) in dee51d08 to force a rollout when the content changes, cheap and robust either way.

helm/extractor/hl7-transformer/values.yaml
andrey-embarklabs: removed the toggle (see johnflavin's thread), so it's just always-on now rather than enabled-by-default. on consistency: the difference is intentional, the modality map is required app data the chart bundles so it's unconditional, whereas sparkDefaults is optional site tuning so it keeps its opt-in switch. happy to revisit sparkDefaults separately if we'd rather they match exactly.

ansible/roles/extractor/defaults/main.yaml
andrey-embarklabs: yep, leftover, removed it in dee51d08. the new comment covers it, and the old "Default points to the reference mapping shipped with Scout" line wasn't true anymore now that the default is empty.

andrey-embarklabs:
> Couple other requests that didn't really fit inline:
> 
> 1. Stale docs. docs/source/operate/ingest.md:27 and docs/source/operate/inventory.md:821 both name the deleted path and describe Ansible building the ConfigMap.
> 2. Nothing triggers a rollout when the map changes, seems like a `checksum/modality-mapping` pod annotation is worth adding (spark-defaults has the same gap)
> 3. Orphan cleanup. The PR body tells operators to go delete the old modality-mapping ConfigMap by hand. A kubernetes.core.k8s: state: absent task would make the upgrade self-cleaning.

thanks, all three make sense:

- stale docs, fixed ingest.md and inventory.md in dee51d08, both still named the old extractor/hl7-transformer/ path and described Ansible building the ConfigMap.
- rollout, added checksum/modality-mapping (+ spark-defaults) pod annotations in dee51d08 so a content change forces a rollout, details in johnflavin's thread re: the subPath nuance.
- orphan cleanup, i'd rather not bake a state: absent task into Ansible while it's being retired. the old modality-mapping ConfigMap is harmless once the chart owns it (nothing mounts it), so there's nothing to do on a fresh deploy, and on an existing cluster it can just be dropped whenever or swept up when the Ansible roles retire.

PR #631, opened 2026-08-10
feat(ci): haul the two appVersion-derived wrapper-chart upstream images

andrey-embarklabs:
### Technical
Follow-up to #619, which folded the wrapper charts' third-party images into the build-lane haul but deferred two: `temporalio/admin-tools` (temporal-bootstrap) and `ghcr.io/open-webui/open-webui` (open-webui-bootstrap). Those two take their image tag from an upstream Helm chart's **appVersion**, not a tag var in `versions.yaml`, so `resolve_upstream.py` couldn't source them.

Resolves them the way Scout does at deploy time: `versions.yaml` pins the *chart* version (a range); a build-time `helm show chart --repo <url> --version <range>` reads that chart's appVersion, which is the image tag. New `resolve_upstream_helm.py` (reuses `resolve_upstream.load_versions`) reads a `<image-repo> <helm-repo-url> <chart> <chart-version-var>` map (`upstream-images-helm.txt`); the `publish-haul` job runs the `helm show` lookup and appends the resolved refs to the same `@digest` pin + no-key upstream sync. Fail-closed: resolver crash aborts via `set -e`, unresolvable appVersion → `exit 1`; job stays `continue-on-error`.

## Type of change
- [x] New feature

### Backward compatibility
None. Images pinned by immutable digest exactly as the other upstream images; no new signing surface.

## Testing
New `test_resolve_upstream_helm.py` (4 offline tests); full `tooling/manifest` suite green (35 passed). End-to-end against real `versions.yaml`: resolves `temporalio/admin-tools:1.31.0` and `ghcr.io/open-webui/open-webui:0.11.0`, both inspect to a digest and render through `render_images`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1246,6 +1246,22 @@ jobs:
           python3 tooling/manifest/resolve_upstream.py \
             tooling/manifest/upstream-images.txt \
             ansible/group_vars/all/versions.yaml > upstream-tags.txt
+          # Two wrapper-chart images take their tag from an upstream chart's
+          # appVersion (helm lookup, not a versions.yaml tag var -- see
+          # upstream-images-helm.txt). Resolve the chart version here, then the
+          # appVersion with `helm show`, and append to the same tag list so they
+          # flow through the digest + no-key sync below, exactly as Scout resolves
+          # them at deploy time. helm is preinstalled on the runner.
+          python3 tooling/manifest/resolve_upstream_helm.py \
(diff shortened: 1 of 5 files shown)


kathrynalpert:
We've discussed swapping off the wildcard versions, aka `~` in versions.yaml, for better provenance. Would that simplify this effort?

andrey-embarklabs:
superseded by #637. went with the pin-the-image-tags approach you suggested instead of the helm-show lookup. closing.

PR #632, opened 2026-08-10
ci: require upgrade notes for breaking PRs

andrey-embarklabs:
### Technical
Phase 1 (ADR 0030 release automation) calls for an upgrade-notes check alongside the Conventional-Commit title lint. Adds an `upgrade-notes` job to the PR-title-lint workflow: when a PR is marked breaking (a `!` marker in the title, or a `BREAKING CHANGE:` footer in the body), the **Backward compatibility** section of the PR description must be filled in, otherwise the check fails. Reads the title/body from the event payload, so no checkout and no repo write.

Gates section *presence*, not correctness: a behavioral break that adds no variable stays a review concern, per ADR 0030. The PR template's Backward-compatibility comment now notes the requirement.

## Type of change
- [x] Improvement

### Backward compatibility
None. Adds a new PR check; existing non-breaking PRs are unaffected (the gate is a no-op unless a breaking marker is present).

## Testing
Extracted the job's actual `run` script and executed it against 8 title/body cases (non-breaking, `feat!:`, `feat(api)!:`, `BREAKING CHANGE:` footer, filled vs empty section, trailing-footer edge, missing section) using the real PR template body: all pass.


.github/workflows/pr-title-lint.yaml
```diff
@@ -22,3 +22,40 @@ jobs:
...
+          breaking = (
+              re.match(r"^[a-z]+(\([^)]*\))?!:", title, re.I) is not None
+              or re.search(r"^BREAKING[ -]CHANGE:", body, re.M) is not None
+          )
+          if not breaking:
+              print("No breaking-change marker; upgrade notes not required.")
+              sys.exit(0)
+          # Grab the "Backward compatibility" section: everything until the next
+          # heading or a BREAKING CHANGE footer (so a trailing footer isn't absorbed).
+          m = re.search(r"^#{2,4}\s*Backward compatibility\s*$(.*?)(?=^#{1,4}\s|^BREAKING[ -]CHANGE:|\Z)", body, re.M | re.S)
```
kathrynalpert: I think there's a bug or two here. From Claude:

> **`^BREAKING[ -]CHANGE:` → false fail.** The comment on line 52 explains the intent ("so a trailing footer isn't absorbed"), but it also terminates the section when the footer is the *first* thing under the heading:
> 
> ```markdown
> ### Backward compatibility
> BREAKING CHANGE: operators must set foo_bar in inventory.
> ```
> 
> Section reads empty → check fails.
> 
> **^#{1,4}\s → false pass.** The template already uses ##### Authorization / ##### Appsec, and a 5-hash heading doesn't match this terminator, so the section swallows the next heading plus its prose, goes non-empty, and an untouched placeholder passes. Since we only want to know whether the author wrote anything under the heading, ^#+\s drops the bound entirely and fails closed on any hash run.
> 
> ```markdown
> ### Backward compatibility
> 
> 
> ##### Some subsection
> unrelated prose
> ```
> 
> One line fixes both:
> 
> ```suggestion
> m = re.search(r"^#+\s*Backward compatibility\s*$(.*?)(?=^#+\s|\Z)", body, re.M | re.S)
> ```
> 
> Note: this deliberately flips one case from your test matrix: a placeholder section with a trailing `BREAKING CHANGE:` footer and no later heading now passes instead of failing (a footer *is* the operator upgrade path, so it shouldn't matter whether it sits above or below the heading).

Overall, though, I'm not sure PR body is really where we need the breaking change info (it doesn't hurt to have it, but I'm not sure it warrants a CI check). Release please is going to read from the commit message, so that's where it really matters.

andrey-embarklabs: good catches on both, applied your #+ one-liner in 85e8ae07 and re-ran the matrix, all green including the footer-under-heading and 5-hash cases. and agreed on the flip you flagged, a footer's a footer wherever it sits.

on the bigger question, i don't want to oversell this one. the case for keeping it: releases are the operator upgrade contract, and the failure mode it guards is a breaking change merging with zero operator-facing "what do i change" written down anywhere. review usually catches that, but a cheap gate means it can't slip through on a busy day, and it's the check ADR 0030 calls for alongside the title lint.

you're also right that release-please reads the commit message, so the real home for the note might be the BREAKING CHANGE: footer that lands in the changelog rather than a template section, and with squash merges the two are close but not identical. rather than settle that now, i'd park this PR until release-please actually goes live (that cutover is deferred to the phase-5 Ansible retirement anyway), then revisit whether to gate the footer or the section with the real release flow in front of us. it's not blocking anything.

sound ok? happy to just close it and reopen when release-please lands if you'd rather not carry an open PR.

kathrynalpert: I think close + reopen might be better so it doesn't add noise for reviewers visiting the PRs page.

andrey-embarklabs:
closing for now

PR #636, opened 2026-08-11
feat(launchpad): Customizable launchpad

### LINKED ISSUE #480: Launchpad should support dev "chips"
This is something @johnflavin has mentioned, and it seems pretty useful to add. Developers should be able to add dev application chips to the launchpad without building a custom launchpad image in some way. Maybe they're stored in a configmap?

johnflavin:
### Product

- End user: no change. Launchpad still looks exactly like it did before
- Site admin / customizer: Added the ability to deploy new components into launchpad, or change existing components, by defining and editing ConfigMaps. A user-facing docs page docs/source/customize/launchpad-chips.md explains the process.

### Technical

See detailed description in ADR 0034.

Launchpad is rebuilt to render its UI based on configs that can be provided and changed at runtime. The process goes as such:
1. A new launchpad sidecar container watches for ConfigMaps bearing the label `launchpad.scout.xnat.org/catalog: "true"`
2. When the sidecar sees a change (new, updated, or deleted) in a labeled ConfigMap, it dumps the contents into an `emptyDir` mount shared with the main launchpad container.
3. Separately, the core launchpad elements are mounted directly into the main container. That way if there is some weirdness or outage with the sidecar or its mount, we still get the core stuff.
4. On page load, the main launchpad container either loads the contents of those files or serves a cached snapshot; this snapshot is a validated normalized in-memory representation of everything in the files: not what is on disk but also not 100% what gets rendered on the page. If the last snapshot was built <10s ago, we serve it. If the snapshot is older, we stat the files in the dir and generate a signature; if they haven't changed, we serve the cached snapshot and reset its clock. We only read files if things have changed since last check. When any files are changed, all files are read and a new snapshot is built and cached. Lastly there are some fallbacks where if the rebuild fails or if there are suddenly no configs in the dirs, we serve the last snapshot.
5. On page load, after the snapshot is rebuilt or grabbed from the cache, the snapshot is fully rendered out to the page elements.

This also moves the launchpad chips of several scout services—chat, minio, playbooks, maybe a couple others—into configmaps that get deployed by those playbooks, not built into launchpad.

Closes #480 

## Type of change
- [X] New feature
- [ ] Improvement
- [ ] Bug fix
- [ ] Documentation update
- [ ] Test update

##### Authorization

- Admin chips (`audience: admin`) are filtered server-side, so a non-admin's HTML never contains admin chips.
- The `audience` field fails closed. If the value isn't one of `user` or `admin` the chip isn't shown to anyone. That way a chip intended for admins with the incorrect value `audience: Admin` wouldn't fail into being shown to everyone. Still, that wouldn't be so bad because...
- Chip visibility is not authorization. This doesn't change anything about the how individual services authenticate. All the admin-only services still reject non-admin users themselves. So even if an admin-only chip shows up (which it shouldn't) that doesn't suddenly expose an admin-only service, just a chip that is non-functional for most users.
- Launchpad chart's ServiceAccount gets a ClusterRole granting cluster-wide ConfigMap `get`/`list`/`watch` for the sidecar.

##### Appsec

Catalog content originates in other namespaces and is treated as untrusted: 
- destinations admit only `http(s)` URLs, rooted paths, and DNS-label subdomains (so `javascript:`/`data:`
links reject the chip), 
- `iconData` is limited to an allowlisted image data-URI prefix and size cap and renders only via ``, and 
- all strings render as text nodes — nothing is interpolated as HTML. 
 
Accepted residual risk, documented in ADR 0034: any workload that can create a labelled ConfigMap in any namespace can place a chip on the landing page.

### Performance

The landing page is one small server render per request. The catalog is parsed once and
held as an in-process snapshot behind a 10s TTL keyed by a cheap directory signature, with
single-flight rebuilds, so steady-state requests do no filesystem work. The sidecar is
pinned at 10m CPU / 32Mi requests with a 10Mi emptyDir. Scale is tens of chips, not
thousands.

### Data

The data concerns here are about interpreting user-supplied ConfigMaps for rendering launchpad chips. We don't want typos and misconfigurations to bring down the whole page. So we take a layered approach to failure:
- A bad field costs the field. If a single non-critical field is missing or wrong, we can fall back to a default or leave it off and keep rendering the chip.
- A bad chip costs the chip. A required field being missing or wrong means the chip won't render and just gets skipped.
- A bad document from a ConfigMap costs only its own chips.
- A React error boundary per group costs at most a section. 

Unreadable catalog directories, non-YAML keys, and files that vanish mid-write surface as diagnostics; if every directory becomes unreadable the previous snapshot is served rather than blanking the page.

Any error in reading or rendering is logged and also displayed on the launchpad page (to admins only) as a visible error message. That way if someone is iterating on their config, they can see errors right away rather than needing to dig through logs.

### Backward compatibility

Breaking for the launchpad's own configuration surface: `ENABLE_CHAT`, `ENABLE_PLAYBOOKS`,
and `ENABLE_MINIO` are retired from the deployment, chart, and role, replaced by
`LAUNCHPAD_CATALOG_DIRS` plus catalog ConfigMaps published by each owning component. The
inventory `enable_*` flags are unchanged and still gate the components themselves. Known
accepted regression (documented in `scout_common/tasks/launchpad_catalog.yaml`): the
Ansible publisher only creates, so flipping a feature flag back to false leaves the chip
behind until the GitOps migration (ADR 0031) moves publication into charts. New
dependencies: `zod` and `yaml` in the launchpad image, and the `kiwigrid/k8s-sidecar`
image pinned in `versions.yaml` with a Renovate annotation.

## Testing

Launchpad gets a new ~~vitest~~ unit-test suite for the catalog module — schema validation and degradation, multi-document parsing, loader snapshot/signature/fallback behavior, and a drift guard asserting the docs' icon and tone galleries match the launchpad's own registries. Wired into CI as `Unit tests: launchpad`. 

**UPDATE:** vitest was pulling in transitive dependencies that were old/deprecated and tripped the OpenSSF Scorecard check. Rewrote the tests to use the more basic built-in node testing.

This was all deployed to dev03 and verified to work as expected. 
- With only a launchpad redeploy, only the core services appeared on the page
- Had to run other playbooks that now publish service configmaps to get their chips to appear
- Manually created and deleted configmaps containing groups and chips, enabled and disabled, and verified everything renders as advertised.

## Note for reviewers
The user-facing doc docs/source/customize/launchpad-chips.md is probably the most important entrypoint into this functionality, so I would appreciate a careful review there.

## Checklist


.github/workflows/ci.yaml
andylassiter, line 1: Nit: drop both comments.

ansible/roles/scout_common/tasks/launchpad_catalog.yaml
```diff
@@ -0,0 +1,45 @@
+---
```
andylassiter, line 1: Is there a more GitOpys way to do this instead of adding more Ansible deployment steps?

docs/internal/adr/0034-runtime-configurable-launchpad-catalog.md
```diff
@@ -0,0 +1,279 @@
...
+- Putting a service on the landing page = one labelled ConfigMap in the service's own
+  chart or role; no launchpad rebuild, no launchpad edit. Removal is deletion.
+- The launchpad stops knowing what is on the page: presence bugs become config bugs
+  with per-chip diagnostics instead of image rebuilds.
+- The three `ENABLE_*` env vars and their plumbing disappear; the launchpad role stops
+  passing content flags. The page reflects what is installed.
+- New moving parts: one sidecar container (pinned + Renovate-watched, ADR 0015), zod
+  and yaml as launchpad dependencies, and the repo's first explicitly-authored
+  ClusterRole (ConfigMap get/list/watch, cluster-wide, attached to the launchpad pod's
+  ServiceAccount — shared by both containers, since ServiceAccounts are pod-scoped).
```
andylassiter: Can we limit the SA to the sidecar only? Claude flagged this and suggested:

> Consequences, and worth considering `automountServiceAccountToken: false` on the pod with an explicitly projected token volume mounted only into `catalog-sidecar`

docs/source/customize/launchpad-chips.md
```diff
@@ -0,0 +1,231 @@
+# Put Your Service on the Scout Launchpad
```
andylassiter: Suggestion: "Launchpad Chips" or "Adding a Launchpad Chip"

docs/source/customize/index.md
```diff
@@ -0,0 +1,10 @@
+# Customize Scout
```
andylassiter: Suggestion: "Extending Scout". Customize reads more like theming.

docs/source/customize/index.md
```diff
@@ -0,0 +1,10 @@
...
+
+Build and deploy custom Scout components.
+
+- **[Launchpad Chips](launchpad-chips.md)** — Add your own UI components ("chips") to the Scout Launchpad.
+
+```{toctree}
+:maxdepth: 1
+launchpad-chips
+```
```
andylassiter, line 10: There is a manual bullet list followed by an automated document tree. This feels repetative and looks odd.

Image

docs/source/index.md
```diff
@@ -16,6 +16,7 @@ Analytics, Chat, and Notebooks.
 
 * **[Using Scout](user/index.md)** — explore and analyze data: logging in, Analytics, Chat, Notebooks, and understanding what data you can see.
 * **[Operating Scout](operate/index.md)** — deploy and run Scout: the Ansible inventory, air-gapped installation, report ingestion, and per-user data-authorization configuration.
+* **[Customize Scout](customize/index.md)** — customize Scout: deploy your own playbooks and services into the Scout platform
```
andylassiter: This repeats itself "Customize Scout — customize Scout:"

docs/source/customize/launchpad-chips.md
```diff
@@ -0,0 +1,231 @@
...
+it discovers at runtime. To put a service on the page, publish a ConfigMap — in your
+service's own namespace — labelled `launchpad.scout.xnat.org/catalog: "true"`. The
+launchpad imports it within seconds. No launchpad configuration, rebuild, or restart is
+involved, and deleting the ConfigMap removes the chip.
+
+This works the same for Scout's own components, for site-installed services, and for
+anything else running in the cluster: presence on the landing page follows
+installation.
+
+## The 30-second version
```
andylassiter: Suggestion: "A minimal chip". Easier to read in a TOC and makes no promises of the reader's time.

docs/source/customize/launchpad-chips.md
```diff
@@ -0,0 +1,231 @@
...
+
+`kubectl apply` that and a "My Service" card appears on the launchpad, in the **More**
+section — `more` is the default group, defined by the launchpad's own catalog. See
+[groups](#defining-a-group-section) to put the chip in a section of your own.
+
+Ship it however you deploy: a template in your Helm chart (so uninstalling the chart
+removes the chip), a Flux-reconciled manifest, or — for Scout-internal roles — the
+`scout_common` `launchpad_catalog` task.
+
+## The document
```
andylassiter: This is vague. Suggestion: "Catalog documents"

docs/source/customize/launchpad-chips.md
```diff
@@ -0,0 +1,231 @@
...
+
+You state a subdomain, never a hostname — the launchpad resolves it against the host
+the browser is already on. Path suffixes matter for services with their own login
+bootstrap (Temporal's `/auth/sso` is the canonical example: the bare subdomain lands on
+an unauthenticated UI). Only `http(s)` URLs, rooted paths, and DNS-label subdomains
+validate; anything else (e.g. `javascript:`) rejects the whole chip. A rooted path
+means exactly one leading slash — protocol-relative `//host` values, which the browser
+would resolve off-site, are rejected too.
+
+## Defining a group (section)
```
andylassiter: Suggestion: "Groups". Aligns better with the other section titles.

ansible/roles/scout_common/tasks/launchpad_catalog.yaml
johnflavin: Yes, there is, we would put the configmap into the chart of whatever service is making the chip. And IMO it's strictly better than this, since right now uninstalling a service would leave this configmap behind but if it's all helm managed removing the service = removing the configmap = removing the chip. But we aren't there yet. This is just another thing that we accept as a limitation for now which will need to be migrated as part of that process.

docs/source/customize/index.md
johnflavin: I would like the pluggability to include things like theming or similar cosmetic adjustments in the future, so I don't want to drop "Customize". But I'll add the word "Extend".

docs/source/customize/index.md
johnflavin: Yep, sorry, forgot to add `:hidden:` to the toctree.

docs/internal/adr/0034-runtime-configurable-launchpad-catalog.md
johnflavin: I did do this. It involved a bit of hackery and manually mounting the SA token into the sidecar, so I'm not sure it was worth it. But now that it's there I think it works fine and is unlikely to change, so I'll leave it.

johnflavin:
On the potential problem
> Cross-document duplicate chip ids are neither deduped nor reported. (`schema.ts`, `seenChipIds` is per-document; `assemble.ts:56-72` applies precedence to groups only.)

This is something I noticed during dev and I think I took care of it. The per-document IDs can collide with no issue because in `CatalogSections.tsx:134` the key is `${chip.source}:${chip.id}`, not `chip.id`. So we do keep track of where they came from when rendering.

PR #639, opened 2026-08-12
feat(deploy): Phase 3 GitOps deployment base — Kustomize bases + Flux DAG (scaffold)

andrey-embarklabs:
### Product
None. Delivery migration (ADR 0031, Phase 3).

### Technical
Introduces `deploy/`: the GitOps deployment base that stands Scout up by **pulling** signed, digest-pinned artifacts through Flux, replacing the Ansible push. This is the **scaffold** (Kustomize bases + the Flux dependency graph) and is **unconsumed** — nothing deploys from it yet, no existing deploy/CI gate is touched, and the Ansible path stays the deploy path until the cutover.

Covers **22 components** as Flux bases, each mirroring its Ansible role (built + adversarially verified against the roles):
- **Ingest slice:** postgres, minio, cassandra, elasticsearch, hive, temporal, extractor
- **Auth/analytics:** valkey, keycloak (+ realm via keycloak-config-cli), oauth2-proxy, opa, trino (ro + rw), superset (+ scout-dashboards import)

Structure:
- `deploy/base/<component>/{operator,cluster,...}/` — Kustomize bases. CRD-owning operators split into an `operator/` (install) base and the CR base, so a CR never dry-runs before its CRDs exist.
- `deploy/flux/<component>.yaml` — **22 Flux `Kustomization`s** wiring the `dependsOn` DAG that reproduces the Ansible deploy order (acyclic, no dangling refs), with CEL `healthCheckExprs` on the CRs (CNPG Cluster, CassandraDatacenter, Elasticsearch, Keycloak).
- `deploy/required-vars.txt` — the site `cluster-vars` contract (48 keys).
- `.github/workflows/validate-deploy.yaml` — renders every base + checks the DAG on `deploy/**` PRs.

Conventions (see `deploy/README.md`): site scalars are `${var}` postBuild substitutions from a `cluster-vars` ConfigMap (`StrictPostBuildSubstitutions`); Scout chart/image refs are stamped from the build-lane haul at publish (`0.0.0` placeholders in git); secrets referenced by fixed name only (seeded by CI/site, never committed).

### Deferred (gated on the build lane going live)
- The **config-artifact publish job** that stamps the Scout charts' `0.0.0` placeholders with real published versions.
- The **`deploy-and-test` switch** (deploy via Flux; ingest suite = gate).
- Consolidating the co-declared `scout-charts` `HelmRepository` into a per-namespace `sources` base.
- Remaining bases: jupyter, report-viewer, monitoring, launchpad, feature Components.

## Type of change
- [x] New feature (deployment base scaffold; non-breaking, unconsumed)

### Backward compatibility
None. Adds `deploy/` + one PR-only validation workflow; changes nothing that runs today. The Ansible deploy path and all existing gates are untouched.

## Testing
`kubectl kustomize` renders all 22 bases cleanly; the Flux `dependsOn` graph verified acyclic with no dangling refs; the new `validate-deploy` workflow automates both on `deploy/**` PRs.

## Note for reviewers
Static scaffold — establishes the pattern and the DAG, deploys nothing yet, so review is about shape (bases faithful to the Ansible roles, DAG order, conventions) rather than runtime. The whole of Phase 3 accumulates on this one branch. Anchored to `docs/internal/phase3-implementation-plan.md`.


deploy/base/hive/db/resources.yaml
```diff
@@ -0,0 +1,120 @@
...
+                secretKeyRef:
+                  name: hive-metastore-readonly-secret
+                  key: HIVE_METASTORE_PASSWORD
+          command:
+            - sh
+            - -ec
+            - |
+              # 1) hive owner role + hive database (connect to the maintenance DB;
+              #    CREATE DATABASE cannot run inside a transaction, hence \gexec).
+              psql --no-psqlrc -v ON_ERROR_STOP=1 -v hive_pw="$HIVE_PW" \
```
kathrynalpert: Two separate problems in this heredoc, and both repeat in `keycloak-db-init` and `superset-db-init`.

First, `psql` doesn't interpolate `:'hive_pw'` inside a dollar-quoted block, so the literal text reaches the server. Reproduced against `postgres:16` with this exact script:

```
ERROR:  syntax error at or near ":"
LINE 4: CREATE USER hive WITH PASSWORD :'hive_pw';
```

Under `sh -ec` with `ON_ERROR_STOP=1` the container exits 3 and `backoffLimit: 3` exhausts, so neither the role nor the database gets created — `ON_ERROR_STOP` aborts before the `\gexec` too. Both metastore releases then crash-loop on JDBC auth and neither Trino has a catalog. Same thing at `keycloak/instance/resources.yaml:113` and `superset/server/resources.yaml:179`. Moving the `CREATE USER` out of the `DO` block into plain SQL is enough, since interpolation works there.

Second, `$HIVE_PW`, `$HIVE_RO_PW` and both `$do$` tags are envsubst targets. `kustomize-controller` runs `drone/envsubst`, which takes unbraced `$NAME` as well as `${NAME}` — hence Flux's `$$` escape. None of `HIVE_PW`, `HIVE_RO_PW` or `do` are in `required-vars.txt`, so with `StrictPostBuildSubstitutions` this Kustomization won't build at all; without it you get `-v hive_pw=""` and a bare `DO` with no quote body. Either way hive-db never provisions and hive-metastore, trino, trino-rw and extractor all sit behind it.

`trino/ro/resources.yaml` handles this correctly with `$${ENV:KEYSTORE_PASSWORD}`, but its comment describes the hazard as applying to a "single-`$` *braced* token", which undersells it. The `required-vars` check has the same assumption in its regex, which is why CI is green on all of this.

.github/workflows/validate-deploy.yaml
```diff
@@ -0,0 +1,98 @@
...
+
+      - name: required-vars covers every substitution
+        run: |
+          set -euo pipefail
+          # kustomize build leaves ${...} literal (no postBuild subst), so an
+          # incomplete required-vars.txt renders green but fails
+          # StrictPostBuildSubstitutions at deploy. Assert the contract lists every
+          # var the bases actually use. Regenerate required-vars.txt with:
+          #   grep -rhoE '\$\{[a-z0-9_]+\}' deploy/base deploy/flux | tr -d '${}' | sort -u
+          used=$(grep -rhoE '\$\{[a-z0-9_]+\}' deploy/base deploy/flux | tr -d '${}' | sort -u)
```
kathrynalpert: This step is pointed at the right failure — the comment above it says so — but the pattern has two gaps and the bugs it's meant to catch fall into both.

```
grep -rhoE '\$\{[a-z0-9_]+\}' deploy/base deploy/flux
```

`envsubst` substitutes unbraced `$NAME` as well as `${NAME}`, so every bare `$HIVE_PW`, `$HIVE_RO_PW`, `$MINIO_ROOT_USER`, `$MINIO_ROOT_PASSWORD` and `$do$` in the bootstrap Jobs is invisible here. And `[a-z0-9_]` means even `${HIVE_PW}` wouldn't match — all the shell vars are uppercase, so bracing them wouldn't help either.

So the step prints "required-vars.txt covers all N substitution vars" and passes, while four Jobs are staged to fail at deploy. Making the braces optional and widening the character class (skipping `$$`-escaped tokens) would cover it. Running the real `envsubst` over the built output and diffing would be better still, since that's what `kustomize-controller` actually does.

deploy/base/minio/tenant/resources.yaml
```diff
@@ -0,0 +1,400 @@
...
+            - |
+              set -euo pipefail
+              # Root creds + region/OIDC config, seeded by the site.
+              . /rootcreds/config.env
+
+              # GitOps has no external readiness gate — the Ansible role waited
+              # on the pool StatefulSet before creating this Job. Poll the tenant
+              # endpoint until it answers, then configure IAM.
+              until mc alias set local http://${minio_tenant_name}-hl.${minio_tenant_namespace}:9000 \
+                "$MINIO_ROOT_USER" "$MINIO_ROOT_PASSWORD" --insecure >/dev/null 2>&1 && \
```
kathrynalpert: Same envsubst problem as the db-init Jobs. These come from the mounted `config.env` at runtime, but as unbraced tokens they're substitution targets at build time, and neither is in `required-vars.txt`.

With strict substitutions the `minio-tenant` Kustomization won't build — no Tenant, no buckets, no IAM users. Without it, `mc alias set` runs with empty credentials, `mc ready local` never succeeds, and the `until` loop spins until the 10-minute timeout. hive-metastore, extractor, opa and trino all `dependsOn: minio-tenant`, so the lake and the ingest slice never come up.

`$${MINIO_ROOT_USER}` / `$${MINIO_ROOT_PASSWORD}` fixes it.

deploy/base/oauth2-proxy/resources.yaml
```diff
@@ -0,0 +1,159 @@
...
+  url: https://oauth2-proxy.github.io/manifests
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: oauth2-proxy-templates
+  namespace: "${oauth2_proxy_namespace}"
+data:
+  error.html: |
+    <<error.html: byte-exact copy of ansible/roles/oauth2-proxy/files/error.html, each line indented 4 spaces, with the single occurrence of 'https://washu-scout.readthedocs.io/en/latest' rewritten to '${scout_docs_url}'. Full 353-line content is on disk in the written file.>>
```
kathrynalpert: This ConfigMap has the authoring placeholder text in it rather than the HTML — both keys read as `<<error.html: byte-exact copy of ansible/roles/oauth2-proxy/files/error.html … Full 353-line content is on disk in the written file.>>`. The file is 159 lines total, so neither template ever got inlined.

It fails quietly. `${scout_docs_url}` inside the placeholder is a valid substitution, so the build succeeds and the ConfigMap applies fine. But `custom_templates_dir` on line 87 points here, so oauth2-proxy parses these as its sign-in and error templates, and everyone redirected by the `oauth2-proxy-error` middleware — Superset, Jupyter, MinIO console, launchpad — gets that sentence as their login page with no way to start the OIDC flow.

Given the size, these probably want to be chart files or a `configMapGenerator` with `files:` instead of inline literals, so they can't drift from `ansible/roles/oauth2-proxy/files/` either.

deploy/base/extractor/resources.yaml
```diff
@@ -0,0 +1,259 @@
...
+          metrics:
+            tags:
+              namespace: "${scout_extractor_namespace}"
+        s3:
+          endpoint: "${s3_endpoint}"
+          region: "${s3_region}"
+          path-style-access: true
+        spring:
+          datasource:
+            url: "jdbc:postgresql://${postgres_cluster_name}-rw.${postgres_cluster_namespace}:5432/${ingest_postgres_table_name}"
```
kathrynalpert: Nothing in `deploy/` creates this database or its owner role. `ansible/roles/extractor/tasks/deploy.yaml:8-11` calls `create_db_with_owner` for `ingest` / `{{ postgres_user }}` before either chart deploys, and hive, Keycloak and Superset each got a db-init Job here — but grepping for `ingest_postgres_table_name` only turns up this JDBC URL and `required-vars.txt`. `postgres/cluster/resources.yaml` is the bare CNPG Cluster with no `managed.roles` and no `bootstrap.initdb`.

The header comment says the ingest DB is "created against the shared CNPG cluster (postgres-cluster), not here", which reads as deliberate, but there's no manifest that does it.

So the extractor pod fails startup on `FATAL: database "ingest" does not exist` and crash-loops, and no worker polls `ingest-hl7-log` — the slice this is scaffolding can't run a workflow. Since it's CNPG anyway, `spec.managed.roles` plus `bootstrap.initdb` on the Cluster might be a better home for all four databases than a fourth psql Job.

deploy/base/keycloak/instance/resources.yaml
```diff
@@ -0,0 +1,268 @@
...
+# the bootstrap admin creds from keycloak-admin-secret (existingSecret ->
+# KEYCLOAK_PASSWORD; existingSecretKey defaults to `password`). The chart runs as a
+# Helm post-install/upgrade/rollback hook Job (chart-default annotations), so it
+# re-runs on every HelmRelease upgrade. config-hash preserves the role's
+# realm-change trigger: the realm content lives in the Secret (not in these values),
+# so a realm-only change would not otherwise re-run the import — CI stamps the realm
+# Secret's sha256 here at config-artifact publish (mirroring the role's truncated
+# sha256), which changes the Job spec and fires the hook.
+apiVersion: helm.toolkit.fluxcd.io/v2
+kind: HelmRelease
```
kathrynalpert: No readiness gate on Keycloak before the realm import. Ansible ran `deploy.yaml` then `configure.yaml`, and `deploy.yaml:270-280` waits on the Keycloak CR's `Ready` condition first. Here the CR and the keycloak-config-cli release are in one Kustomization and apply together.

The post-install hook then starts hitting `http://keycloak-service.<ns>:8080` while Keycloak is still on first boot, and DB migration takes minutes. The hook fails, the release fails inside its 5m timeout, and `keycloak-instance` never goes Ready — so the realm never imports and oauth2-proxy, trino-ro and superset-server (all `dependsOn` it with `wait: true`) never deploy. Bootstrap stalls with no login and nothing user-facing.

Splitting the import into its own Kustomization that `dependsOn` keycloak-instance restores the gate; `healthCheckExprs` on the CR would make it explicit.

kathrynalpert (commented):
Went through the bases and the Flux DAG against the Ansible roles they came from. I read the deferred list at the bottom of `deploy/README.md` first, so nothing below repeats the shared-`Namespace`/prune item or the chart version stamping — and the per-namespace `foundation/` base described there looks like the right shape for the first one.

Four things will stop a cold stand-up. The db-init `psql` heredoc has two separate bugs in it, the MinIO bootstrap has one of the same two, the oauth2-proxy templates ConfigMap still has the authoring placeholder text where the HTML should be, and nothing in `deploy/` creates the extractor's `ingest` database. None of them can show up yet, since `deploy-and-test` doesn't reconcile through Flux — so they'd all land together at the cutover.

Most of the rest is site parameterization that got flattened into literals during the port. Those render cleanly, so `StrictPostBuildSubstitutions` won't catch any of them.

On the `required-vars` step: it's aimed at exactly the right failure, but `\$\{[a-z0-9_]+\}` matches only the braced form and only lowercase, and every shell variable in the bootstrap Jobs is unbraced and uppercase. So they all sail through and the step reports success. Widening that pattern, or just running `envsubst` over the built output, catches the first three problems above and anything like them later. The other check I'd want is a normalized diff of each inlined copy against its Ansible original — right now nothing stands between an edit to `policy/trino/main.rego` and GitOps sites quietly enforcing the old authorization policy.

Last group is duplication. I'm raising it because it's load-bearing here rather than cosmetic: the three copy-pasted db-init Jobs are why the heredoc bugs land three times over, and the duplicated metastore release is why the hardcoded access key shows up twice.

deploy/base/superset/server/resources.yaml
```diff
@@ -0,0 +1,732 @@
...
+                "name": "keycloak",
+                "icon": "fa-key",
+                "token_key": "access_token",
+                "remote_app": {
+                    "client_id": "superset",
+                    # Confidential-client secret sourced from the env (superset-env),
+                    # never inlined into config committed to git.
+                    "client_secret": os.environ.get("KEYCLOAK_SUPERSET_CLIENT_SECRET"),
+                    "client_kwargs": {"scope": "openid microprofile-jwt"},
+                    "api_base_url": "https://keycloak.${server_hostname}/realms/scout/protocol/",
```
kathrynalpert: Grouping five of these together since they're the same class and the same fix. All of them render cleanly, so `StrictPostBuildSubstitutions` won't flag any — nothing's undefined, the values are just wrong anywhere the defaults don't hold.

Here on 338-369, the Keycloak host and realm, the client id, and the sign-out host are literals. The sibling bases already have vars for all of them (`${keycloak_realm_url}` in oauth2-proxy, `${keycloak_realm_name}` in trino/ro, `${oauth2_proxy_subdomain}` in oauth2-proxy), and `values.yaml.j2` parameterized all four. Override `keycloak_realm_name` or `oauth2_proxy_subdomain` and sign-in gets a Keycloak 404 while logout redirects to a vhost that doesn't exist.

`superset/server/resources.yaml:239` and `minio/tenant/resources.yaml:116` hardcode the `kube-system-` middleware prefix, but oauth2-proxy creates those Middlewares in `${oauth2_proxy_namespace}` and its header tells consumers to reference `${oauth2_proxy_namespace}-oauth2-proxy-{auth,error}@kubernetescrd`. Different namespace means Traefik can't build the router, and both the Superset host and the MinIO console return 500.

`hive/metastore/resources.yaml:79` and `:177` have `S3_ACCESS_KEY: lake-writer` / `lake-reader` literal, where the MinIO base provisions those users and their Secrets from `${s3_lake_writer}` / `${s3_lake_reader}` — both site-supplied per `required-vars.txt`. Rename either and the metastore authenticates with a key that doesn't exist: S3 `InvalidAccessKeyId` on every Delta write and Trino read, after a bootstrap that looked fine.

`keycloak/instance/resources.yaml:267` has `KEYCLOAK_USER: admin` hardcoded, which is also the chart default, while the password comes from `keycloak-admin-secret` and that Secret's username key goes unused. The CR's `bootstrapAdmin` reads both from it and Ansible had `keycloak_bootstrap_admin_user`. Seed a different admin name and realm import 401s.

`postgres/cluster/resources.yaml:18` emits `${postgres_storage_class}` unconditionally. Every Ansible consumer wrote `| default(omit, true)` and all those vars default to empty, so the PVC came up with no `storageClassName` and inherited the cluster default. Now an unset value becomes `storageClassName: ""`, and `required-vars.txt` makes five `*_storage_class` vars mandatory that used to be optional.

deploy/base/superset/server/resources.yaml
```diff
@@ -0,0 +1,732 @@
...
+            metrics_thread.start()
+            logger.info("Started Superset metrics collection thread")
+
+            app.wsgi_app = DispatcherMiddleware(app.wsgi_app, {"/metrics": make_wsgi_app()})
+
+            return app
+
+
+        FLASK_APP_MUTATOR = setup_metrics
+      trino_auth: |
```
kathrynalpert: These three `configOverrides` inline copies of files from `ansible/roles/superset/files/` have already drifted — `prometheus_metrics` on line 419 drops 26 comment lines from `superset_metrics.py`, and the `superset-statsd-mapping` ConfigMap on line 54 reflows the comment banners in `statsd_mapping.yaml`. A diff guard would need normalization before it could even be written, which is the argument for not keeping two copies.

`trino_auth` is the one I'd worry about. It reproduces `superset_trino_auth.py` including the `DB_CONNECTION_MUTATOR` gating that decides whether `X-Trino-User` gets set (the request-context vs. Celery-task branch). Fix that logic on the Ansible side and GitOps sites keep the old branch, either clamping async queries or impersonating where they shouldn't, with nothing failing to show it.

The Superset chart config is Scout-owned, so a `configMapGenerator` with `files:` pointing at the same paths Ansible reads would leave one copy.

deploy/base/opa/resources.yaml
```diff
@@ -0,0 +1,605 @@
...
+      # bundleReader.credsHash (the Ansible role's rollout trigger on credential
+      # rotation) is intentionally omitted: hashing secret values in git is not
+      # allowed, and rotation-triggered rollout is deferred to Phase 4 (ESO/Reloader).
+      secretName: opa-bundle-reader
+    # Static AuthZ policy: a verbatim copy of policy/trino/main.rego (the file the
+    # opa Ansible role loads via lookup('file')). The chart bundles no default, so
+    # the base must supply it. Keep in sync with policy/trino/main.rego until the
+    # config-artifact publish job owns injecting it.
+    policy:
+      rego: |
```
kathrynalpert: This is byte-identical to `policy/trino/main.rego` — I diffed them, both 465 lines, no differences — inlined into HelmRelease values with a "keep in sync" comment and nothing enforcing it.

ADR 0020 makes `policy/trino/main.rego` authoritative, `ansible/roles/opa/tasks/deploy.yaml:13` loads it with `lookup('file', ...)`, and the `data-authorization` tests exercise it. Nothing diffs the two; `validate-deploy.yaml` only renders bases. So the next row-filter or column-mask change lands there, passes every check, and GitOps sites keep enforcing the old policy — users still seeing rows and unmasked identifiers the current policy denies, with no test failing anywhere.

`scout-opa` is our chart, so shipping `main.rego` as a chart file read through `.Files.Get` would remove the copy instead of guarding it.

Same file, around 560-576: `attribute_filters`, `masked_columns`, `view_owner_principals` and `filtered_tables` are literals in the data document. `scout_common/defaults/main.yaml:410` calls `trino_attribute_filters` a "single source of truth consumed by two roles" — Keycloak renders the user-profile attribute from it, OPA renders the filter. Here the Keycloak half arrives in the CI-seeded realm Secret and the OPA half is a git literal, so the two can disagree with nothing noticing.

deploy/base/hive/db/resources.yaml
```diff
@@ -0,0 +1,120 @@
...
+#   hive-metastore-secret       — HIVE_METASTORE_PASSWORD = hive owner password
+#   hive-metastore-readonly-secret — HIVE_METASTORE_PASSWORD = hive_readonly password
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: hive-db-init
+  namespace: "${hive_namespace}"
+spec:
+  backoffLimit: 3
+  ttlSecondsAfterFinished: 300
```
kathrynalpert: `ttlSecondsAfterFinished` on a Flux-managed Job loops forever: the TTL controller deletes the finished Job, `kustomize-controller` re-applies the manifest set on its interval, the Job comes back and runs again.

All four bootstrap Jobs do this — the three db-inits at 300s, `bootstrap-minio-iam` at 600s — so on a 30m interval that's four psql/mc pods hitting the CNPG primary and the MinIO tenant every half hour for the life of the cluster, each with a `wait: true` health assessment behind it. Nothing breaks, since the SQL and `mc` calls are idempotent, but it's permanent churn that'll read like a reconciliation bug to whoever finds it in six months.

Dropping the TTL leaves the completed Jobs in place, which is what Flux expects to own anyway.

deploy/flux/extractor.yaml
```diff
@@ -0,0 +1,27 @@
...
+  wait: true
+  timeout: 15m
+  sourceRef:
+    kind: OCIRepository
+    name: scout-config
+  postBuild:
+    substituteFrom:
+      - kind: ConfigMap
+        name: cluster-vars
+  dependsOn:
```
kathrynalpert, line 23: This lists `postgres-cluster`, `minio-tenant` and `temporal-bootstrap` but not `hive-metastore` or `trino-rw`, and the hl7-transformer values hardwire both — `hiveMetastoreUri: ${hive_metastore_endpoint}` on line 256, `TRINO_HOST: ${trino_rw_endpoint_host}` with `TRINO_VIEW_ENABLED: "true"` on 204-214.

`ansible/playbooks/main.yaml` runs Install Lake and Install Trino before Install Extractor, so the header comment on line 3 about the edges reproducing the Ansible order doesn't hold for this one.

Cold stand-up, Flux reconciles extractor as soon as postgres/minio/temporal are Ready, ahead of the metastore and trino-rw. First `IngestHl7ToDeltaLake` activity fails on Hive metastore connection refused and `deriveDeltaTables` fails on the `_epic_view` DDL — a race the ingest gate loses instead of waiting.

deploy/flux/postgres.yaml
```diff
@@ -0,0 +1,53 @@
...
+  wait: true
+  timeout: 5m
+  sourceRef:
+    kind: OCIRepository
+    name: scout-config
+  postBuild:
+    substituteFrom:
+      - kind: ConfigMap
+        name: cluster-vars
+  healthChecks:
```
kathrynalpert: Flux ignores `spec.healthChecks` when `spec.wait` is true, since `wait` already health-checks everything it reconciles, so this block does nothing. It also hardcodes a guess at the chart's Deployment fullname, so if the release or chart name changes a reader thinks there's a gate here that never existed.

It's the only Kustomization using `healthChecks` — the other four CR gates use `healthCheckExprs` — so worth deleting before the pattern gets copied into the next operator base.

deploy/base/hive/db/resources.yaml
```diff
@@ -0,0 +1,120 @@
...
+# hive covers the tables the write metastore's schematool creates on first boot,
+# so readonly works even though this Job runs before the metastore (unlike the
+# Ansible order). On-prem CNPG superuser is a real superuser, so the aws_deployment
+# `GRANT hive TO CURRENT_USER` dance is not needed and is omitted.
+# Secrets are referenced by fixed name only (seeded by CI/site, never in git):
+#   superuser-secret            — CNPG basic-auth (username/password), also seeded here
+#   hive-metastore-secret       — HIVE_METASTORE_PASSWORD = hive owner password
+#   hive-metastore-readonly-secret — HIVE_METASTORE_PASSWORD = hive_readonly password
+apiVersion: batch/v1
+kind: Job
```
kathrynalpert, line 7: The three db-init Jobs don't have the `kustomize.toolkit.fluxcd.io/force: "Enabled"` annotation that `bootstrap-minio-iam` carries, and `minio/tenant/resources.yaml:223` spells out why it needs it: "Immutable Job fields change with the config."

Same applies here. Add a `GRANT` to the readonly SQL, publish, and you get `Job.batch "hive-db-init" is invalid: spec.template: field is immutable` — hive-db goes permanently NotReady with hive-metastore, trino-ro and superset-server behind it. Lower priority since it needs a future edit to trigger, but it's a one-line add now.

.github/workflows/validate-deploy.yaml
```diff
@@ -0,0 +1,98 @@
...
+  validate:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Install kustomize
+        run: |
+          set -euo pipefail
+          VER=5.5.0
+          curl -sSL "https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize%2Fv${VER}/kustomize_v${VER}_linux_amd64.tar.gz" \
```
kathrynalpert: No checksum and no `curl -f` here, extracting as root straight from the stream.

`.github/actions/install-hauler-oras` already does this properly, with pinned versions, explicit `*_SHA256` values, and a comment stating the rule: "Binary downloads (not curl | bash), sha256-verified before extract so a tampered/hijacked release asset can't execute in the job."

As written a hijacked or corrupted asset executes in CI, and a transient GitHub error page gets untarred silently instead of failing. Same shape as that action — pinned version, `SHA256`, `sha256sum -c` — or just extend it to cover kustomize.

.github/workflows/validate-deploy.yaml
```diff
@@ -0,0 +1,98 @@
...
+            | sudo tar xz -C /usr/local/bin kustomize
+          kustomize version
+
+      - name: Render every base
+        run: |
+          set -euo pipefail
+          fail=0
+          while read -r d; do
+            if kustomize build "$d" >/dev/null; then
```
kathrynalpert, line 44: This makes three network calls on every PR run, because `deploy/base/keycloak/operator/kustomization.yaml` lists the Keycloak CRDs and `kubernetes.yml` as remote `raw.githubusercontent.com` resources.

So a raw outage, TLS hiccup or anonymous rate limit fails Validate-deploy on PRs that never touched Keycloak, and it surfaces as "kustomize build failed" with nothing pointing at the network.

`kustomize-controller` can't fetch those in an air-gapped cluster either, and the base's own comment already flags vendoring as a pre-go-live task. Doing it now covers both.

.github/workflows/validate-deploy.yaml
```diff
@@ -0,0 +1,98 @@
...
+          nodes = {}; edges = []
+          for f in sorted(glob.glob("deploy/flux/*.yaml")):
+              try:
+                  docs = list(yaml.safe_load_all(open(f)))
+              except Exception as e:
+                  print(f"::error file={f}::YAML parse failed: {e}"); sys.exit(1)
+              for d in docs:
+                  if not d:
+                      continue
+                  nodes[d["metadata"]["name"]] = f
```
kathrynalpert: Two bugs here, and one of them is a hole in the check that would otherwise catch the missing `dependsOn` edges I commented on in `deploy/flux/extractor.yaml`.

`nodes[d["metadata"]["name"]] = f` overwrites on a repeated name, so two Kustomizations that accidentally share a `metadata.name` across files pass this gate and then collide in `flux-system`, where the second takes over the first's inventory and prunes its resources. Counting names before inserting catches that.

Separately, when a `dependsOn` names a Kustomization that doesn't exist, its dependents never get their indegree decremented, so the topological sort under-counts and the job prints "the Flux dependsOn graph has a cycle (N/M sortable)" on top of the real dangling-ref error — which sends you looking for a cycle that isn't there. Skipping edges with a missing target before the sort fixes it.

deploy/base/hive/db/resources.yaml
```diff
@@ -0,0 +1,120 @@
...
+spec:
+  backoffLimit: 3
+  ttlSecondsAfterFinished: 300
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: hive-db-init
+    spec:
+      restartPolicy: Never
+      containers:
```
kathrynalpert, line 22: Grouping the duplication in one comment. I'd normally leave this alone on a first-pass port, but it's load-bearing here — several of the bugs above land two or three times because of these copies, and this tree becomes the source of truth, so it's cheaper now than after the next few components land on top.

These three db-init Jobs repeat about 60 identical lines each — same `postgres:16` image, securityContext, `HOME`/`PG*` env, `DO $do$ … CREATE USER` + `\gexec` shell, tmp emptyDir — differing only in database name and secret key. Ansible said it once in `scout_common/tasks/create_db_with_owner` with `db_name`/`db_owner`. It's why the two heredoc bugs land in triplicate and probably why the extractor's missing Job slipped through. With CNPG, `spec.managed.roles` + `bootstrap.initdb` on the Cluster could replace all three.

`hive/metastore/resources.yaml:119` is a 93-line near-verbatim duplicate of the release above it. 17-115 and 119-211 differ in name/`fullnameOverride`, `HIVE_METASTORE_USER`, `S3_ACCESS_KEY` and the envFrom secret; chart ref, image, resources, both probes, warehouse dir and S3 settings are all repeated. Ansible rendered both from one `deploy_metastore.yaml` include with three variables — and this is why the hardcoded access key appears twice.

`trino/rw/resources.yaml:70` shares about 100 lines with the trino-ro base, including a byte-identical jmx exporter block (`trino/ro:153-169`) and four repeats of the same coordinator/worker shape. Change the shared half in one place and the two Trinos end up mismatched — Prometheus quietly stopping on one, say. A shared base with two patch overlays would keep ADR 0019's deliberate read-only/writable difference visible instead of buried in a diff.

`minio/tenant/resources.yaml:272` unrolls five users, seven buckets and five policy create/attach/detach blocks by hand, 17 repeats of three shapes, with each entity also spelled out in `tenant.buckets`/`tenant.users`, a per-policy ConfigMap, and a volume/volumeMount pair. Adding one bucket-and-user pair is five coordinated edits in one file, and missing the consoleAdmin detach leaves a service account with full admin on the lake. Two `for` loops and a projected policies volume would take ~65 lines to ~15.

`postgres/operator/resources.yaml:27` and 20-odd others re-pin chart and image versions as literals under `deploy/`. The comment says "Renovate-tracked", but `renovate.json5` limits `managerFilePatterns` to `versions.yaml`, `keycloak/*` and `helm/superset/VERSION` — nothing in `deploy/`. Same literals recur across cassandra, elasticsearch, minio, oauth2-proxy, trino, valkey, temporal and superset plus image tags. Per ADR 0015 these are exactly what CVE monitoring should see, and the gap widens with each component converted.

`hive/metastore/resources.yaml:84` restates 14 empty cloud env stubs in both releases, because Helm replaces list values wholesale so every consumer setting `env` has to re-enumerate all of them or the entrypoint fails. That one's really a chart problem — better fixed in `helm/hive-metastore` by emitting the required stubs from the chart than worked around in each caller.

deploy/base/opa/resources.yaml
kathrynalpert: The inlined policy is still byte-identical to `policy/trino/main.rego` across all 465 lines, and nothing diffs them. Since `scout-opa` is our chart, `.Files.Get` on `main.rego` would leave one copy and no check needed - wouldn't that be better than duplicating?

deploy/base/superset/server/resources.yaml
kathrynalpert: Both copies differ from their Ansible sources right now (comment-only in both cases, so no behavior difference), so nothing detects drift. As above, is there a way to avoid duplication?
Claude suggests: `configMapGenerator` with `files:` pointing at the same paths Ansible reads would collapse all three to one copy.

deploy/base/hive/db/resources.yaml
kathrynalpert: The SQL fix here is good — `format(..., %L)` with `WHERE NOT EXISTS` is both interpolation-correct and idempotent, which is better than what I suggested, and it's applied consistently across all the Jobs including the new `extractor/db-init`.

Can we DRY this out? Fixing the missing `ingest` database needed a Job, and with no shared form to reuse it became a fourth ~60-line copy of the same skeleton in `deploy/base/extractor/db-init/resources.yaml`. The SQL bug had to be fixed in three places, the `$$` escaping in four, the `force` annotation in four, and the TTL removal in four. All four were done correctly this round, so this isn't a complaint about the fixes. It's that the next component makes five, and each round of edits is another chance for one of them to be missed.

Claude: CNPG's `spec.managed.roles` plus `bootstrap.initdb` on the Cluster would replace all four with declarative config on the object that owns the databases.

deploy/base/keycloak/instance/resources.yaml
kathrynalpert: `remediation.retries: 3` will most likely get there — four attempts at a 5m timeout is more runway than a first-boot migration usually needs, and the comment documents the race honestly.

While it's retrying, `keycloak-instance` reports NotReady, and oauth2-proxy, trino-ro and superset-server all `dependsOn` it with `wait: true`. So a slow Keycloak boot doesn't just delay the realm import, it holds the whole auth and analytics branch until the import converges. And if all four attempts are exhausted, nothing retries until the next 30m interval.

Splitting the import into its own Kustomization with `dependsOn: keycloak-instance` gets the same convergence from the DAG instead of from remediation, and the `healthCheckExprs` you already put on the Keycloak CR becomes the gate. Your call whether that's worth it now or Phase 4 — retries do work, they're just slower and noisier on a cold cluster.

deploy/base/superset/server/resources.yaml
kathrynalpert: One note on the storage-class item specifically, since it was fixed by removing `storageClassName` from all five bases rather than by making it optional. That's right for the default path: no key means the PVC inherits the cluster default, which is what Ansible produced when the var was unset.

It does drop the override, though. `ansible/inventory.example.yaml:220-225` lets a site pin `postgres_storage_class`, `valkey_storage_class`, `cassandra_storage_class`, `elasticsearch_storage_class` and `minio_storage_class`, and that's the mechanism the multidisk setup in ADR 0004 uses to put databases and object storage on different disks. As it stands `deploy/` can't express that, and the vars are no longer in `required-vars.txt` either.

deploy/base/hive/metastore/resources.yaml
```diff
@@ -0,0 +1,211 @@
...
+    # S3_SECRET_KEY (+ HIVE_METASTORE_PASSWORD) sourced from the fixed-name secret;
+    # seeded by CI/site (Phase 3) or ESO (Phase 4), never committed.
+    envFrom:
+      - secretRef:
+          name: hive-metastore-secret
+---
+# Readonly metastore — serves the user-facing (read-only) Trino. Same hive DB,
+# but connects as the SELECT-only `hive_readonly` role; S3 creds = lake-reader.
+apiVersion: helm.toolkit.fluxcd.io/v2
+kind: HelmRelease
```
kathrynalpert: The read-only release is still a near-copy of the one above it. Ignoring comments, the two differ in four values - `name`, `fullnameOverride`, `HIVE_METASTORE_USER` and `S3_ACCESS_KEY` - plus the envFrom secret name. Everything else is duplicated, which is why the `S3_ACCESS_KEY` fix this round had to be made twice. Ansible rendered both from one `deploy_metastore.yaml` include with three variables. Can we do the same here rather than maintain two copies?

Claude suggests: a shared HelmRelease base with two patch overlays for the four differing values.

deploy/base/hive/metastore/_base/helmrelease.yaml
```diff
@@ -0,0 +1,211 @@
...
+        value: 'true'
+      - name: S3_ENDPOINT
+        value: '${s3_endpoint}'
+      - name: S3_ACCESS_KEY
+        value: '${s3_lake_writer}'
+      - name: REGION
+        value: '${s3_region}'
+      - name: HIVE_METASTORE_USERS_IN_ADMIN_ROLE
+        value: admin
+      # All of these empty values must be present (image entrypoint requirement).
```
kathrynalpert: These 14 empty cloud stubs are restated in both releases (here and at :182), and every future consumer will have to restate them too, because Helm replaces list values wholesale and the entrypoint fails on a missing one.

`helm/hive-metastore` is ours - can we fix it there instead of duplicating the stubs in each caller?

Claude suggests: emit the required stubs from the chart and let `env` be additive, which drops 28 lines here.

deploy/base/trino/rw/resources.yaml
```diff
@@ -0,0 +1,168 @@
...
+      - secretRef:
+          name: trino-rw-s3
+    server:
+      workers: 1
+      config:
+        query:
+          # workers x worker_heap x per-node fraction = 1 x 1G x 0.3.
+          maxMemory: '0.3GB'
+    # JMX exporter for Prometheus (parity with the user-facing Trino).
+    jmx:
```
kathrynalpert: This jmx block is byte-identical to the one in `deploy/base/trino/ro/resources.yaml`, the coordinator/worker stanzas repeat the same shape across both files, and the two Delta catalogs differ only by metastore URI and the `READ_ONLY` line. Any change to the shared half needs both edits. Can these be a single source with the differences expressed as patches?

Claude suggests: a shared HelmRelease base with two overlays, so ADR 0019's read-only vs. writable difference is the only thing in the diff.

deploy/base/postgres/operator/resources.yaml
```diff
@@ -0,0 +1,30 @@
...
+  name: cnpg
+  namespace: '${cnpg_operator_namespace}'
+spec:
+  interval: 30m
+  # Version pinned in versions.yaml (cnpg_version), Renovate-tracked; stamped into
+  # the config artifact at publish. Literal here until the stamping job lands.
+  chart:
+    spec:
+      chart: cloudnative-pg
+      version: '0.28.3'
```
kathrynalpert, line 22: The comment says Renovate-tracked, but `renovate.json5` scopes `managerFilePatterns` to `versions.yaml`, `keycloak/*`, `helm/superset/*` and `helm/scout-dashboards/Chart.yaml` - nothing under `deploy/`. So this pin, and the twenty-odd others like it across cassandra, elasticsearch, minio, oauth2-proxy, trino, valkey, temporal and superset plus the image tags, are outside what CVE monitoring sees (ADR 0015).

The version now lives in `versions.yaml` and again here. Can we avoid the second copy?

If the config-artifact publish job will stamp these from the haul, then the pin here is a placeholder and the comment should say so rather than claiming it's tracked.

.github/workflows/ci.yaml
```diff
@@ -1379,3 +1335,119 @@ jobs:
...
+    permissions:
+      contents: read
+    env:
+      CONFIG_REPO: ghcr.io/washu-tag/manifests/scout-config
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - name: Install k3s + Flux
+        run: |
+          set -euo pipefail
+          curl -sfL https://get.k3s.io | sh -
```
kathrynalpert: `Semgrep OSS` is failing on these two lines, and it's an error-level rule so it blocks merge. It's also the same `curl | sh` pattern we fixed at `validate-deploy.yaml:27` this round - the k3s and Flux installers in the new `deploy-flux` job go straight into `sh -` and `sudo bash` with no checksum. The Flux one has no `-f` either, so a GitHub error page pipes into `sudo bash`.

Can these use the same pinned-version + sha256 approach as the kustomize install?

Claude: k3s publishes a per-release `sha256sum-amd64.txt` and Flux ships `install.sh` in its release assets, so both can be fetched to a file, verified, then executed - same shape as `.github/actions/install-hauler-oras`.

tooling/deploy/jvm_memory.py
```diff
@@ -0,0 +1,81 @@
+"""JVM heap -> Kubernetes memory conversion for the Phase 3 config-artifact lane.
+
+Byte-faithful stdlib port of ``ansible/filter_plugins/jvm_memory.py`` (the
```
kathrynalpert: This is a port of `ansible/filter_plugins/jvm_memory.py` - the two `jvm_memory_to_k8s` functions are 49 code lines each and 48 of the 49 are identical (the one difference is a trailing comment). `tooling/deploy/test_jvm_memory.py` is a second test suite for the same behaviour alongside `ansible/tests/unit/filter_plugins/test_jvm_memory.py`.

The docstring gives stdlib-only as the reason, but `ansible/filter_plugins/jvm_memory.py` imports nothing but `re` - `FilterModule` is just a class at the bottom, so importing the function doesn't pull in Ansible. Can we import it rather than port it?

Claude suggests: `sys.path` entry for `ansible/filter_plugins`, or move the pure function to a shared module both sides import, leaving `FilterModule` as the Ansible-side wrapper.

deploy/base/opa/resources.yaml
```diff
@@ -0,0 +1,99 @@
...
+          "filtered_tables": [
+            { "catalog": "delta", "schema": "default", "table_prefix": "${report_delta_table_name}" }
+          ],
+          "baseline_hidden_tables": [
+            { "catalog": "delta", "schema": "default", "table": "${report_delta_table_name}_report_patient_mapping" },
+            { "catalog": "delta", "schema": "default", "table": "${report_delta_table_name}_report_patient_mapping_history" }
+          ],
+          "hidden_tables": [],
+          "view_owner_principals": ["trino"],
+          "attribute_filters": {
```
kathrynalpert, line 67: `attribute_filters`, `masked_columns`, `view_owner_principals` and `filtered_tables` are literals here, but `scout_common/defaults/main.yaml:410` calls `trino_attribute_filters` a "single source of truth consumed by two roles" - Keycloak renders the user-profile attribute from it and OPA renders the filter. In this tree the Keycloak half arrives in the CI-seeded realm Secret and the OPA half is a git literal, so the two can disagree with nothing detecting it.

The policy itself is single-sourced now via `.Files.Get`. Can the data document come from one place too?

deploy/base/postgres/cluster/resources.yaml
```diff
@@ -0,0 +1,150 @@
+# Shared CNPG cluster. Roles + databases provisioned declaratively via
+# spec.managed.roles + kind: Database below (not bootstrap.initdb, which only
+# covers one DB on an uninitialized cluster); Database CRs reconcile continuously.
+# Secrets (fixed names, seeded by CI/site, never committed):
+#   superuser-secret — CNPG basic-auth for the operator.
+#   cnpg-role-*      — one basic-auth Secret per managed role in THIS namespace.
+#     CNPG drives each role's password FROM cnpg-role-*, so it MUST be seeded from
```
kathrynalpert: Checking I read this right: `cnpg-role-hive.password` and `hive-metastore-secret.HIVE_METASTORE_PASSWORD` now both have to carry the same value, and that's true for all five roles, so it's 9 secrets that have to agree pairwise instead of 4 standalone ones. The old psql Job read the app secret directly, so there was one source.

The header documents it and points at ESO in Phase 4, which seems right. Two questions in the meantime: is there anything that catches a mismatch before it shows up as a runtime auth failure, and could the apps read their password from `cnpg-role-*` so there's only one secret per role?

deploy/README.md
```diff
@@ -0,0 +1,55 @@
+# `deploy/` — Scout GitOps deployment base (ADR 0031, Phase 3)
+
+Kustomize bases + Flux `Kustomization`s that stand Scout up by *pulling* signed,
+digest-pinned artifacts instead of the Ansible push. WIP scaffold: the ingest
+vertical slice first (postgres -> lake -> orchestrator -> extractor), unconsumed
+until CI switches `deploy-and-test` to deploy from it. See
+`docs/internal/phase3-implementation-plan.md`.
```
kathrynalpert: `docs/internal/phase3-implementation-plan.md` doesn't exist on this branch or on `main`, so this is the one pointer to the design context and it's a dead link.

tooling/deploy/stamp_config.py
```diff
@@ -0,0 +1,303 @@
...
+                )
+            if _V_CONFIG_HASH0.search(line):
+                problems.append("{}:{}: config-hash still 0".format(rel, n))
+            # washu values.image tag must not be latest/0.0.0
+            m = _RE_REPO.match(line)
+            if m:
+                pending_repo = m.group(1)
+                continue
+            m = _RE_TAG.match(line)
+            if m and pending_repo is not None and m.group(1) == pending_repo:
```
kathrynalpert: `verify_clean` reuses the stamper's assumption that `repository:` immediately precedes `tag:` (same pairing as lines 156-173), so it can't catch the case the stamper misses. Swapping those two keys - identical YAML semantics - escapes both:

```yaml
    image:
      tag: latest
      repository: ghcr.io/washu-tag/hl7-transformer
```

Ran it against this file: no stamp for that block, and `verify_clean` returns no problems. The same block with `repository:` first stamps to the haul tag correctly. So a key reorder in any values block ships `:latest` to a Flux cluster with the fail-closed gate green, which is the pinning ADR 0030 is there to guarantee. The inline `image:` form is fine - `_V_IMG_BAD` is order-independent.

Both need fixing, not just the check - key order carries no meaning in YAML, so neither the stamper nor the verifier can depend on it. A stamper that skips a valid values block is a broken publish even if the verifier catches it.

Can both locate the images by parsing rather than by matching adjacent lines? The docstring says stdlib-only, but PyYAML is already used in `validate-deploy.yaml:111` for the DAG checker and arrives transitively via `ansible-core` for these tests, and this only runs in CI.

Claude suggests: `yaml.compose_all()` gives nodes carrying `start_mark.line`, so one shared helper can return every washu image mapping with its `tag:` line number - `stamp_lines` rewrites just those lines (comments and formatting survive, no `ruamel.yaml` needed) and `verify_clean` checks the values. Tried it against `deploy/base/extractor/resources.yaml`: finds both image blocks and the same two tag lines with the keys in either order.

andrey-embarklabs:
thanks for the thorough pass. pushed fixes for all of it, so the branch is a few commits past what you reviewed (8407deb4) and the threads show as outdated. quick rundown:

- the curl | sh k3s/flux installers now download the pinned binary, verify sha256, then run (same shape as install-hauler-oras), so the semgrep block clears. the flux -f-less error-page-into-sudo-bash case is gone too.

- stamp_config was the real bug, nice repro. rewrote it and verify_clean to locate images/charts/config-hash by parsing (yaml node line numbers) instead of adjacent lines, so tag: before repository: (or any reorder) gets stamped now. added a regression test for that exact block.

- jvm_memory: dropped the port, gen_cluster_vars imports the function from ansible/filter_plugins now (it only pulls in re), so it's one impl and the canonical test stays in ansible/.
- opa data doc: added a validate-deploy check that fails red if the deploy literal drifts from the ansible trino_* vars, and put the ansible defaults in the trigger paths so an ansible-side edit re-runs it. the real single-source across opa and keycloak is the config-artifact realm generation, tracked as a follow-up.
- cnpg-role: yeah, structural. cnpg reads passwordSecret in the cluster namespace and the apps read in their own, so the value lives in two places, and a mismatch only shows up as a runtime login failure (no clean pre-runtime check). eso collapses it to one source in phase 4. tightened the header to say that.
- readme link fixed, points at gitops-implementation-plan.md now.

tooling/deploy/stamp_config.py
```diff
@@ -0,0 +1,392 @@
...
+        docs = list(yaml.compose_all(text))
+    except yaml.YAMLError as exc:
+        # Fail closed: a deploy file we can't parse can't be checked for
+        # placeholders, so refuse to publish rather than skip it silently.
+        raise StampError("{}: not valid YAML, cannot stamp: {}".format(rel, exc))
+    lines = text.split("\n")
+    patches: list = []
+
+    def add(node, new_value, quote, kind, name, tag):
+        ln = node.start_mark.line
```
kathrynalpert, line 172: Taking `node.start_mark.line` and rewriting that whole line (via `lines[ln]` here, applied at :267) holds only while each patched node owns its own line. A flow mapping breaks that:

```yaml
image: { repository: ghcr.io/washu-tag/hl7-transformer, tag: latest }
```

The `tag` node starts on the `image:` line, so the rewrite replaces the whole mapping - the file ends up as `image: '0.20260814.5'`, `image` is a string instead of a dict, and `verify_clean` returns no problems because there's no `tag: latest` left. So this one corrupts rather than skips. Related: `:267` does `lines[line0] = new_line`, so two patches landing on one line would silently keep the last.

Not reachable right now - no flow mappings anywhere in `deploy/`, no anchors or merge keys either (a merged `repository` would be missed the same way, since it isn't a direct key), and prettier doesn't turn block mappings into flow. So this is about the assumption rather than a live bug.

Given that, is it worth teaching the rewriter flow style, or just asserting the assumption so it can't silently corrupt?

Claude suggests: in `add`, fail closed when the node doesn't own its line - `node.start_mark.line != node.end_mark.line`, or a second patch targeting a line index already patched.

PR #641, opened 2026-08-12
ci: cut over to release-please-driven releases (Phase 1)

andrey-embarklabs:
### Product
None. Release-automation cutover (ADR 0030, Phase 1).

### Technical
Makes release-please **live**: merging its accumulating release PR now cuts the release, so the version and changelog are computed from Conventional-Commit history instead of a hand-typed `workflow_dispatch` version. The manual dispatch stays as a fallback and for recovery.

The hardened `release.yaml` lane is preserved (it still stamps versions, waits for the build, verifies images published, tags, creates the Release, and resets to dev). The cutover wires the release-PR merge to it without changing that flow's guarantees:

- **release-please**: out of shadow (`draft-pull-request: false`); `skip-github-release: true` so it maintains the PR + manifest but does not double-cut the tag/Release (release.yaml stays the single releaser, so the image-verify-before-release guard is kept); `paths-ignore` its own merge commit.
- **`release-dispatch.yaml`** (new): the release-PR merge bumps `.release-please-manifest.json` (the only push that touches it), which triggers this workflow. It creates a boundary tag `v<version>` on the merge commit so release-please can scope its changelog during the build instead of bootstrapping a spurious next-version PR, then dispatches `release.yaml`.
- **`release.yaml`**: `validate` now tolerates a tag-without-Release (fails only if both exist, preserving already-released protection and retryability); the release job anchors the tag to the built commit before creating the Release; a new `verify-dev-reset` guardrail fails loud if a *successful* release left main stamped.
- **`update-versions.sh`**: the dev reset is now idempotent (a no-op is tolerated when resetting to dev; a real stamp still fails on a non-matching pattern).
- **`pr-title-lint.yaml`**: adds the upgrade-notes gate (a breaking PR must fill the "Backward compatibility" section) — the last Phase-1 PR-quality gate (supersedes #632).

### Prerequisite (repo settings, not in this PR)
release-please trusts squashed Conventional-Commit subjects, so before relying on the computed version, enable on `main`: **squash-only merges** (squash message = PR title), **require linear history**, **require status checks up-to-date**. (Greenlit by Kate.)

### Residual known-limitations (documented in `release-dispatch.yaml`; follow-ups)
- A feature merged to `main` *during* a release is built into the images (the lane builds current HEAD by design, to fold in fix-forwards) but is not in the tag/notes. Avoid merging while a release is in flight; a release-in-progress lock is a possible follow-up.
- An abandoned release burns a version number and leaves a phantom changelog entry; recovery runbook is in the workflow comments.

## Type of change
- [x] CI / release automation (non-breaking)

### Backward compatibility
No product impact. `main` stays on dev placeholders exactly as today (release.yaml's stamp-then-reset is unchanged and Ansible reads role defaults, not the files release-please writes). The manual `workflow_dispatch` release path is unchanged.

## Testing
`actionlint` clean on all touched workflows. `update-versions.sh` idempotency verified empirically (dev-on-dev → exit 0, zero diff; stamp→dev roundtrip → net zero; repeat stamp still errors). The design was reviewed adversarially across three rounds (release-please 17.x source + GitHub REST semantics); the flow will exercise end-to-end on the first real release PR.

## Note for reviewers
The intent is to leave the battle-tested `release.yaml` behavior intact and only wire the trigger + guardrails around it. Key things to sanity-check: `skip-github-release` + the boundary tag (so release-please never bootstraps), and the `verify-dev-reset` guardrail's scope (successful-release-but-stamped-main only).

### DIFF
--- a/.github/scripts/update-versions.sh
+++ b/.github/scripts/update-versions.sh
@@ -20,6 +20,9 @@ if [[ "$VERSION" == "dev" ]]; then
     DOCKER_TAG="$DEV_DOCKER_TAG"
     HELM_VERSION="$DEV_HELM_VERSION"
     PYTHON_VERSION="$DEV_PYTHON_VERSION"
+    # A dev reset is idempotent: a file already at its dev value is expected (e.g. a
+    # component that was never stamped, or re-running the reset), so a no-op is fine.
+    ALLOW_NOOP=true
     echo "Resetting to dev versions..."
 else
     # Validate release version format
@@ -30,6 +33,9 @@ else
     DOCKER_TAG="$VERSION"
     HELM_VERSION="$VERSION"
     PYTHON_VERSION="$VERSION"
+    # During a real stamp a no-op means the pattern did not match (e.g. a renamed
+    # variable) and must fail loudly, so no-ops are NOT tolerated here.
+    ALLOW_NOOP=false
     echo "Updating version files to $VERSION..."
 fi
 
@@ -54,11 +60,17 @@ update_file() {
     # Using -i.bak creates a backup we can compare against to verify the change.
     sed -i.bak -E "s|$pattern|$replacement|" "$file"
 
-    # Fail if sed didn't change anything (pattern may not have matched)
+    # sed made no change. In a dev reset that just means the file is already at its
+    # dev value (tolerated, keeps the reset idempotent). During a stamp it means the
+    # pattern did not match, which must fail.
     if cmp -s "$file" "${file}.bak"; then
+        rm -f "${file}.bak"
+        if [[ "${ALLOW_NOOP:-false}" == "true" ]]; then
+            echo "  - $description: $file (already at target)"
+            return 0
+        fi
         echo "  ERROR: No changes made to $file for $description"
         echo "         Pattern may not have matched: $pattern"
-        rm -f "${file}.bak"
         exit 1
     fi
 
--- a/.github/workflows/release-dispatch.yaml
+++ b/.github/workflows/release-dispatch.yaml
@@ -1,0 +1,76 @@
+name: Release Dispatch
+
+# Bridges release-please (LIVE) to the release lane. When a release-please PR is
+# merged, its squashed commit bumps .release-please-manifest.json on main -- the
+# ONLY push that touches that file -- so the path filter below is an exact "a
+# release was just approved" signal. On that signal this workflow:
+#
+#   1. Creates the boundary tag v<version> on the merge commit. release-please runs
+#      with skip-github-release, so it never tags; without a tag it would, during
+#      release.yaml's long build, see a manifest version ahead of the last tag and
+#      bootstrap a spurious next-version PR from repo start. The pre-created tag is
+#      the marker that keeps release-please's changelog scoping correct.
+#   2. Dispatches release.yaml, which stays the single authoritative lane: stamp
+#      versions -> wait for the build -> verify images published -> create the GitHub
+#      Release ON THIS TAG -> reset to dev. The Release is minted LAST, after image
+#      verification, so a failed build never yields a Release (only the boundary tag,
+#      which is retryable via a manual re-dispatch and which release.yaml tolerates).
+#
+# Recovery + known limitations:
+#   - Failed release (main left stamped): release.yaml's verify-dev-reset job fails
+#     LOUD. Retry the SAME version with:  gh workflow run release.yaml --ref main
+#     -f version=X.Y.Z  (validate tolerates the boundary tag; version-bump reuses the
+#     existing bump commit).
+#   - Abandoned release: the boundary tag v<version> persists, so release-please treats
+#     the version as released and the next PR proposes the NEXT number (the number is
+#     burned) while a CHANGELOG entry for it stays on main. To RECLAIM the number
+#     instead: delete tag v<version> and its Release FIRST, then revert the release-PR
+#     merge (drops the manifest + CHANGELOG bump); release-please then re-proposes it.
+#     Note the revert changes the manifest, so it re-triggers this workflow; deleting
+#     the tag+Release first lets that run complete cleanly instead of failing at
+#     release.yaml's validate on an already-released version.
+#   - A feature PR merged to main DURING a release is built into the release images
+#     (release.yaml builds current HEAD by design, to fold in fix-forward commits) but
+#     is not in the tag/notes. Avoid merging to main while a release is in flight; a
+#     release-in-progress lock on main is a possible future hardening.
+on:
+  push:
+    branches: [main]
+    paths:
+      - .release-please-manifest.json
+
+concurrency:
+  group: release-dispatch
+  cancel-in-progress: false
+
+permissions:
+  contents: write # create the boundary tag
+  actions: write # dispatch release.yaml (workflow_dispatch via GITHUB_TOKEN is allowed to trigger a run)
+
+jobs:
+  dispatch:
+    name: Tag + dispatch release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Create the boundary tag + dispatch release.yaml
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          VERSION="$(jq -r '."."' .release-please-manifest.json)"
+          if ! [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "::error::manifest version '${VERSION}' is not X.Y.Z; not releasing"
+            exit 1
+          fi
+          if gh api "repos/${GITHUB_REPOSITORY}/git/ref/tags/v${VERSION}" >/dev/null 2>&1; then
+            echo "::notice::tag v${VERSION} already exists; leaving it in place"
+          else
+            echo "::notice::creating boundary tag v${VERSION} at ${GITHUB_SHA}"
+            gh api -X POST "repos/${GITHUB_REPOSITORY}/git/refs" \
+              -f ref="refs/tags/v${VERSION}" -f sha="${GITHUB_SHA}"
+          fi
+          echo "::notice::dispatching release.yaml for v${VERSION}"
+          gh workflow run release.yaml --ref main -f version="${VERSION}"
--- a/.github/workflows/release-dispatch.yaml
+++ b/.github/workflows/release-dispatch.yaml
@@ -23,12 +23,10 @@ name: Release Dispatch
 #     existing bump commit).
 #   - Abandoned release: the boundary tag v<version> persists, so release-please treats
 #     the version as released and the next PR proposes the NEXT number (the number is
-#     burned) while a CHANGELOG entry for it stays on main. To RECLAIM the number
-#     instead: delete tag v<version> and its Release FIRST, then revert the release-PR
-#     merge (drops the manifest + CHANGELOG bump); release-please then re-proposes it.
-#     Note the revert changes the manifest, so it re-triggers this workflow; deleting
-#     the tag+Release first lets that run complete cleanly instead of failing at
-#     release.yaml's validate on an already-released version.
+#     burned) while a CHANGELOG entry for it stays on main. Reverting the release PR
+#     re-triggers this workflow with the PRIOR manifest version, which validate rejects
+#     as already-released -- harmless (nothing gets stamped), but not a clean reclaim;
+#     renumbering past a burned version is manual.
 #   - A feature PR merged to main DURING a release is built into the release images
 #     (release.yaml builds current HEAD by design, to fold in fix-forward commits) but
 #     is not in the tag/notes. Avoid merging to main while a release is in flight; a
--- a/.github/workflows/release-dispatch.yaml
+++ b/.github/workflows/release-dispatch.yaml
@@ -42,8 +42,8 @@ concurrency:
   cancel-in-progress: false
 
 permissions:
-  contents: write # create the boundary tag
-  actions: write # dispatch release.yaml (workflow_dispatch via GITHUB_TOKEN is allowed to trigger a run)
+  contents: read # checkout (the boundary tag is written with the app token)
+  actions: write # dispatch release.yaml (workflow_dispatch via GITHUB_TOKEN triggers a run)
 
 jobs:
   dispatch:
@@ -53,9 +53,18 @@ jobs:
       - name: Checkout
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
-      - name: Create the boundary tag + dispatch release.yaml
+      # App token so the boundary tag has the SAME writer as release.yaml's tag move
+      # (single identity, one bypass to configure if v* tags ever get a ruleset).
+      - name: Generate token from GitHub App
+        id: app_token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          client-id: ${{ secrets.RELEASE_APP_ID }}
+          private-key: ${{ secrets.RELEASE_APP_PRIVATE_KEY }}
+
+      - name: Create the boundary tag
         env:
-          GH_TOKEN: ${{ github.token }}
+          GH_TOKEN: ${{ steps.app_token.outputs.token }}
         run: |
           set -euo pipefail
           VERSION="$(jq -r '."."' .release-please-manifest.json)"
@@ -63,6 +72,7 @@ jobs:
             echo "::error::manifest version '${VERSION}' is not X.Y.Z; not releasing"
             exit 1
           fi
+          echo "VERSION=${VERSION}" >> "$GITHUB_ENV"
           if gh api "repos/${GITHUB_REPOSITORY}/git/ref/tags/v${VERSION}" >/dev/null 2>&1; then
             echo "::notice::tag v${VERSION} already exists; leaving it in place"
           else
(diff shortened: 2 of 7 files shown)


.github/workflows/release.yaml
```diff
@@ -385,18 +395,32 @@ jobs:
...
           COMMIT_SHA: ${{ needs.wait-for-build.outputs.commit_sha }}
         run: |
+          set -euo pipefail
+          # release-dispatch.yaml pre-creates the boundary tag v${VERSION} at the
+          # release-PR merge commit so release-please can scope its changelog during
+          # the build. Now that the built + image-verified commit exists, point the tag
+          # at it so `git checkout v${VERSION}` reproduces the released source (parity
+          # with the manual-dispatch path, where the tag does not pre-exist and we
+          # create it). gh then anchors the Release to that tag, so no --target needed.
+          # The built commit is a descendant of the merge commit and the only commit
```
kathrynalpert, line 431: The invariant in this comment — that the only commit between the two positions is
`Update to version X.Y.Z` — isn't enforced, and the retry path breaks it by design.
`wait-for-build` deliberately re-pulls main so a fix-forward gets included, so after a failed
attempt where someone merged `fix(ci): ...`, the tag moves from the merge commit *past* that
fix.

The tag is the only record of how far the release accounted for, so that fix then appears in
neither changelog: not in 4.2.0's (release-please wrote it at the merge commit, before the fix
existed) and not in 4.3.0's (it's behind the boundary). It also doesn't count toward the next
version bump, so a `feat!:` merged in that window would ship without a major.

The release-in-progress lock fixes this by construction. Until it exists, can we make the
violation loud? Roughly:

```bash
old="$(gh api "repos/${REPO}/git/ref/tags/v${VERSION}" --jq .object.sha)"
extra="$(git log --format='%s' "${old}..${COMMIT_SHA}" | grep -v "^Update to version ${VERSION}$" || true)"
[ -z "$extra" ] || { echo "::error::merged during the release, will be dropped from the changelog: ${extra}"; exit 1; }
```

This job's checkout is shallow, so it would need `fetch-depth: 0` (line 363).

If you'd rather not hard-fail before the lock lands, the alternative is to move the tag to the
stamp commit (`validate.outputs.version_bump_sha`) instead of `wait-for-build`'s HEAD. Then
anything merged during the release stays after the boundary and lands correctly in the next
changelog. The cost is that after a fix-forward the tag's tree no longer matches the images —
which I'd rather have than an invisible changelog hole.

Related, and part of why I want the lock: while main is stamped, any PR merged to main
republishes the `X.Y.Z` images over the ones this job just verified, because ci.yaml's
`publish` fires on `changes.<image> == 'true'` and the tag is mutable. Build-lane digests are
unaffected so the haul is fine, but `X.Y.Z` isn't immutable during that window. Pre-existing,
not this PR's doing.

.github/workflows/release.yaml
```diff
@@ -385,18 +395,32 @@ jobs:
...
+          # between them ("Update to version X.Y.Z") is not a Conventional Commit, so
+          # moving the tag forward does not disturb release-please's next-cycle scoping.
+          if gh api "repos/${REPO}/git/ref/tags/v${VERSION}" >/dev/null 2>&1; then
+            gh api -X PATCH "repos/${REPO}/git/refs/tags/v${VERSION}" -F sha="${COMMIT_SHA}" -F force=true
```
kathrynalpert: Nit: `-F` parses the value as a typed JSON literal, so an all-digit SHA would go out as a
number. `-f sha="${COMMIT_SHA}" -F force=true` is the safe form.

.github/workflows/release-dispatch.yaml
```diff
@@ -0,0 +1,76 @@
...
+#     -f version=X.Y.Z  (validate tolerates the boundary tag; version-bump reuses the
+#     existing bump commit).
+#   - Abandoned release: the boundary tag v<version> persists, so release-please treats
+#     the version as released and the next PR proposes the NEXT number (the number is
+#     burned) while a CHANGELOG entry for it stays on main. To RECLAIM the number
+#     instead: delete tag v<version> and its Release FIRST, then revert the release-PR
+#     merge (drops the manifest + CHANGELOG bump); release-please then re-proposes it.
+#     Note the revert changes the manifest, so it re-triggers this workflow; deleting
+#     the tag+Release first lets that run complete cleanly instead of failing at
+#     release.yaml's validate on an already-released version.
```
kathrynalpert: I don't think this recovery works as written. Reverting the release-PR merge restores the
manifest to the *previous* version, so the revert push dispatches `release.yaml` with
`version=4.1.0` — whose tag and Release both exist, so `validate` fails. Deleting v4.2.0's tag
and Release first doesn't change that.

It's harmless, since validate fails before anything is stamped, but the comment promises a
clean run you won't get. Suggest saying that instead: the revert re-triggers this workflow and
its run fails at `validate` on the previous version, and that failure is expected.

.github/workflows/release-dispatch.yaml
```diff
@@ -0,0 +1,76 @@
...
+#   - A feature PR merged to main DURING a release is built into the release images
+#     (release.yaml builds current HEAD by design, to fold in fix-forward commits) but
+#     is not in the tag/notes. Avoid merging to main while a release is in flight; a
+#     release-in-progress lock on main is a possible future hardening.
+on:
+  push:
+    branches: [main]
+    paths:
+      - .release-please-manifest.json
```
kathrynalpert, line 38: Worth noting in the comment block: this same push also kicks off a full `ci.yaml` run on the
merge commit, which publishes nothing (no image paths changed, and `latest` is already
published) but still spends a complete deploy-and-test cycle, and it overlaps the
version-bump run that lands a minute later. Two `publish-charts` jobs then race to advance the
floating `main` haul tag, and the order isn't deterministic.

Pre-existing hazard, but this flow guarantees it on every release rather than leaving it to
chance. Not asking for a fix here — just don't want it to surprise us the first time.

.github/workflows/release-dispatch.yaml
```diff
@@ -0,0 +1,76 @@
...
+          VERSION="$(jq -r '."."' .release-please-manifest.json)"
+          if ! [[ "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+            echo "::error::manifest version '${VERSION}' is not X.Y.Z; not releasing"
+            exit 1
+          fi
+          if gh api "repos/${GITHUB_REPOSITORY}/git/ref/tags/v${VERSION}" >/dev/null 2>&1; then
+            echo "::notice::tag v${VERSION} already exists; leaving it in place"
+          else
+            echo "::notice::creating boundary tag v${VERSION} at ${GITHUB_SHA}"
+            gh api -X POST "repos/${GITHUB_REPOSITORY}/git/refs" \
```
kathrynalpert, line 80: Minor: the tag is created here with `GITHUB_TOKEN` but moved in `release.yaml` with the
RELEASE_APP token. Works today, but since we're changing branch protection as a prerequisite
anyway — if we ever add a ruleset on `v*` tags, both identities need a bypass. Might be
simpler to mint the app token here too and keep a single writer for that tag.

.github/scripts/update-versions.sh
```diff
@@ -20,6 +20,9 @@ if [[ "$VERSION" == "dev" ]]; then
     DOCKER_TAG="$DEV_DOCKER_TAG"
     HELM_VERSION="$DEV_HELM_VERSION"
     PYTHON_VERSION="$DEV_PYTHON_VERSION"
+    # A dev reset is idempotent: a file already at its dev value is expected (e.g. a
+    # component that was never stamped, or re-running the reset), so a no-op is fine.
+    ALLOW_NOOP=true
```
kathrynalpert, line 25: Worth recording the interaction with `verify-dev-reset`: that job detects "main is still
stamped" by running this script in dev mode and looking for a diff. Since a dev reset now
tolerates a pattern that doesn't match, a renamed variable would let both the reset and the
verification pass while that one file stays stamped.

The stamp path still fails loudly, so it's caught at the next release — a "found one release
late" issue rather than a hole. Fine to leave as is; I just want it written down.

.github/workflows/release.yaml
```diff
@@ -464,3 +488,39 @@ jobs:
...
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+
+      - name: Assert VERSION files are back at dev placeholders
+        env:
+          VERSION: ${{ inputs.version }}
+          REF_NAME: ${{ github.ref_name }}
+        run: |
+          git pull origin "${REF_NAME}"
```
kathrynalpert: Small one: this pulls immediately after `reset-dev` pushed. A stale read here would fail a
release that actually succeeded, and the message tells the operator to go fix something that
isn't broken. Given how loud the failure is, a short retry (or re-checking once before
failing) might be worth it.

kathrynalpert (commented):
I like that `release.yaml` stays the engine with the manual dispatch surviving as a fallback.

Three requests:

**1. The tag move can silently drop commits from the changelog.** Details inline. The
release-in-progress lock you list as a possible follow-up seems like the fix, and I'd like us to
commit to it rather than leave it as "possible."

**2. Charts don't publish at `X.Y.Z`.** ADR 0030's release lane says "Charts publish as OCI
artifacts at `X.Y.Z`, at the same registry paths as the build lane," and the lane table says a
release covers all artifacts. Today `publish-charts` lives only in `ci.yaml`, always stamps the
build-lane version, and fires only on changed paths — `release.yaml` publishes no charts at
all. So once this lands, merging a release PR cuts a release that ADR 0030 says is incomplete.
Nothing breaks today because on-prem installs from the checkout rather than from OCI, but the
GitOps lane can't pin a release at all.

Claude suggests a job between `wait-for-build` and `release`, so it inherits the existing
"verify everything exists before minting the Release" guarantee:

```yaml
# per ADR 0030: any directory under helm/ with a Chart.yaml, excluding vendored charts/
helm package "$dir" --version "$VERSION" --app-version "$VERSION" -d "$RUNNER_TEMP"
helm push "$RUNNER_TEMP/$name-$VERSION.tgz" "oci://ghcr.io/washu-tag/charts"
```

Two gotchas:

- Pass `--version` / `--app-version` explicitly rather than relying on the stamped
  `Chart.yaml`. `update-versions.sh` stamps 12 of the 14 charts — `dcm4chee` and `orthanc`
  aren't in its list and would go out at `0.0.0-dev`.
- Verification is nearly free. `ghcr-tag-published.sh` takes the name as a path segment, so
  `ghcr-tag-published.sh charts/launchpad 4.2.0` works unmodified; the existing verify step
  just needs a second loop.

**3. Nothing maps `X.Y.Z` to a haul.** #642 and #643 are both build-lane only, so after this
lands an air-gapped site can transport a build but can't find the one that is 4.2.0. This
shouldn't need a second haul though — the build haul from the release's own CI run is already
the release content, so it may be as small as `oras tag`-ing that bundle to `X.Y.Z` once the
release is verified, the same way we advance `:main`. Cosign signatures are keyed by digest, so
the existing signature still verifies through the alias. Only catch is `release.yaml` needs the
build version of the run it waited on, which is probably easiest if `ci.yaml` surfaces it.

I (erm, Claude) added a few smaller nits inline.

andrey-embarklabs:
> I like that `release.yaml` stays the engine with the manual dispatch surviving as a fallback.
> 
> Three requests:
> 
> **1. The tag move can silently drop commits from the changelog.** Details inline. The release-in-progress lock you list as a possible follow-up seems like the fix, and I'd like us to commit to it rather than leave it as "possible."
> 
> **2. Charts don't publish at `X.Y.Z`.** ADR 0030's release lane says "Charts publish as OCI artifacts at `X.Y.Z`, at the same registry paths as the build lane," and the lane table says a release covers all artifacts. Today `publish-charts` lives only in `ci.yaml`, always stamps the build-lane version, and fires only on changed paths — `release.yaml` publishes no charts at all. So once this lands, merging a release PR cuts a release that ADR 0030 says is incomplete. Nothing breaks today because on-prem installs from the checkout rather than from OCI, but the GitOps lane can't pin a release at all.
> 
> Claude suggests a job between `wait-for-build` and `release`, so it inherits the existing "verify everything exists before minting the Release" guarantee:
> 
> ```yaml
> # per ADR 0030: any directory under helm/ with a Chart.yaml, excluding vendored charts/
> helm package "$dir" --version "$VERSION" --app-version "$VERSION" -d "$RUNNER_TEMP"
> helm push "$RUNNER_TEMP/$name-$VERSION.tgz" "oci://ghcr.io/washu-tag/charts"
> ```
> 
> Two gotchas:
> 
> * Pass `--version` / `--app-version` explicitly rather than relying on the stamped
>   `Chart.yaml`. `update-versions.sh` stamps 12 of the 14 charts — `dcm4chee` and `orthanc`
>   aren't in its list and would go out at `0.0.0-dev`.
> * Verification is nearly free. `ghcr-tag-published.sh` takes the name as a path segment, so
>   `ghcr-tag-published.sh charts/launchpad 4.2.0` works unmodified; the existing verify step
>   just needs a second loop.
> 
> **3. Nothing maps `X.Y.Z` to a haul.** #642 and #643 are both build-lane only, so after this lands an air-gapped site can transport a build but can't find the one that is 4.2.0. This shouldn't need a second haul though — the build haul from the release's own CI run is already the release content, so it may be as small as `oras tag`-ing that bundle to `X.Y.Z` once the release is verified, the same way we advance `:main`. Cosign signatures are keyed by digest, so the existing signature still verifies through the alias. Only catch is `release.yaml` needs the build version of the run it waited on, which is probably easiest if `ci.yaml` surfaces it.
> 
> I (erm, Claude) added a few smaller nits inline.

thanks for the thorough pass. rundown of what landed:

tag-move / mid-release drift: went with the pin-to-fixed-commit route over the branch lock. wait-for-build now targets the version-bump (stamped) commit instead of re-pulling HEAD, so a feature merged mid-release can't be swept into the images or past the tag. folding a fix into a failed release now goes through reset + re-dispatch.

charts at X.Y.Z: added a publish-charts job to the release lane that packages/pushes/signs all 14 charts at X.Y.Z from the stamped commit, so a release is a coherent artifact set per ADR 0030.

nits: -f sha (not -F), verify-dev-reset re-fetches before failing so a stale read can't fail a good release, and the abandoned-release comment now says what actually happens (the revert re-triggers with the prior version, which validate rejects, so not a clean reclaim).

also folded in: the release-PR merge no longer triggers a redundant ci.yaml run (paths-ignore on the push trigger, so PR checks are untouched), and the boundary tag is now minted with the app token so it has a single writer.

.github/workflows/release.yaml
```diff
@@ -464,3 +490,103 @@ jobs:
...
+          done
+          echo "::error::Release v${VERSION} succeeded but ${REF_NAME} is still STAMPED (reset-dev did not complete). A dev deploy would pull v${VERSION} instead of latest. Re-dispatch will NOT help (v${VERSION} is already released); reset manually:"
+          echo "::error::  git checkout ${REF_NAME} && bash .github/scripts/update-versions.sh dev && git commit -am 'Reset to dev versions' && git push"
+          git --no-pager diff --stat
+          exit 1
+
+  publish-charts:
+    name: Publish charts at the release version
+    runs-on: ubuntu-latest
+    needs: [validate, version-bump, release]
```
kathrynalpert: This landed *after* `release`: gated on `needs.release.result == 'success'`, the GitHub Release is minted first and the
charts publish after. This means there's no chart equivalent of the "Verify the release images were
published" step. If any of
the 14 legs fails — a transient GHCR 5xx, a cosign hiccup — the release is published with
charts partly missing, and the release job has finished, so nothing re-checks.

Suggest moving `publish-charts` between `wait-for-build` and `release` (`needs: [validate,
version-bump, wait-for-build]`) and adding a second loop to the existing verify step;
`ghcr-tag-published.sh charts/<name> "$VERSION"` works unmodified. Then charts inherit the same
"nothing is claimed until everything exists" property the images already have.

.github/workflows/release.yaml
```diff
@@ -233,15 +243,17 @@ jobs:
       - name: Determine target commit
         id: target
         env:
-          REF_NAME: ${{ github.ref_name }}
+          # Pin the release to the version-bump (stamped) commit, not current HEAD, so a
+          # feature merged mid-release can't be swept into the images or past the tag (it
```
kathrynalpert: The pin fixes the tag and changelog half, and it does it more cleanly than either option Claude
suggested — the tag can now only move from the merge commit to the stamp commit, so no
Conventional Commit can ever be swallowed.

But "can't be swept into the images" isn't accurate:
Images are published by whatever `ci.yaml` run fires, and while `main` is stamped, a PR merged
mid-release still derives `X.Y.Z` from the version files and pushes over the release tags —
`publish` gates on `changes.<image> == 'true'`, and the tag is mutable. Pinning changes which run
we *verify*, not who can push after we've verified it.

Closing this still requires a release-in-progress lock, I think. If we don't want to pursue that option, I'm ok with communication via Slack instead, but the comment should be corrected.

andrey-embarklabs:
> Looks like mapping `X.Y.Z` to a haul still isn't in here: there's no `oras` or haul reference anywhere in `release.yaml` (and #643 is missing retention for release-tagged hauls). Is this planned as a follow-up?

all three addressed, pushed as 3346e6ad:

X.Y.Z -> haul: new alias-haul job. after the release is minted it reads the build
version of the CI run wait-for-build waited on (ci.yaml now surfaces it as a
haul-version artifact) and oras-tags the manifest + bundle to X.Y.Z. cosign sig is
digest-keyed so it verifies through the alias, no re-sign. if the artifact isn't
there yet (the window before the build lane is live) it skips with a warning + the
manual oras command rather than redding a release that already succeeded.

publish-charts before release: moved to needs [validate, version-bump,
wait-for-build]; release now needs + gates on it and re-verifies all 14 chart tags
next to the images (ghcr-tag-published charts/<name> works unmodified), so nothing
is claimed until every artifact exists.

target-commit comment: corrected. it now says pinning fixes the tag/changelog
boundary but does NOT stop a mid-release merge overwriting the X.Y.Z images (mutable
tag + changes-gated publish). left the release-in-progress lock as the real fix /
tracked follow-up; i'll hold merges during a release in the meantime, or we do the
lock next, your call.

merge order: alias-haul reads the artifact #642's ci.yaml produces, so #642 wants to
land with or before this one.

PR #642, opened 2026-08-12
ci: finalize the build-lane haul (Phase 2 go-live)

andrey-embarklabs:
### Product
None. Turns on the ADR 0033 build-lane haul (Phase 2 finalize).

### Technical
Two changes to `ci.yaml`:
- **`publish-charts`** now force-publishes all 14 charts on a CI/build-tooling change (`changes.ci`), mirroring `build-and-upload` for images (ADR 0030: CI changes rebuild everything). A never-published chart otherwise fails the haul's fail-closed carry, so this seeds every chart into the registry.
- **`publish-haul`** is un-paused (it was `if: false` since #634, pending the seed).

**Merging this PR is the go-live.** The merge is a CI change, so on the first main run `changes.ci` is true → `publish-charts` publishes all 14 charts (10 of them for the first time) and signs them → `publish-haul` carries the 8 images at digest from the bootstrap predecessor and folds in the 14 fresh charts plus the wrapper-chart upstream images = the first **complete signed haul**. Every main build thereafter carries unchanged components by digest.

Rebuilt fresh onto current `main` (the previously-held `ci/relight-build-lane` conflicted on #635's action-SHA bumps; it's superseded).

## Type of change
- [x] CI / build automation

### Backward compatibility
No deploy impact. The haul job is `continue-on-error` and main-only; it publishes artifacts (charts + the haul bundle) and does not gate releases or touch the Ansible deploy path.

## Testing
`actionlint` clean on the diff. The seed + carry mechanics were validated during the Phase-2 build-out (bootstrap seeded the 8 images; predecessor-carry proven); this run exercises the chart seed end-to-end for the first time.

## Note for reviewers
After merge, `publish-haul` runs on every main build. The immediate fast-follow is **haul retention/pruning** (each haul is the whole ~5GB platform, republished each build; prune scoped to the two `manifests/*` packages only, keeping `:main` + the last N, never the digest-pinned image/chart packages). Other hardening (weekly full-rebuild backstop, checksum-pinned hauler/oras/cosign, superset version-sourcing) and the downstream enclave reconciler (Phase 3) are separate.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1011,6 +1011,10 @@ jobs:
   # Publish changed Helm charts as versioned OCI artifacts (ADR 0030/0031 phase 0),
   # gated like the image `publish` job. Vertical slice: hl7-transformer only.
   publish-charts:
+    # Per-chart steps publish a chart whose dir changed, OR every chart on a
+    # CI/build-tooling change (changes.ci), mirroring build-and-upload for images
+    # (ADR 0030: CI/build-tooling changes rebuild everything). That force-all also
+    # gives publish-haul its one-time all-14-charts seed on any ci-touching merge.
     if: github.ref_name == 'main' && needs.deploy-and-test.result == 'success' && needs.smoke-test.result == 'success'
     runs-on: ubuntu-latest
     permissions:
@@ -1066,14 +1070,14 @@ jobs:
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - name: Login to GHCR (for the imagetools digest read-back)
-        if: needs.changes.outputs[matrix.changed] == 'true'
+        if: needs.changes.outputs[matrix.changed] == 'true' || needs.changes.outputs.ci == 'true'
         uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: 'Publish chart - ${{ matrix.chart-name }}'
-        if: needs.changes.outputs[matrix.changed] == 'true'
+        if: needs.changes.outputs[matrix.changed] == 'true' || needs.changes.outputs.ci == 'true'
         shell: bash
         # helm is preinstalled on the runner (OCI push needs >= 3.8). Only the
         # chart version is stamped (--version); Chart.yaml keeps its 0.0.0-dev /
@@ -1105,10 +1109,10 @@ jobs:
           printf '%s %s@%s\n' "${CHART_NAME}" "${ref}" "${digest}" \
             > "${RUNNER_TEMP}/chart-digests/${CHART_NAME}.txt"
       - name: Install cosign
-        if: needs.changes.outputs[matrix.changed] == 'true'
+        if: needs.changes.outputs[matrix.changed] == 'true' || needs.changes.outputs.ci == 'true'
         uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
       - name: 'Sign chart - ${{ matrix.chart-name }}'
-        if: needs.changes.outputs[matrix.changed] == 'true'
+        if: needs.changes.outputs[matrix.changed] == 'true' || needs.changes.outputs.ci == 'true'
         # Sign the pushed OCI chart by digest with the keyed cosign pair, matching
         # the image signing in `publish`, so the build-lane haul verifies charts and
         # images alike at sync time (ADR 0033). Keyed, no transparency log.
@@ -1123,7 +1127,7 @@ jobs:
           cosign sign --key env://COSIGN_PRIVATE_KEY \
             --use-signing-config=false --tlog-upload=false --yes "${ref}"
       - name: 'Upload chart digest - ${{ matrix.chart-name }}'
-        if: needs.changes.outputs[matrix.changed] == 'true'
+        if: needs.changes.outputs[matrix.changed] == 'true' || needs.changes.outputs.ci == 'true'
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: chart-digest-${{ matrix.chart-name }}
@@ -1233,14 +1237,11 @@ jobs:
     # (harmless) on partial builds until a one-time bootstrap seeds the first
     # predecessor. Sync verifies every component's cosign signature (--key) and
     # fails closed if any component is unsigned or unverifiable.
-    # PAUSED until the build-lane haul project is finalized. Chart carry needs a
-    # one-time publish-all seed of the 14 charts (see components.txt); until that
-    # lands this job fails on every main run. continue-on-error keeps it non-gating,
-    # but it still leaves a red X on the run, so it is turned off while the whole
-    # GitOps build-lane is still in progress rather than run half-built. Re-enable
-    # (after the seed) by restoring the condition:
-    #   github.ref_name == 'main' && needs.publish.result == 'success' && needs.publish-charts.result == 'success'
-    if: false
+    # LIVE (build-lane finalize). publish-charts now force-publishes all 14 charts on a
+    # ci/build-tooling change (changes.ci), so the first main run that carries this change
+    # seeds all 14 into the registry and this job folds them into the haul; each then
+    # carries by digest until it changes. Main-only, after images + charts publish.
+    if: github.ref_name == 'main' && needs.publish.result == 'success' && needs.publish-charts.result == 'success'
     runs-on: ubuntu-latest
     continue-on-error: true
     needs: [changes, publish, publish-charts]
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1226,20 +1226,13 @@ jobs:
             exit 1
           fi
   publish-haul:
-    # Build-lane haul (ADR 0033): after images publish, render the manifest from
-    # this run's fresh digests plus the previous build's haul (carry), sync it into
-    # a .tar.zst bundle, and publish the bundle + the manifest so the next run can
-    # carry from it. continue-on-error so it never gates a release; it fails closed
-    # (harmless) on partial builds until a one-time bootstrap seeds the first
-    # predecessor. Sync verifies every component's cosign signature (--key) and
-    # fails closed if any component is unsigned or unverifiable.
-    # LIVE (build-lane finalize). Carries every chart by digest, so all 14 must be in
-    # the registry: publish-charts is changed-only, so run seed-charts.yaml once at
-    # go-live to publish the never-changed charts (deliberately NOT wired to changes.ci
-    # -- ADR 0030 keeps routine builds to changed charts only). Main-only, after publish.
+    # Build-lane haul (ADR 0033): render the manifest (fresh digests + carry), sync to a
+    # signed .tar.zst, publish the bundle + manifest. GATES the build (no continue-on-error):
+    # a broken/incomplete/unsigned air-gap haul must fail loud. Safe to merge only after
+    # seed-charts.yaml has published all 14 charts (publish-charts is changed-only) and the
+    # Docker Hub secret is seeded (the upstream pulls below are otherwise rate-limited).
     if: github.ref_name == 'main' && needs.publish.result == 'success' && needs.publish-charts.result == 'success'
     runs-on: ubuntu-latest
-    continue-on-error: true
     needs: [changes, publish, publish-charts]
     permissions:
       contents: read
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -1066,14 +1070,14 @@ jobs:
...
-        if: needs.changes.outputs[matrix.changed] == 'true'
+        if: needs.changes.outputs[matrix.changed] == 'true' || needs.changes.outputs.ci == 'true'
         uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: 'Publish chart - ${{ matrix.chart-name }}'
-        if: needs.changes.outputs[matrix.changed] == 'true'
+        if: needs.changes.outputs[matrix.changed] == 'true' || needs.changes.outputs.ci == 'true'
```
kathrynalpert: We do want to seed the charts, but it seems better to do so as a one-time `workflow_dispatch`. `changes.ci` covers
`.github/workflows/ci.yaml` and `.github/actions/**` — 48 of the last 174 merges to main, about
one every other day. Each one publishes 14 new chart versions and triggers a ~5GB haul re-push.

ADR 0030 is explicit that charts differ from images here: chart versions land in `helm.sh/chart`
pod-template labels, so packaging every chart on every merge restarts pods for no reason, and
consumers pin exact versions with Renovate bumping them.

The comment above describes this as "mirroring build-and-upload for images," which is worth being
precise about. `build-and-upload` does use `|| changes.ci`, but the image `publish` job gates on
`changed || !exists` and never on `ci`. So on a CI-only merge images rebuild and publish nothing,
while charts here would publish 14 new artifacts. The mirror holds for building and breaks for
publishing, which is the half that matters.

Suggest seeding the 14 charts through `bootstrap-haul.yaml` — or a `workflow_dispatch` input on
this workflow — the way the 8 images were seeded, and leaving these conditions changed-only.

One counter point: the `helm package` invocation does live in `ci.yaml`, so a
change there genuinely can alter chart output. If we want to keep a content-driven trigger, scope
it to that file rather than to all of `.github/actions/**`.

.github/workflows/ci.yaml
```diff
@@ -1011,6 +1011,10 @@ jobs:
   # Publish changed Helm charts as versioned OCI artifacts (ADR 0030/0031 phase 0),
   # gated like the image `publish` job. Vertical slice: hl7-transformer only.
   publish-charts:
+    # Per-chart steps publish a chart whose dir changed, OR every chart on a
+    # CI/build-tooling change (changes.ci), mirroring build-and-upload for images
+    # (ADR 0030: CI/build-tooling changes rebuild everything). That force-all also
+    # gives publish-haul its one-time all-14-charts seed on any ci-touching merge.
```
kathrynalpert: "one-time all-14-charts seed **on any ci-touching merge**" — those two halves contradict each
other, and the second is what the code does. Whichever way the condition lands, worth making the
comment say which.

.github/workflows/ci.yaml
```diff
@@ -1233,14 +1237,11 @@ jobs:
...
-    # (after the seed) by restoring the condition:
-    #   github.ref_name == 'main' && needs.publish.result == 'success' && needs.publish-charts.result == 'success'
-    if: false
+    # LIVE (build-lane finalize). publish-charts now force-publishes all 14 charts on a
+    # ci/build-tooling change (changes.ci), so the first main run that carries this change
+    # seeds all 14 into the registry and this job folds them into the haul; each then
+    # carries by digest until it changes. Main-only, after images + charts publish.
+    if: github.ref_name == 'main' && needs.publish.result == 'success' && needs.publish-charts.result == 'success'
     runs-on: ubuntu-latest
     continue-on-error: true
```
kathrynalpert: Can we flip this to `false`? The reason for the flag was that the job couldn't succeed before the
seed, and this PR fixes that. Now that the haul is the air-gap delivery path, a build that
published a broken or incomplete haul should fail loudly.

One note:

- **The job pulls Docker Hub anonymously.** It logs into GHCR only, but 6 of the 9 upstream
  images come from Docker Hub (`dcm4che` ×3, `jodogne/orthanc-plugins`,
  `adorsys/keycloak-config-cli`, `openpolicyagent/opa`), and both `imagetools inspect` and
  `hauler store sync` pull them unauthenticated from a shared runner IP. That is the classic
  rate-limit flake, and under this flip it turns a third-party hiccup into a red main — and into a
  blocked release, since `release.yaml`'s `wait-for-build` treats any CI conclusion other than
  success as fatal. A Docker Hub login in this job would remove the biggest source of spurious
  failure before we make it blocking.

kathrynalpert (commented):
Two requests and a sequencing note.

**1. Don't make the seed permanent.** `|| needs.changes.outputs.ci == 'true'` is a permanent
condition doing a one-time job. `changes.ci` is `.github/workflows/ci.yaml` plus
`.github/actions/**`, which is 48 of the last 174 merges to main — about one every other day —
and each one would publish 14 new chart versions and re-push the ~5GB haul. ADR 0030 rules this
out for charts specifically:

> Even that minimal job packages only charts that changed: chart versions end up in pod-template
> labels (`helm.sh/chart`), so packaging every chart on every merge would restart pods for no
> reason. Consumers pin exact chart versions and Renovate bumps them.

We already have the right pattern: `bootstrap-haul.yaml` is a one-time `workflow_dispatch` that
seeded the 8 images, and its own header says "Run once." Seeding the 14 charts the same way keeps
`publish-charts` changed-only per ADR 0030 and costs one dispatch. Details inline.

**2. Flip `continue-on-error` to `false`.** The haul is the air-gap delivery path now, and a
green build that published a broken haul is worse than a red one. The original reason for the
flag was that the job couldn't succeed pre-seed, which this PR fixes. One thing to handle
alongside it: the job pulls its Docker Hub images anonymously, so it should get a Docker Hub
login before we make it blocking (more details inline).

**3. Sequencing with #643.** Every main build republishes the full ~5GB bundle, and at 174 merges
per 90 days that is roughly 10GB/day accruing from the moment this merges. #643 is dry-run by
default and prunes nothing until `HAUL_RETENTION_ENABLED` is `'true'`. I'd land #643 first, or
the same day, and set `HAUL_RETENTION_ENABLED` — so retention exists before the accumulation
starts rather than after.

andrey-embarklabs:
> Two requests and a sequencing note.
> 
> **1. Don't make the seed permanent.** `|| needs.changes.outputs.ci == 'true'` is a permanent condition doing a one-time job. `changes.ci` is `.github/workflows/ci.yaml` plus `.github/actions/**`, which is 48 of the last 174 merges to main — about one every other day — and each one would publish 14 new chart versions and re-push the ~5GB haul. ADR 0030 rules this out for charts specifically:
> 
> > Even that minimal job packages only charts that changed: chart versions end up in pod-template
> > labels (`helm.sh/chart`), so packaging every chart on every merge would restart pods for no
> > reason. Consumers pin exact chart versions and Renovate bumps them.
> 
> We already have the right pattern: `bootstrap-haul.yaml` is a one-time `workflow_dispatch` that seeded the 8 images, and its own header says "Run once." Seeding the 14 charts the same way keeps `publish-charts` changed-only per ADR 0030 and costs one dispatch. Details inline.
> 
> **2. Flip `continue-on-error` to `false`.** The haul is the air-gap delivery path now, and a green build that published a broken haul is worse than a red one. The original reason for the flag was that the job couldn't succeed pre-seed, which this PR fixes. One thing to handle alongside it: the job pulls its Docker Hub images anonymously, so it should get a Docker Hub login before we make it blocking (more details inline).
> 
> **3. Sequencing with #643.** Every main build republishes the full ~5GB bundle, and at 174 merges per 90 days that is roughly 10GB/day accruing from the moment this merges. #643 is dry-run by default and prunes nothing until `HAUL_RETENTION_ENABLED` is `'true'`. I'd land #643 first, or the same day, and set `HAUL_RETENTION_ENABLED` — so retention exists before the accumulation starts rather than after.

agreed on the permanent-condition-for-a-one-time-job. reverted publish-charts back to changed-only and added seed-charts.yaml, a one-time workflow_dispatch that publishes + signs all 14 charts. go-live is now merge + run the seed once.

also flipped publish-haul off continue-on-error so a broken or unsigned haul fails the build, and added a docker hub login for the upstream pulls. two hard prereqs before this merges: run seed-charts once, and seed the DOCKERHUB_USERNAME/TOKEN secrets, otherwise the now-gating job reds every build.

.github/workflows/seed-charts.yaml
```diff
@@ -0,0 +1,62 @@
+name: Seed charts
+
+# One-time manual publish of ALL charts to the OCI registry so the build-lane haul
```
kathrynalpert: This is the part that doesn't hold: Publishing the charts to GHCR
isn't what makes them carryable — `build_haul.py`'s carry is strictly
`carried.get(repo)` from the **predecessor haul manifest**, with no live registry lookup. That's a
deliberate ADR 0033 choice, since vendored images have no stable tag to inspect.

So on the first main build after seeding:

- charts aren't in `fresh` — `publish-charts` is correctly back to changed-only, so at most one or
  two are
- charts aren't in the predecessor — the bootstrap seeded the 8 images only
- `resolve_refs` raises `no tag+digest for ghcr.io/washu-tag/charts/… (neither rebuilt nor carried)`

and with `continue-on-error` gone, that reds every main build until it's fixed. It also blocks
releases, since `wait-for-build` treats any non-success CI conclusion as fatal.

The fix is to give this workflow the same tail `bootstrap-haul.yaml` has: oras-pull
`manifests/scout-manifest:main`, add the 14 fresh chart digests to the carried 8 images, then
render / sync / publish. Then the next `ci.yaml` run has all 22 in its predecessor and carries
normally. Bootstrap's own header already flags that this tail is duplicated and wants to be a
composite action — this is probably the moment.

Minor, while you're in here: `VERSION` is minted from `GITHUB_RUN_NUMBER`, which is scoped to
*this* workflow, so the seeded charts get very low run numbers (`0.<date>.1`) rather than
continuing `ci.yaml`'s sequence. Benign — they sort below every future build tag, which is the
right order for older artifacts — but it's exactly the footgun `ci.yaml`'s "Mint build version"
comment warns about.

.github/workflows/ci.yaml
```diff
@@ -1228,23 +1228,13 @@ jobs:
...
-    # one-time publish-all seed of the 14 charts (see components.txt); until that
-    # lands this job fails on every main run. continue-on-error keeps it non-gating,
-    # but it still leaves a red X on the run, so it is turned off while the whole
-    # GitOps build-lane is still in progress rather than run half-built. Re-enable
-    # (after the seed) by restoring the condition:
-    #   github.ref_name == 'main' && needs.publish.result == 'success' && needs.publish-charts.result == 'success'
-    if: false
+    # Build-lane haul (ADR 0033): render the manifest (fresh digests + carry), sync to a
+    # signed .tar.zst, publish the bundle + manifest. GATES the build (no continue-on-error):
+    # a broken/incomplete/unsigned air-gap haul must fail loud. Safe to merge only after
```
kathrynalpert, line 1233: The stated order can't be executed: `workflow_dispatch` only triggers for workflows on the default
branch, so `seed-charts.yaml` can't be run until this PR merges. The prerequisite therefore can't
be satisfied before the thing it's a prerequisite for.

With the job now gating, that leaves a guaranteed red-main window between merge and seed — plus
any release attempted in that window aborts at `wait-for-build`. Two ways out: land the seed
workflow in its own PR first, run it, then merge this one; or merge this with the job still
non-gating and flip `continue-on-error` in a follow-up once the first complete haul exists.

(Both of these fold together if the seed workflow also publishes the haul, per the other comment:
then seeding leaves a complete predecessor and the first gated build passes.)

andrey-embarklabs:
> Claude's still concerned about this one: the seed publishes the charts but never reaches the path the haul actually carries from, so the first main build after merge fails, and it now fails the build. `bootstrap-haul.yaml` doesn't just publish artifacts — it publishes a full haul, which is what leaves a predecessor behind. `seed-charts.yaml` does the publish half only.

good catch that publishing the charts wasn't enough. reworked it, pushed as 541370bc:

seed-charts is now a single job that publishes + signs the 14 charts AND leaves a
complete predecessor: it oras-pulls :main (bootstrap's 8 images), folds in the 14
fresh chart digests, and publishes a full 22-component haul through the new composite.
so the next ci.yaml run carries all 22.

lifted the sync/save/push/sign/tag-main tail into .github/actions/publish-haul, shared
by publish-haul, seed-charts, and bootstrap. bonus: bootstrap's copy was missing the
--key fail-closed verification, so the composite closes that drift too.

the merge-run race (workflow_dispatch can't run the seed until this merges, but the
merge's own build hits publish-haul because publish/publish-charts report success on
skipped steps): added a build_haul --check preflight. on a pre-seed run the predecessor
is images-only, so --check returns incomplete and render/publish/surface/upload all skip
with a warning instead of reddening. a genuine drift still fails loud in the render. once
the seed runs, every build publishes normally.

guarded the Docker Hub login on env.DH_USER so an unseeded secret doesn't hard-fail the
now-gating job. fixed the stale continue-on-error comment and 6 -> 8 upstream images.

go-live is now: merge -> run seed-charts once (the --check guard keeps builds green in
between). one thing i couldn't prove offline: that hauler 2.0.2 emits "signature verified
for image [" for OCI charts the same as images. if it phrases chart verification
differently the count-gate fails closed (safe/loud), and the seed run is the first real
test of it, so worth watching that run.

PR #643, opened 2026-08-12
ci: bound build-lane haul storage with a retention workflow

andrey-embarklabs:
### Technical
Every main build republishes the whole ~5GB platform to `manifests/scout` + `manifests/scout-manifest`, so unbounded they grow indefinitely. Adds a weekly + manual retention workflow that keeps `:main` + the newest N versioned hauls (and their cosign signatures) and prunes older ones.

- `tooling/manifest/retention.py` — signature-aware keep/delete policy, conservative (under-prunes rather than over-prunes), with offline unit tests.
- `.github/workflows/haul-retention.yaml` — **hard-scoped** to a literal allowlist of the two `manifests/*` packages (never enumerates, so it cannot reach the digest-pinned image/chart packages Flux references). **Dry-run by default**; real deletes only on an explicit dispatch (`dry_run=false`) or a scheduled run once the repo variable `HAUL_RETENTION_ENABLED='true'`.

## Type of change
- [x] CI / build automation

### Backward compatibility
None. Read-only until explicitly enabled; touches only the two haul packages.

## Note for reviewers
Reviewed adversarially for scope-escape and over-deletion (none). To roll out: merge, dispatch once with `dry_run=true` to eyeball the plan against the real registry, then set `HAUL_RETENTION_ENABLED=true` to let the weekly schedule prune.

### DIFF
--- a/.github/workflows/haul-retention.yaml
+++ b/.github/workflows/haul-retention.yaml
@@ -4,8 +4,8 @@ name: Haul retention
 # accumulate unbounded. Scoped HARD to the two manifests/* packages via a literal
 # allowlist below; it never enumerates packages, so it cannot reach the
 # digest-pinned image/chart packages that Flux consumers reference. Keeps :main +
-# the newest N versioned hauls (+ their cosign signatures) for rollback -- policy
-# is tooling/manifest/retention.py (unit-tested offline).
+# hauls younger than keep_days (min_keep newest regardless of age) + their cosign
+# signatures -- policy is tooling/manifest/retention.py (unit-tested offline).
 #
 # SAFE BY DEFAULT: a manual run only PREVIEWS the plan (dry_run=true). The weekly
 # schedule deletes for real only once the repo variable HAUL_RETENTION_ENABLED is
@@ -16,10 +16,14 @@ on:
     - cron: '17 8 * * 1' # Mondays 08:17 UTC
   workflow_dispatch:
     inputs:
-      keep_n:
-        description: 'Versioned hauls to retain per package (for rollback)'
+      keep_days:
+        description: 'Retain hauls younger than this many days (per package)'
         type: number
-        default: 10
+        default: 30
+      min_keep:
+        description: 'Always retain at least this many newest hauls'
+        type: number
+        default: 5
       dry_run:
         description: 'Preview only; do not delete'
         type: boolean
@@ -48,9 +52,9 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           # Schedule has no inputs -> 10. On dispatch, pass the operator value through
-          # verbatim (don't `|| 10`: keep_n=0 is falsy and would silently become 10,
-          # masking retention.py's keep_n>=1 guard -- let it fail loudly instead).
-          KEEP_N: ${{ github.event_name == 'schedule' && 10 || inputs.keep_n }}
+          # verbatim (don't coerce: retention.py's guards reject 0 / negatives loudly).
+          KEEP_DAYS: ${{ github.event_name == 'schedule' && 30 || inputs.keep_days }}
+          MIN_KEEP: ${{ github.event_name == 'schedule' && 5 || inputs.min_keep }}
           # Delete for real ONLY on an explicit dispatch with dry_run=false, or a
           # scheduled run once HAUL_RETENTION_ENABLED is set. Otherwise preview.
           DELETE_FOR_REAL: ${{ (github.event_name == 'workflow_dispatch' && inputs.dry_run == false) || (github.event_name == 'schedule' && vars.HAUL_RETENTION_ENABLED == 'true') }}
@@ -74,10 +78,9 @@ jobs:
               exit 1
             fi
             # Surface retention.py's plan/error even on failure -- set -e would
-            # otherwise abort before the summary line and hide the reason (e.g. the
-            # keep_n>=1 guard).
+            # otherwise abort before the summary line and hide the reason (e.g. a guard).
             if ! printf '%s' "${raw}" | jq -s 'add' \
-                | python3 tooling/manifest/retention.py --keep "${KEEP_N}" - >delete-ids.txt 2>plan.txt; then
+                | python3 tooling/manifest/retention.py --keep-days "${KEEP_DAYS}" --min-keep "${MIN_KEEP}" - >delete-ids.txt 2>plan.txt; then
               echo "::error::retention.py failed for ${pkg}:"; cat plan.txt
               sed 's/^/    /' plan.txt >> "$GITHUB_STEP_SUMMARY"
               exit 1
--- a/.github/workflows/haul-retention.yaml
+++ b/.github/workflows/haul-retention.yaml
@@ -73,24 +73,36 @@ jobs:
               cat err.txt
               exit 1
             fi
-            printf '%s' "${raw}" | jq -s 'add' \
-              | python3 tooling/manifest/retention.py --keep "${KEEP_N}" - >delete-ids.txt 2>plan.txt
+            # Surface retention.py's plan/error even on failure -- set -e would
+            # otherwise abort before the summary line and hide the reason (e.g. the
+            # keep_n>=1 guard).
+            if ! printf '%s' "${raw}" | jq -s 'add' \
+                | python3 tooling/manifest/retention.py --keep "${KEEP_N}" - >delete-ids.txt 2>plan.txt; then
+              echo "::error::retention.py failed for ${pkg}:"; cat plan.txt
+              sed 's/^/    /' plan.txt >> "$GITHUB_STEP_SUMMARY"
+              exit 1
+            fi
             sed 's/^/    /' plan.txt >> "$GITHUB_STEP_SUMMARY"
             count="$(grep -c . delete-ids.txt || true)"
             if [ "${DELETE_FOR_REAL}" = "true" ]; then
-              # Non-fatal per id: a single stuck version (e.g. transient 403) must not
-              # abort the run and block the other package. retention.py recomputes from
-              # live state each run, so a skipped delete just retries next time.
+              # Per-id tolerance so one stuck version doesn't block the rest, but track
+              # the counts: every delete failing (with deletes attempted) is systemic --
+              # usually a permissions gap dry-run can't exercise -- so fail loud.
+              deleted=0; failed=0
               while read -r id; do
                 [ -n "${id}" ] || continue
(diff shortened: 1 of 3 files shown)


.github/workflows/haul-retention.yaml
```diff
@@ -0,0 +1,97 @@
...
+            fi
+            printf '%s' "${raw}" | jq -s 'add' \
+              | python3 tooling/manifest/retention.py --keep "${KEEP_N}" - >delete-ids.txt 2>plan.txt
+            sed 's/^/    /' plan.txt >> "$GITHUB_STEP_SUMMARY"
+            count="$(grep -c . delete-ids.txt || true)"
+            if [ "${DELETE_FOR_REAL}" = "true" ]; then
+              # Non-fatal per id: a single stuck version (e.g. transient 403) must not
+              # abort the run and block the other package. retention.py recomputes from
+              # live state each run, so a skipped delete just retries next time.
+              while read -r id; do
```
kathrynalpert, line 96: Per-id tolerance is right for one stuck version, but nothing distinguishes "one id failed" from
"every id failed." The loop never exits non-zero, and the summary below says "processed N
version(s) for deletion" whether N deletions succeeded or zero did. That is the same silent
success the list step above explicitly refuses to allow.

The most likely systemic cause for failure is permissions, and **dry-run can never
exercise this**. The preview runs the list and the plan but never issues a DELETE, so the first real
run is the first test of whether `GITHUB_TOKEN` with `packages: write` can delete an *org*
container package version — that depends on the package being repo-scoped to this repo, and
org-level packages have historically needed a PAT with `delete:packages`. If it can't, every
Monday's run is green, deletes nothing, and the storage this job exists to bound keeps growing.

Suggest counting successes and failures and failing the job when the plan was non-empty and
nothing was deleted. Cheap, and it turns the one failure mode we can't pre-test into a loud one.

kathrynalpert (commented):
Comments inline. One question that isn't about a specific line: `keep_n: 10` is described as
retention "for rollback." At 174 merges per 90 days that is roughly two hauls a day, so 10 is
about **five days** of rollback for ~50GB. If the rollback horizon is what actually matters,
keep-by-age might express the intent better than keep-by-count (with some minimum count retained).

.github/workflows/haul-retention.yaml
```diff
@@ -0,0 +1,97 @@
...
+              if grep -q 'HTTP 404' err.txt; then
+                echo "  package does not exist yet; skipping" | tee -a "$GITHUB_STEP_SUMMARY"
+                continue
+              fi
+              echo "::error::failed to list versions for ${pkg}:"
+              cat err.txt
+              exit 1
+            fi
+            printf '%s' "${raw}" | jq -s 'add' \
+              | python3 tooling/manifest/retention.py --keep "${KEEP_N}" - >delete-ids.txt 2>plan.txt
```
kathrynalpert: `plan.txt` only reaches the step summary on the success path — `set -e` aborts the step before the
`sed ... >> "$GITHUB_STEP_SUMMARY"` line. So any `retention.py` failure gives a red step with no
visible reason, including the deliberate `keep_n must be >= 1` guard that the `KEEP_N` comment
above relies on ("let it fail loudly instead" — it does fail, but silently about why).

`tee` the stderr, or cat `plan.txt` in a failure branch.

tooling/manifest/retention.py
```diff
@@ -0,0 +1,140 @@
...
+
+    deleted_digests = {digest_of[i] for i in delete if digest_of.get(i)}
+    for vid, subject in sigs:
+        if subject in kept_digests:
+            keep.add(vid)
+        elif subject in deleted_digests:
+            delete.add(vid)  # orphaned by deleting its bundle -> clean it up
+        else:
+            keep.add(vid)  # subject unknown/absent: conservative keep
+    return keep, delete
```
kathrynalpert, line 119: These two sets can overlap. A version tagged `main` whose tags also matched `sha256-<hex>.sig`
lands in `keep` via the `main` branch and in `delete` via the signature loop, and the workflow acts
on `delete` alone.

Not reachable today — cosign never tags a signature `main` — so this is hardening, not a bug. But
in code whose output is a delete list, "keep wins" is worth making structural rather than
incidental: `delete -= keep` before the return, plus an assertion that they're disjoint.

tooling/manifest/retention.py
```diff
@@ -0,0 +1,140 @@
...
+        sys.stdin.read()
+        if args.versions_json == "-"
+        else open(args.versions_json).read()
+    )
+    data = json.loads(raw)
+    versions = [
+        {
+            "id": v["id"],
+            "created_at": v.get("created_at", ""),
+            "digest": v.get("name", ""),  # GHCR container version `name` == its digest
```
kathrynalpert: `partition` is covered well, but this adapter is the only part touching the real API shape and it
has no test. The mapping looks correct — GHCR container versions do use `name` for the full
`sha256:` digest — but if it ever drifted, signature matching would fail *silently*: every sig
would fall through to the conservative keep on line 89 and orphans would accumulate with nothing
reporting it.

One fixture test against a real `GET .../versions` payload would lock the shape in.

tooling/manifest/retention.py
```diff
@@ -0,0 +1,140 @@
...
+    digest_of = {}
+
+    for v in versions:
+        vid = v["id"]
+        tags = v.get("tags") or []
+        digest_of[vid] = v.get("digest")
+        if not tags:
+            keep.add(vid)  # untagged: conservative keep
+            continue
+        if "main" in tags:
```
kathrynalpert: Forward-looking, but I'd rather land it here than discover it later: once the release lane
publishes a haul at `X.Y.Z` — whether as its own push or as an `oras tag` alias of the verified
build bundle — this policy would prune it in about five days.

Ranking is pure recency and tag-agnostic, so a version tagged `4.2.0` is just another entry in
`artifacts` once `main` moves off it, and it drops out at `i >= keep_n`. Deletion is per *version*,
not per tag, so we would not lose a stale build tag and keep `4.2.0` — the whole artifact goes, and
`deleted_digests` takes its cosign signature with it. At ~2 hauls/day every release haul would die
well before the next release is cut.

Suggest a keep rule right next to this one:

```python
_RELEASE_TAG = re.compile(r"^[1-9]\d*\.\d+\.\d+$")   # X.Y.Z, X >= 1
```

The leading `[1-9]` is the part to get right. ADR 0030 makes the permanent leading `0` the
build-lane marker precisely so the lanes are distinguishable, so a naive `^\d+\.\d+\.\d+$` would
also match `0.20260812.1234`, keep every build haul forever, and quietly turn this job off. Worth a
test for that case specifically. (ADR 0030 also reserves `X.Y.Z-rc.<run>` under deferred work, if
you want the pattern to allow for it now.)

The alternative is publishing release hauls to a package outside the two-name allowlist, which is
structurally safer since retention could never reach it — but it forecloses the cheap alias
approach, because you would have to re-push ~5GB into a second repo rather than adding a tag.

andrey-embarklabs:
> Comments inline. One question that isn't about a specific line: `keep_n: 10` is described as retention "for rollback." At 174 merges per 90 days that is roughly two hauls a day, so 10 is about **five days** of rollback for ~50GB. If the rollback horizon is what actually matters, keep-by-age might express the intent better than keep-by-count (with some minimum count retained).

thanks, all five landed:

- delete -= keep before return (plus a disjoint assert) so keep always wins.
- pulled the GHCR-json to partition mapping into versions_from_api with a fixture test against the real GET .../versions shape, so a silent drift in the digest/tags mapping gets caught.
- the workflow now surfaces retention.py's plan on failure (set -e was hiding it) and fails loud when every delete fails (systemic, usually perms, which dry-run can't exercise).

switched it to age-based: keep_days (default 30) with a min_keep floor (default 5) so a quiet period never prunes the rollback set. same signature-follows-bundle handling.

tooling/manifest/retention.py
```diff
@@ -0,0 +1,167 @@
...
+`manifests/scout` (the .tar.zst bundle) and `manifests/scout-manifest` (the
+manifest), tagged `0.YYYYMMDD.<run>` with `:main` moved onto the newest. Left
+unbounded they accumulate ~5GB per build. This computes which package *versions*
+to keep vs delete; the workflow feeds it `GET .../versions` and acts on the split.
+
+Policy is by AGE, not count, so the rollback window is a fixed time horizon
+regardless of build cadence (deliberately conservative -- deleting the wrong
+version breaks Flux consumers that pin by digest, so we under-prune):
+  - KEEP any version tagged `main` (the pointer the next build carries from).
+  - KEEP version-tagged artifacts younger than `keep_days`.
```
kathrynalpert, line 17: Not seeing the release-tag keep rule land here — there's no `_RELEASE_TAG` in the file, and no test for it.

Same suggestion as before, as a sibling of the `main` rule:

```python
_RELEASE_TAG = re.compile(r"^[1-9]\d*\.\d+\.\d+$")   # X.Y.Z, X >= 1
```

The leading `[1-9]` is important — ADR 0030 makes the permanent leading `0` the
build-lane marker, so a naive `^\d+\.\d+\.\d+$` also matches `0.20260812.1234`, keeps every build
haul forever, and quietly turns this job off. Worth a test for that case specifically.

.github/workflows/haul-retention.yaml
```diff
@@ -0,0 +1,112 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+        with:
+          python-version: '3.12'
+      - name: Prune old haul package versions
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # Schedule has no inputs -> 10. On dispatch, pass the operator value through
```
kathrynalpert: Stale from the count-based version: "Schedule has no inputs -> 10" now sets 30 and 5.

andrey-embarklabs:
both follow-ups landed, pushed as c8dd8183:

added _RELEASE_TAG = ^[1-9]\d*\.\d+\.\d+$ as a sibling of the main rule (via
_is_pinned()), so a release haul at X.Y.Z is pinned forever. the leading [1-9] keeps
the build lane (0.YYYYMMDD.run) aging out. tests cover both: ancient 4.2.0 kept,
ancient 0.20260812.1234 deleted, and the release sig follows its bundle.

fixed the stale schedule comment (-> 10 is now keep_days=30, min_keep=5).

heads up: the release-tag hauls this protects only start existing once #641's alias-haul
lands, so this pairs with #641.

